PR #116, opened 2017-12-04
rename mssqldb module as sqldb

krancour:
This got missed when I tweaked the package names of the modules the other night to better align with MASB and with the names of the services they offer.


krancour:
Yep. Investigating.

PR #119, opened 2017-12-05
Fix Helm prereq. command

MikkelHegn:
The command to create the ClusterRoleBinding for Helm fails as it point to the HTML'ified version of the yaml file on GitHub. Changed to use the raw file.

### DIFF
--- a/docs/quickstart-minikube.md
+++ b/docs/quickstart-minikube.md
@@ -189,6 +189,7 @@ Next we will create a local cluster using Minikube. _Support for AKS is coming s
 
     **Bash**
     ```
+    helm repo add azure https://kubernetescharts.blob.core.windows.net/azure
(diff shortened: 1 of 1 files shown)


MikkelHegn:
I ran in to this issue with the workdpress sample - might be worth noting at this point in the quickstart: https://github.com/kubernetes/minikube/issues/2256

krancour:
@MikkelHegn thanks! Solid improvements.

> I ran in to this issue with the workdpress sample - might be worth noting at this point in the quickstart: kubernetes/minikube#2256

Can you open a separate issue for that, please? Merging these fixes and don't want to lose track.

MikkelHegn:
Done: https://github.com/Azure/open-service-broker-azure/issues/122

Issue #126, opened 2017-12-06
Switch quickstart to install ghost instead of wordpress


krancour:
The Wordpress chart does not install cleanly _on minikube_ due to https://github.com/kubernetes/minikube/issues/2256.

To avoid requiring a specific version of minikube, we're going to (temporarily) update the quickstart to install ghost instead-- that chart should not suffer from the same issue.

krancour:
The Ghost chart is proving unreliable (at least on minikube).

@jeremyrickard and I have both encountered this:

```
$ k logs quickstart-ghost-58f59d76ff-lkmms

Welcome to the Bitnami ghost container
Subscribe to project updates by watching https://github.com/bitnami/bitnami-docker-ghost
Submit issues and feature requests at https://github.com/bitnami/bitnami-docker-ghost/issues
Send us your feedback at containers@bitnami.com

nami    INFO  Initializing mysql-client
nami    INFO  mysql-client successfully initialized
nami    INFO  Initializing ghost
handler WARN  Using /opt/bitnami/ghost as working directory
mysql-c INFO  Trying to connect to MySQL server
mysql-c INFO  Found MySQL server listening at b987634c-39df-4197-bd05-f2a2ff61c44c.mysql.database.azure.com:3306
mysql-c INFO  MySQL server listening and working at b987634c-39df-4197-bd05-f2a2ff61c44c.mysql.database.azure.com:3306
ghost   INFO  ==> Configuring Database...
ghost   INFO  ==> Configuring Ghost...
Error executing 'postInstallation': Terminated
```

This change to the QS would be ill-advised right now.

Issue #122, opened 2017-12-05
Quickstart failing on Minikube due to kubernetes/minikube#2256


MikkelHegn:
When installing the Workdpress chart on Minikube, I run in to the issue descirbed here: kubernetes/minikube#2256.

The proposed change to manually create the directory ~~and change permission~~, seems to get around the volume issue.

1. Install the chart
2. Create the directory manually in the minikube VM
3. Delete the first pod that was created

This ends up in a new state where another issue with missing secrets, which I guess is because the Azure resources ends up not being provisioned.

Error: secrets "quickstart-wordpress-mysql-secret" not found
Error syncing pod

MikkelHegn:
The secret error is caused by the chart not supporting the embedded MariaDB option.

krancour:
The workaround seems like it might distract from what we're trying to emphasize-- which is the simplicity with which complex things backed by services in Azure can be deployed to k8s.

I need to look at this chart, but it might have an option to disable this volume mount (and just use ephemeral disk). Maybe that option can be set in the walkthrough. We'll definitely figure _something_ out to get around this.

> This ends up in a new state where another issue with missing secrets, which I guess is because
> the Azure resources ends up not being provisioned.
>
> Error: secrets "quickstart-wordpress-mysql-secret" not found
> Error syncing pod

This is normal. The secret in question isn't created until the MySQL instance is provisioned and bound to. When those things complete, the whole deployment eventually reaches a consistent state.

MikkelHegn:
Thanks @krancour - however I never got to a state where I See any calls to the OSBA web-api or deployments in Azure. Maybe because of the pod deployment error.

krancour:
@MikkelHegn can you try this and let me know if it improves your experience? I'm doing the same right now:

```
helm install azure/wordpress --name quickstart --namespace quickstart --set persistence.enabled=false
```

MikkelHegn:
Yes - this works. I know have to look into OSBA and Azure issue with auth :-)

I see this in the OSBA:

[...]

PR #128, opened 2017-12-06
Switching the quickstart to use Ghost instead of Wordpress

### LINKED ISSUE #125: Port is not right in quickstart
The following line from the quickstart is meant to open Wordpress, but doesn't universally work because the node port the service is assigned to will not always bee `31215`.

```
open http://$(minikube ip):31215/admin
```

arschles:
Because of https://github.com/Azure/open-service-broker-azure/issues/122, Wordpress won't work on the latest Minikube because of a persistent volume related issue in MK. So, we're switching to Ghost for the walkthrough.

@jeremyrickard I think this will work, but I haven't finished testing on Minikube. While I do so, mind checking my work?

Fixes https://github.com/Azure/open-service-broker-azure/issues/126

Note that I did not fix #125 in this PR

cc/ @krancour @carolynvs

### DIFF
--- a/docs/quickstart-minikube.md
+++ b/docs/quickstart-minikube.md
@@ -1,11 +1,11 @@
 # Quickstart: Open Service Broker for Azure on a Minikube cluster
 This quickstart walks-through using the Open Service Broker for Azure (OSBA) to
-deploy WordPress on a local Minikube cluster.
+deploy [Ghost](https://ghost.org/) Blog on a local Minikube cluster.
 
-WordPress requires a backend MySQL database. Without OSBA, we would create a database
+Ghost requires a backend MySQL database. Without OSBA, we would create a database
 in the Azure portal, and then manually configure the connection information. Now
 with OSBA our Kubernetes manifests can provision an Azure MySQL database on our behalf,
-save the connection information in Kubernetes secrets, and then bind them to our WordPress instance.
+save the connection information in Kubernetes secrets, and then bind them to our Ghost instance.
 
 * [Prerequisites](#prerequisites)
 * [Cluster Setup](#cluster-setup)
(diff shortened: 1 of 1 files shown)


krancour:
I've tried multiple times, with the same poor result each time.

Service instance provisioning and binding work as expected, but the Ghost container fails to start properly. These are the logs:

```
$ k logs quickstart-ghost-58f59d76ff-lkmms

Welcome to the Bitnami ghost container
Subscribe to project updates by watching https://github.com/bitnami/bitnami-docker-ghost
Submit issues and feature requests at https://github.com/bitnami/bitnami-docker-ghost/issues
Send us your feedback at containers@bitnami.com

nami    INFO  Initializing mysql-client
nami    INFO  mysql-client successfully initialized
nami    INFO  Initializing ghost
handler WARN  Using /opt/bitnami/ghost as working directory
mysql-c INFO  Trying to connect to MySQL server
mysql-c INFO  Found MySQL server listening at b987634c-39df-4197-bd05-f2a2ff61c44c.mysql.database.azure.com:3306
mysql-c INFO  MySQL server listening and working at b987634c-39df-4197-bd05-f2a2ff61c44c.mysql.database.azure.com:3306
ghost   INFO  ==> Configuring Database...
ghost   INFO  ==> Configuring Ghost...
Error executing 'postInstallation': Terminated
```

This looks more like an issue with the chart or the Ghost software itself than anything having to do with k8s Service Catalog or OSBA.

Should we rethink switching to Ghost?

Switch to Drupal instead?

krancour:
I've discovered that @jeremyrickard ran into the same issues. I don't think I can endorse modifying the quickstart to use Ghost. I'm closing this and #126. We'll have to come up with an alternative solution to #122 and we possibly need some fixes to the Ghost chart.

PR #133, opened 2017-12-06
Update to quick start to have a better command for opening wordpress

jeremyrickard:
After the install of the chart, use kubectl + JSONPath to find the
nodePort of the http port in the service definition.

Fixes: #125

### DIFF
--- a/docs/quickstart-minikube.md
+++ b/docs/quickstart-minikube.md
@@ -259,8 +259,7 @@ quickstart-wordpress   1         1         1            1           2m
 
 1. Run the following command to open WordPress in your browser:
     ```
-    open http://$(minikube ip):$(kubectl get service quickstart-wordpress \ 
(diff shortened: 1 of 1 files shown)


docs/quickstart-minikube.md
```diff
@@ -259,12 +259,13 @@ quickstart-wordpress   1         1         1            1           2m
 
 1. Run the following command to open WordPress in your browser:
     ```
-    open http://$(minikube ip):31215/admin
+    open http://$(minikube ip):$(kubectl get service quickstart-wordpress -o jsonpath={.spec.ports[?\(@.name==\"http\"\)].nodePort})
```
carolynvs: Did we lose the /admin path part?

jeremyrickard: GOOD CATCH!! I accidentally overwrote that bit. Updated the PR.

PR #137, opened 2017-12-07
Added Azure Container Registry Module

torosent:
Added ACR module


grenzr:
What happened to this? I'd quite like to see it added.

PR #33, opened 2017-10-31
Add HDInsight Module

arschles:
_This PR was originally opened by @zhongyi-zhang_

Closes https://github.com/Azure/azure-service-broker/issues/22

# Catalog

```
$ go run contrib/cmd/cli/*.go -H localhost -u username -P password catalog

service: azure-hdinsight      id: c0fba6e1-4ce4-4d93-b751-c8c5e337739c
   plan: Hadoop               id: fab80e5a-54c8-45e3-a466-f390de04e592
   plan: Spark                id: 9815959a-35d2-4bf7-b467-3e77c03dcc3e
   plan: HBase                id: 609c6d56-851e-41cf-8a71-2dde705cf5a5
   plan: Storm                id: ebd2dcf7-c586-42b7-8eeb-06e5641a34aa
```

# Hadoop

## Provision

```
$ go run contrib/cmd/cli/*.go -H localhost -u username -P password provision -sid c0fba6e1-4ce4-4d93-b751-c8c5e337739c -pid fab80e5a-54c8-45e3-a466-f390de04e592 --param location=eastus --poll

Provisioning service instance 3222e23c-3e0c-4f83-b8aa-fef3a97dd7f9

..................................................................................................................................................

Service instance 3222e23c-3e0c-4f83-b8aa-fef3a97dd7f9 has been successfully provisioned
```

## Bind

```
$ go run contrib/cmd/cli/*.go -H localhost -u username -P password bind -iid 3222e23c-3e0c-4f83-b8aa-fef3a97dd7f9

Binding 6be942bb-36e4-4665-8c61-d59a51ae1649 created for service instance 3222e23c-3e0c-4f83-b8aa-fef3a97dd7f9
Credentials:
   clusterEndpoint:     hd-e8d0c9a8-58e1-498b-a7b1-a370359c789f.azurehdinsight.net
   username:            qlcarmcs15
   password:            OayjFabsDMr3Ywls
   storageAccountName:  n55qabzavr
   storageAccountKey:   uOk8zKcHYRWOjUynZtvZK7JzuJ6dNaRqJED1xbbUE95kdqv3qZ8Sd5+HdAhmgpz0tU1hKzKmdvffWmV0steTKQ==
   blobStorageEndpoint: n55qabzavr.blob.core.windows.net
   blobStorageContainerName: vq03a3tpmd
```

## E2E Test

```
[...]

PR #141, opened 2017-12-12
Add HDInsight module

zhongyi-zhang:
Done a series of rebase and re-submit https://github.com/Azure/open-service-broker-azure/pull/33.


zhongyi-zhang:
@krancour I would like to follow up this PR. When will OSBA pay attention to add new modules? Anytime I can get it rebased.

zhongyi-zhang:
@krancour thanks, I got your points. In fact CF team did receive requests for Kafka (one of the cluster type HDInsight supported) service broker as Kafka is getting popular these years. But the intention were not strong. I supposed they are not sure they really wanted it, not sure they could build something making their deign work. So I think it's a good idea to add an experimental module to let them just have a try. And it wouldn't cost us much.
Someday you consider to have this module, no worries about the extra work for me. I am happy to share your burdens.

Issue #124, opened 2017-12-05
Spike: Expanded catalog of sqldb services


krancour:
It's clear (based on user feedback) that wrt sqldb, not everyone wants to provision a new server and a single database upon that server. (One can see where this could get expensive. One would also bump into their quota pretty quickly.)

@zhongyi-zhang (iirc) had anticipated this and built in support for the following:

- New SQL Servers can be created "out of band" by an administrator (e.g. using `az` CLI or the portal).
- Details of such servers are outlined in a JSON doc, the content of which can be set as an environment variable to the broker.
- Broker is restarted.
- Provisioning requests for sqldb can include an optional parameter that is a logical name that references the configuration for one of the existing SQL Servers.

The thing that is problematic about this is there is too much that is done "out of band," which undermines the goal of providing seamless, platform-native workflows for provisioning services.

This issue proposes an alternative to this approach that I'd like to explore.

First, note that these circumstances aren't isolated to sqldb. It's equally possible that users may wish to provision a Postgres or MySQL server and _then_ provision multiple databases upon that server. Storage accounts: One may wish to provision a storage account _then_ multiple containers within. Event hubs and service buses: One may wish to provision a namespace _then_ multiple hubs or buses within. Generalizing, there are many services that involve a hierarchy of resources wherein one might wish to provision and manage the lifecycle of some "parent" or "container" resource separately from multiple "subordinate" or "child" resources.

Note: Despite this generality, the sqldb module is the only one that acknowledges this today (even if the implementation is, possibly, not ideal).

Using sqldb as a guinea pig, I'd like to explore the following:

- Expand the services offered by the sqldb module from one to three:
    - "All in one": Same as the default behavior today. You get a new SQL Server and _one_ new database on that server. Plans of this service are all tiers of SQL Server.
    - "Server only": Provisions a SQL Server only-- no database. A required parameter will be a user-provided identifier that can later be used to reference this server. Plans of this service are all tiers of SQL Server.
    - "Database only": Provisions a database on an existing server. The server on which to provision the database is identified through a required parameter that references an existing server. There is only one plan for this service.

This might requires some architectural changes to the broker itself to make it start to understand parent/child relationships between service instances.

[...]

PR #142, opened 2017-12-12
WIP Refactoring sqldb to POC

jeremyrickard:
Removing the "existing server" logic from the current sqldb
module. This entails removing it from the provision/deprovision
file and removing it from the types. Also removed the config.go
file and updated the broker to not load that. We will need another
way to provide/register existing servers.

We will reimplement this behavior in a new service/service manager.

Begins POC for #124


zhongyi-zhang:
@jeremyrickard @krancour as our discussion in https://github.com/Azure/open-service-broker-azure/issues/124, do you already have a mature design to provide existing servers scenario which allows only the operator to manage the server? I am a little concerned about if it works for CF and if there is an approach to migrate from MASB.

krancour:
@zhongyi-zhang, @jeremyrickard and I are just experimenting. We're incrementally exploring what it takes to achieve something that will make everyone happy. Note that this PR is not targeted at master, but rather at another branch named `sqldb-catalog-poc`. Hopefully, from the name, it's evident that this is only a proof of concept. Nothing we do in this branch is going to be merged into master without some very serious consideration to how it affects use with CF.

zhongyi-zhang:
@krancour ah, thanks for the reminding! I didn't notice the branch. Good luck for your experiment!

krancour:
LGTM. Good start to what I know comes next. Will merge pending CI.

krancour:
One minor lint failure:

```
tests/lifecycle/mssql_cases_test.go:18:5:warning: unused variable or constant armTemplateNewServerBytes (varcheck)
tests/lifecycle/mssql_cases_test.go:18:5:warning: var armTemplateNewServerBytes is unused (U1000) (megacheck)
```

PR #144, opened 2017-12-13
[WIP] Adding support for CosmosDB as a storage backend

arschles:
This patch aims to add support for CosmosDB as a storage backend (note that it doesn't aim to replace the backend for the async queue processing engine - that may come in a follow-up).

It is currently a work in progress. Things to do:

- [ ] Add tests to the integration suite
- [x] Add a config parameter to the `main` function
    - If `main` sees that CosmosDB should be turned on, it should create a connection and pass it the cosmosDB storage backend

@jeremyrickard @krancour I think this would be a good addition to the broker. Do you foresee any problems here?

### DIFF
--- a/cmd/broker/broker.go
+++ b/cmd/broker/broker.go
@@ -96,6 +96,20 @@ func main() {
 		log.Fatal(err)
 	}
 
+	storageConfig, err := getStorageConfig()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	cosmosDBConfig := cosmosDBConfig{}
+	if storageConfig.StorageType == StorageTypeCosmosDB {
+		var err error
+		cosmosDBConfig, err = getCosmosDBConfig()
+		if err != nil {
+			log.Fatal(err)
+		}
+	}
+
 	// Create broker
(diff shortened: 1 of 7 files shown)


krancour:
I was actually thinking earlier today that this might be a nice change. I don't know that I would _ever_ opt for using Cosmos as the backing store for the async engine, just because the reliable queue pattern is something that Redis does _really_ well. (But I'm open to exploring it.)

All this said, there is another change to storage that I might like to get in ahead of this because it will eliminate many DRY exceptions. See #151.

arschles:
@krancour sounds good RE the async engine. I see that #153 fixed #151, so I'll fix conflicts.

Also, do we have a good way to test this? I haven't found anything like integration tests in here, but I'm probably missing something.

krancour:
As discussed, we're going to deprioritize any possibility of changing the underlying datastore until all the fundamental features are stabilized and we're reasonably sure the storage schema is one we can live with long term.

krancour:
Going to close this. Storage has evolved so much since this was opened that this would really need to be redone from scratch. fwiw, I do still think this is a good idea.

Issue #146, opened 2017-12-13
Firewall rules aren't specifiable


krancour:
At minimum, MySQL, PostgreSQL, and MS SQL are affected by this...

The ARM templates for each of those parameterize the start and end IPv4 addresses for a (single, currently) firewall rule, but we never actually collect those parameters from the inbound provisioning requests... and therefore never pass them through to the ARM templates.

These need to be specifiable because many customers will object to the default rule of 0.0.0.0 - 255.255.255.255.

Note also that the range 0.0.0.0 - 0.0.0.0 denotes all Azure _internal_ IPv4 addresses and would be a more sensible default than 0.0.0.0 - 255.255.255.255.

PR #152, opened 2017-12-15
Firewall rule parameterization

jeremyrickard:
Refactor to support allowing firewall rules.
Fixes: #146

Changes to suport optional parameters to allow specifying start
and end IP address for firewall rules on MySQL, PostgreSQL and
MSSQL. Previously, this defaulted to allow any connection. This
isn't ecure in practice, so allowing the customer to specify desired
rules. Also changed the default to 0.0.0.0 for both, which results in
only Azure internal addresses. Added some parameter validation
for the firewall values as well:

* Valid IPV4 addresses
* Start <= End

Extracted a method to build the arm template parameter map. Providing
empty strings is invalid for the ARM template, so only including
when non-empty.

Added parameters to the lifecycle tests to allow running them
from local machines (will need ao enhance the charts in helm-charts)

Added some tests.

### DIFF
--- a/pkg/services/mysqldb/provision.go
+++ b/pkg/services/mysqldb/provision.go
@@ -69,7 +69,7 @@ func (s *serviceManager) ValidateProvisioningParameters(
 	if bytes.Compare(startBytes, endBytes) > 0 {
 		return service.NewValidationError(
 			"firewallEndIPAddress",
-			fmt.Sprintf(`invalid value: "%s". firewallEndIPAddress must be 
+			fmt.Sprintf(`invalid value: "%s". must be 
 				greater than or equal to firewallStartIPAddress`, pp.FirewallIPEnd),
 		)
 	}
--- a/pkg/services/mysqldb/provision.go
+++ b/pkg/services/mysqldb/provision.go
@@ -36,13 +36,13 @@ func (s *serviceManager) ValidateProvisioningParameters(
 		if pp.FirewallIPStart == "" {
 			return service.NewValidationError(
 				"firewallStartIPAddress",
-				fmt.Sprintf(`invalid option: "%s"`, pp.FirewallIPStart),
+				"must be set when firewallEndIPAddress is set",
 			)
 		}
 		if pp.FirewallIPEnd == "" {
 			return service.NewValidationError(
 				"firewallEndIPAddress",
-				fmt.Sprintf(`invalid option: "%s"`, pp.FirewallIPEnd),
+				"must be set when firewallStartIPAddress is set",
 			)
 		}
 	}
@@ -50,14 +50,14 @@ func (s *serviceManager) ValidateProvisioningParameters(
 	if pp.FirewallIPStart != "" && startIP == nil {
 		return service.NewValidationError(
 			"firewallStartIPAddress",
-			fmt.Sprintf(`invalid option: "%s"`, pp.FirewallIPStart),
+			fmt.Sprintf(`invalid value: "%s"`, pp.FirewallIPStart),
 		)
 	}
 	endIP := net.ParseIP(pp.FirewallIPEnd)
 	if pp.FirewallIPEnd != "" && endIP == nil {
 		return service.NewValidationError(
 			"firewallEndIPAddress",
-			fmt.Sprintf(`invalid option: "%s"`, pp.FirewallIPEnd),
+			fmt.Sprintf(`invalid value: "%s"`, pp.FirewallIPEnd),
 		)
 	}
 	//The net.IP.To4 method returns a 4 byte representation of an IPv4 address.
@@ -69,7 +69,8 @@ func (s *serviceManager) ValidateProvisioningParameters(
 	if bytes.Compare(startBytes, endBytes) > 0 {
 		return service.NewValidationError(
 			"firewallEndIPAddress",
-			fmt.Sprintf(`invalid option: "%s"`, pp.SSLEnforcement),
+			fmt.Sprintf(`invalid value: "%s". firewallEndIPAddress must be 
+				greater than or equal to firewallStartIPAddress`, pp.FirewallIPEnd),
 		)
 	}
 	return nil
--- a/pkg/services/mysqldb/provision.go
+++ b/pkg/services/mysqldb/provision.go
@@ -29,42 +29,49 @@ func (s *serviceManager) ValidateProvisioningParameters(
 		sslEnforcement != "disabled" {
 		return service.NewValidationError(
 			"sslEnforcement",
-			fmt.Sprintf(`invalid sslEnforcement option: "%s"`, pp.SSLEnforcement),
+			fmt.Sprintf(`invalid option: "%s"`, pp.SSLEnforcement),
 		)
 	}
-
+	if pp.FirewallIPStart != "" || pp.FirewallIPEnd != "" {
+		if pp.FirewallIPStart == "" {
+			return service.NewValidationError(
+				"firewallStartIPAddress",
+				fmt.Sprintf(`invalid option: "%s"`, pp.FirewallIPStart),
+			)
+		}
+		if pp.FirewallIPEnd == "" {
+			return service.NewValidationError(
+				"firewallEndIPAddress",
+				fmt.Sprintf(`invalid option: "%s"`, pp.FirewallIPEnd),
+			)
+		}
+	}
 	startIP := net.ParseIP(pp.FirewallIPStart)
-	if pp.FirewallIPStart != "" && net.ParseIP(pp.FirewallIPStart) == nil {
+	if pp.FirewallIPStart != "" && startIP == nil {
 		return service.NewValidationError(
-			"firewallStartIpAddress",
-			fmt.Sprintf(`invalid firewallStartIpAddress option: "%s"`, pp.SSLEnforcement),
+			"firewallStartIPAddress",
+			fmt.Sprintf(`invalid option: "%s"`, pp.FirewallIPStart),
 		)
 	}
-
 	endIP := net.ParseIP(pp.FirewallIPEnd)
-	if pp.FirewallIPEnd != "" && net.ParseIP(pp.FirewallIPEnd) == nil {
+	if pp.FirewallIPEnd != "" && endIP == nil {
 		return service.NewValidationError(
-			"firewallEndIpAddress",
-			fmt.Sprintf(`invalid firewallEndIpAddress option: "%s"`, pp.SSLEnforcement),
+			"firewallEndIPAddress",
+			fmt.Sprintf(`invalid option: "%s"`, pp.FirewallIPEnd),
 		)
 	}
-
-	if startIP != nil && endIP == nil {
-		return service.NewValidationError(
-			"firewallEndIpAddress",
-			fmt.Sprintf(`invalid firewallEndIpAddress option: "%s"`, pp.SSLEnforcement),
-		)
-	}
-
+	//The net.IP.To4 method returns a 4 byte representation of an IPv4 address.
+	//Once converted,comparing two IP addresses can be done by using the
+	//bytes. Compare function. Per the ARM template documentation,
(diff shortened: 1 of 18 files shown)


pkg/services/mysqldb/provision.go
```diff
@@ -30,6 +32,39 @@ func (s *serviceManager) ValidateProvisioningParameters(
 			fmt.Sprintf(`invalid sslEnforcement option: "%s"`, pp.SSLEnforcement),
 		)
 	}
+
+	startIP := net.ParseIP(pp.FirewallIPStart)
+	if pp.FirewallIPStart != "" && net.ParseIP(pp.FirewallIPStart) == nil {
```
krancour: Suggestion-- don't do `net.ParseIP(pp.FirewallIPStart)`, because you've already done it above and assigned it to the `startIP` variable. Change this line to:

```
if pp.FirewallIPStart != "" && startIP == nil {
```

pkg/services/mysqldb/provision.go
```diff
@@ -30,6 +32,39 @@ func (s *serviceManager) ValidateProvisioningParameters(
...
+		return service.NewValidationError(
+			"firewallStartIpAddress",
+			fmt.Sprintf(`invalid firewallStartIpAddress option: "%s"`, pp.SSLEnforcement),
+		)
+	}
+
+	endIP := net.ParseIP(pp.FirewallIPEnd)
+	if pp.FirewallIPEnd != "" && net.ParseIP(pp.FirewallIPEnd) == nil {
```
krancour: Similar to earlier comment.

pkg/services/mysqldb/provision.go
```diff
@@ -30,6 +32,39 @@ func (s *serviceManager) ValidateProvisioningParameters(
 			fmt.Sprintf(`invalid sslEnforcement option: "%s"`, pp.SSLEnforcement),
 		)
 	}
+
+	startIP := net.ParseIP(pp.FirewallIPStart)
+	if pp.FirewallIPStart != "" && net.ParseIP(pp.FirewallIPStart) == nil {
+		return service.NewValidationError(
+			"firewallStartIpAddress",
+			fmt.Sprintf(`invalid firewallStartIpAddress option: "%s"`, pp.SSLEnforcement),
```
krancour: Field name is already included (in the previous parameter). Just say what was wrong with the field without naming it again.

pkg/services/mysqldb/provision.go
```diff
@@ -30,6 +32,39 @@ func (s *serviceManager) ValidateProvisioningParameters(
...
+		)
+	}
+
+	endIP := net.ParseIP(pp.FirewallIPEnd)
+	if pp.FirewallIPEnd != "" && net.ParseIP(pp.FirewallIPEnd) == nil {
+		return service.NewValidationError(
+			"firewallEndIpAddress",
+			fmt.Sprintf(`invalid firewallEndIpAddress option: "%s"`, pp.SSLEnforcement),
```
krancour: Field name is already included (in the previous parameter). Just say what was wrong with the field without naming it again.

pkg/services/mysqldb/provision.go
```diff
@@ -30,6 +32,39 @@ func (s *serviceManager) ValidateProvisioningParameters(
...
+		)
+	}
+
+	if startIP != nil && endIP == nil {
```
krancour: This seems to catch the scenario where a start ip was provided, but no end ip was provided. But the opposite scenario is equally a possibility and looks unaccounted for.

Maybe we can actually move the check for this higher up in this validation function (before we do any parsing) and just have it be that if one is defined, they _both_ must be?

pkg/services/mysqldb/provision.go
```diff
@@ -30,6 +32,39 @@ func (s *serviceManager) ValidateProvisioningParameters(
...
+		return service.NewValidationError(
+			"firewallEndIpAddress",
+			fmt.Sprintf(`invalid firewallEndIpAddress option: "%s"`, pp.SSLEnforcement),
+		)
+	}
+
+	startBytes := startIP.To4()
+	endBytes := endIP.To4()
+	if bytes.Compare(startBytes, endBytes) > 0 {
```
krancour: This is a clever trick. Can you just add a comment to explain how it works?

pkg/services/postgresqldb/provision.go
```diff
@@ -31,6 +33,39 @@ func (s *serviceManager) ValidateProvisioningParameters(
...
+
+	startBytes := startIP.To4()
+	endBytes := endIP.To4()
+	if bytes.Compare(startBytes, endBytes) > 0 {
+		return service.NewValidationError(
+			"firewallEndIpAddress",
+			fmt.Sprintf(`invalid firewallEndIpAddress option: "%s"`, pp.SSLEnforcement),
+		)
+	}
+
```
krancour: Similar comments to those already made for mysql validation.

pkg/services/sqldb/provision.go
```diff
@@ -33,6 +35,39 @@ func (s *serviceManager) ValidateProvisioningParameters(
...
+
+	startBytes := startIP.To4()
+	endBytes := endIP.To4()
+	if bytes.Compare(startBytes, endBytes) > 0 {
+		return service.NewValidationError(
+			"firewallEndIpAddress",
+			fmt.Sprintf(`invalid firewallEndIpAddress option: "%s"`, pp.FirewallIPEnd),
+		)
+	}
+
```
krancour: Similar comments to those already made for mysql validation.

pkg/services/mysqldb/provision.go
jeremyrickard: Yeah sorry! Was refactoring that and forgot to replace it. Thanks!

pkg/services/mysqldb/provision.go
jeremyrickard: Yep, will do.

pkg/services/mysqldb/provision.go
```diff
@@ -77,6 +112,39 @@ func (s *serviceManager) preProvision(
 	return pc, nil
 }
 
+func buildARMTemplateParameters(
```
krancour: +1 on breaking this out into its own function. Much more readable.

pkg/services/mysqldb/provision_test.go
```diff
@@ -0,0 +1,66 @@
...
+func TestValidateMissingEndFirewallConfig(t *testing.T) {
+
+	sm := &serviceManager{}
+
+	pp := &ProvisioningParameters{
+		FirewallIPStart: "192.168.86.1",
+	}
+
+	error := sm.ValidateProvisioningParameters(pp)
+	assert.NotNil(t, error)
```
krancour: Can we assert that the error is a `service.Validation` error and that the `Field` attribute of that error is the one we expected? Same comment applies for test cases below.

pkg/services/postgresqldb/provision_test.go
```diff
@@ -0,0 +1,66 @@
...
+func TestValidateMissingEndFirewallConfig(t *testing.T) {
+
+	sm := &serviceManager{}
+
+	pp := &ProvisioningParameters{
+		FirewallIPStart: "192.168.86.1",
+	}
+
+	error := sm.ValidateProvisioningParameters(pp)
+	assert.NotNil(t, error)
```
krancour: Again, check error type and field.

pkg/services/sqldb/provision_test.go
```diff
@@ -0,0 +1,66 @@
...
+func TestValidateMissingEndFirewallConfig(t *testing.T) {
+
+	sm := &serviceManager{}
+
+	pp := &ProvisioningParameters{
+		FirewallIPStart: "192.168.86.1",
+	}
+
+	error := sm.ValidateProvisioningParameters(pp)
+	assert.NotNil(t, error)
```
krancour: Check error type and field.

tests/lifecycle/mssql_cases_test.go
```diff
@@ -141,9 +141,12 @@ func getMssqlCases(
 			standardProvisioningContext: service.StandardProvisioningContext{
 				Location: "southcentralus",
 			},
-			provisioningParameters: &sqldb.ProvisioningParameters{},
-			bindingParameters:      &sqldb.BindingParameters{},
-			testCredentials:        testMsSQLCreds(),
+			provisioningParameters: &sqldb.ProvisioningParameters{
+				FirewallIPStart: "0.0.0.0",
+				FirewallIPEnd:   "255.255.255.0",
```
krancour: Why not `255.255.255.255`?

krancour:
Registered a few small comments, but overall, this is just what I was hoping for. Love it.

pkg/services/mysqldb/provision.go
```diff
@@ -27,7 +29,47 @@ func (s *serviceManager) ValidateProvisioningParameters(
...
 			"sslEnforcement",
-			fmt.Sprintf(`invalid sslEnforcement option: "%s"`, pp.SSLEnforcement),
+			fmt.Sprintf(`invalid option: "%s"`, pp.SSLEnforcement),
+		)
+	}
+	if pp.FirewallIPStart != "" || pp.FirewallIPEnd != "" {
+		if pp.FirewallIPStart == "" {
+			return service.NewValidationError(
+				"firewallStartIPAddress",
+				fmt.Sprintf(`invalid option: "%s"`, pp.FirewallIPStart),
```
krancour: Can we change the error text? Something that explains that when one of these fields is set, they both must be.

Same goes for the next check. Ditto for the other modules.

pkg/services/mysqldb/provision.go
```diff
@@ -27,7 +29,47 @@ func (s *serviceManager) ValidateProvisioningParameters(
...
+	if pp.FirewallIPEnd != "" && endIP == nil {
+		return service.NewValidationError(
+			"firewallEndIPAddress",
+			fmt.Sprintf(`invalid option: "%s"`, pp.FirewallIPEnd),
+		)
+	}
+	//The net.IP.To4 method returns a 4 byte representation of an IPv4 address.
+	//Once converted,comparing two IP addresses can be done by using the
+	//bytes. Compare function. Per the ARM template documentation,
+	//startIP must be <= endIP.
```
krancour: 👍

pkg/services/mysqldb/provision_test.go
```diff
@@ -0,0 +1,75 @@
...
+func TestValidateMissingEndFirewallConfig(t *testing.T) {
+	sm := &serviceManager{}
+	pp := &ProvisioningParameters{
+		FirewallIPStart: "192.168.86.1",
+	}
+	error := sm.ValidateProvisioningParameters(pp)
+	assert.NotNil(t, error)
+	v, ok := error.(*service.ValidationError)
+	assert.True(t, ok)
+	assert.Equal(t, v.Field, "firewallEndIPAddress")
```
krancour: 👍

pkg/services/mysqldb/provision.go
```diff
@@ -27,7 +29,47 @@ func (s *serviceManager) ValidateProvisioningParameters(
...
+		if pp.FirewallIPStart == "" {
+			return service.NewValidationError(
+				"firewallStartIPAddress",
+				fmt.Sprintf(`invalid option: "%s"`, pp.FirewallIPStart),
+			)
+		}
+		if pp.FirewallIPEnd == "" {
+			return service.NewValidationError(
+				"firewallEndIPAddress",
+				fmt.Sprintf(`invalid option: "%s"`, pp.FirewallIPEnd),
```
krancour: Can we change this to "invalid value?" Same in other spots.

krancour:
No remaining technical concerns. I made some comments about the wording of error messages. After that, LGTM.

krancour:
@jeremyrickard all LGTM. Thank you for all the effort that went into this. Just waiting for CI. Note I amended the PR just now with updated example manifests.

PR #153, opened 2017-12-16
make storage know how to deal with service-specific types

### LINKED ISSUE #151: Improve retrieval from storage / unmarshaling
There's a pattern that repeats many time in our code base to solve the following problem.

The problem is we retrieve instances or bindings as JSON from storage (Redis). We cannot initially unmarshal these objects in their entirety because until we look at the service ID, we don't know what concrete types to use for service-specific parameters, context, etc.

The solution that repeats looks something like this:

- Take a first pass at unmarshaling. Some field will not be unmarshaled at that time.
- Get the service ID
- Use the service ID to retrieve a service from the catalog
- The service points to a service manager
- The service manager can return empty, instances of appropriate concrete types-- service specific parameters, contexts, etc.
- These can be added to the existing, incomplete object
- A second pass is made at unmarshaling

Given how frequently this pattern repeats, what would be most ideal would be to make the storage layer implement this pattern itself so that it can be eliminated elsewhere. This will involve giving the storage layer access to the catalog so it can do lookups of services and service managers by service ID, but that shouldn't be a problem.

This change will _dramatically_ simplify code throughout the broker.

krancour:
Fixes #151 

There are probably other improvements that can follow on the heels of this one, but this moves us, incrementally, in a positive direction with fewer DRY exceptions.

### DIFF
--- a/pkg/api/bind.go
+++ b/pkg/api/bind.go
@@ -24,7 +24,7 @@ func (s *server) bind(w http.ResponseWriter, r *http.Request) {
 
 	log.WithFields(logFields).Debug("received binding request")
 
-	instance, ok, err := s.store.GetInstance(instanceID, nil, nil, nil)
+	instance, ok, err := s.store.GetInstance(instanceID)
 	if err != nil {
 		logFields["error"] = err
 		log.WithFields(logFields).Error(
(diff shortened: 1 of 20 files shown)


pkg/api/bind.go
```diff
@@ -24,7 +24,7 @@ func (s *server) bind(w http.ResponseWriter, r *http.Request) {
 
 	log.WithFields(logFields).Debug("received binding request")
 
-	instance, ok, err := s.store.GetInstance(instanceID, nil, nil, nil)
+	instance, ok, err := s.store.GetInstance(instanceID)
```
jeremyrickard: 👍

PR #155, opened 2017-12-18
kick off redis tests sooner

### LINKED ISSUE #148: Start Redis lifecycle tests sooner
Our lifecycle tests run up to ten test cases in parallel. (If we do more than that, we aren't likely to be able to do CI on multiple PRs in parallel without bumping into various quotas on our CI/CD Azure subscription.)

If I am counting correctly, Redis (which takes the longest to provision) is currently the tenth test case-- meaning it's just barely sneaking into the first wave. The test cases are currently ordered alphabetically by module. Addition of new modules (or additional test cases for existing modules) will bump Redis out of the first wave of tests and this will certainly have a deleterious effect on the total runtime of the lifecycle tests.

The Redis test cases should be kicked off _first_ so we're launching these long-running tests earlier, rather than on the tail end of the process.

krancour:
Fixes #148

### DIFF
--- a/tests/lifecycle/test_cases_test.go
+++ b/tests/lifecycle/test_cases_test.go
@@ -18,6 +18,7 @@ func getTestCases(resourceGroup string) ([]moduleLifecycleTestCase, error) {
 		armDeployer arm.Deployer,
 		resourceGroup string,
 	) ([]moduleLifecycleTestCase, error){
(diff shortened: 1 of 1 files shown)


krancour:
Nevermind... this is such a trivial change that I am merging this in good faith and taking responsibility for whatever happens.

PR #162, opened 2017-12-18
[WIP] Server VM only option for #124

jeremyrickard:
@krancour, here is a WIP for adding new service for server vm only sqldb. Looking for feedback on this approach.

Summary
- Refactor of sqldb/ to support multiple services/service managers
-  Two service managers for sqldb now: 
   - One for the All In One service
   - One for the server VM only
   - I embedded the existing serviceManager in both to reuse bits
- Added new Service to sqldb catalog (server vm only)
   - Not bindable
- Added GetBindable() to Service interface
- Added check to see if Service is bindable in lifecycle tests (skip if so). 

The new plan is not bindable, as it's VM only and not a database itself. It didn't appear that we've done this yet (non-bindable), so I did a small refactor to the services/catalog.go to expose a GetBindable() method on the Service interface and used this in the service lifecycle tests to not do the binding/unbinding tests. On a related note, should we error in the bind/unbind methods for services that do not support it or simply return empty? The spec doesn't indicate what to do when something [isn't bindable](https://github.com/openservicebrokerapi/servicebroker/blob/v2.13/spec.md#binding). That's a WIP. 

Last question: As you can see from this change, Iv've added service specific Provision/Bind/Unbind methods in the respective files. I'm wondering if we want define a pattern for modules with multiple services and different logic between the services? Having the single provision.go/etc makes it easy to find where something is done, but I feel like maybe having separate files for each might be a better way?


pkg/service/catalog.go
```diff
@@ -206,6 +207,11 @@ func (s *service) GetName() string {
 	return s.Name
 }
 
+// GetPlans returns all of the service's plans
```
krancour: Comment looks wrong...

pkg/services/sqldb/bind.go
```diff
@@ -16,6 +16,15 @@ func (s *serviceManager) ValidateBindingParameters(
...
 
+//TODO: What behavior do we want for bind on a non-bindable service.
+//Appropriate error?
+func (s *serverOnlyServiceManager) Bind(
+	instance service.Instance,
+	_ service.BindingParameters,
+) (service.BindingContext, service.Credentials, error) {
+	return nil, nil, nil
+}
+
```
krancour: This is fine for now-- we'll just no-op it. We can figure out how to deal with this more elegantly later.

pkg/services/sqldb/mssql.go
```diff
@@ -7,25 +7,41 @@ import (
...
+	serverOnlyServiceManager *serverOnlyServiceManager
 }
 
 type serviceManager struct {
 	armDeployer  arm.Deployer
 	mssqlManager mssql.Manager
 }
 
+type allInOneServiceManager struct {
+	serviceManager
```
krancour: Does this result in us having to further qualify the access path to get down to an `armDeployer` or `mssqlManager`? i.e. Do we have to do `s. serviceManager. armDeployer` where we could previously do just `s.armDeployer`?

jeremyrickard: No. I refactored in the sqldb/provision.go like so:

In provision.go, now we have two methods that take care of building up proper parameters and determining which arm template to use. these are receivers for the two new service managers. 
>func (s *serverOnlyServiceManager) deployARMTemplate(....)
>func (a *allInOneServiceManager) deployARMTemplate(

I refactored the old logic about actually calling the arm deployer like so:
>func (s *serviceManager) deployTemplate(......)

Through embedding, this applies to both and it's getting the old service manager. It also gets the arm template + the params as method parameters.  It's fairly generic now and allows a little more DRY in there and can be used by these two + our third (db only) and any other.

pkg/services/sqldb/bind.go
```diff
@@ -16,6 +16,15 @@ func (s *serviceManager) ValidateBindingParameters(
...
 func (s *serviceManager) Bind(
```
krancour: Looks like the receiver should be updated on this because the functionality is specific to the `allInOneServiceManager`.

pkg/service/catalog.go
```diff
@@ -206,6 +207,11 @@ func (s *service) GetName() string {
...
+func (s *service) GetBindable() bool {
+	return s.Bindable
+}
```
krancour: 👍 This will be useful.

pkg/services/sqldb/bind.go
jeremyrickard: I originally had changed it, but reverted it back  because it should work out of the box for the db only mechanism as well. I went with providing new implementations when needed and leaving the base case when it would cover multiple, but I think it would be more explict to change.

pkg/service/catalog.go
jeremyrickard: Yeah, everything had just been bindable so far, but I think that will change in the future and this made sense to me. I caught my issue with the Bind() method on the server only case because of the tests. I assumed if it was non-bindable it wouldn't try and bind/unbind, but it was so I plumbed this in.  

Changed the comment locally and haven't pushed.

pkg/services/sqldb/mssql.go
krancour: Ah... wasn't sure what the benefit to this was, at first, but now I see that there are spots where you've defined functions required by the `ServiceManager` interface with `*serviceManager` (the "base class" if this were OO) as the receiver. `GetDeprovisioner(...)` is one such example...

I'm not sure how I feel about this, because anywhere we do that, we're going to require that the types those functions operate on are common to the multiple `ServiceManager` implementations in question-- and that might not be ideal, because the service-specific details are still going to vary from one `Service` to the next...

For example, I see that this context type is common to both serviceManagers:

```
type mssqlProvisioningContext struct {
	ARMDeploymentName          string `json:"armDeployment"`
	ServerName                 string `json:"server"`
	AdministratorLogin         string `json:"administratorLogin"`
	AdministratorLoginPassword string `json:"administratorLoginPassword"`
	DatabaseName               string `json:"database"`
	FullyQualifiedDomainName   string `json:"fullyQualifiedDomainName"`
}
```

While we _can_ make that work, it's a little awkward.

What do you think about separating _all_ the types and functions so that all are unique to a given `ServiceManager` implementation?

Eventually, where I think we'll land is that _after_ we implement the "database only" service, we'll have a context type for server only and another context type for database only and there won't be much, if any overlap between the two. _Then_ we can maybe make the "all in one" context be a composition of the other two.

jeremyrickard: Yeah, I agree. For the provisoner case, they were mostly similar (except that database), but like I mentioned in a comment on the Bind method, I think more explicit is better in the end and once we implement that database only step, it will for sure be a composition of the two. I’ll decompose more.

pkg/services/sqldb/bind.go
krancour: I think we might be better off keeping types and functions for different `ServiceManager` impls completely separate from one another. Main reason-- as I'd cited in another comment-- is that sharing bits of functionality between `ServiceManagers` in this way forces those `ServiceManagers` into using types that are common to both-- and often types, like contexts, will have fields that vary significantly by service.

When we inevitably end up having to DRY things up due to this approach, we can just have the discreet functions used by each `ServiceManager` delegate to common helper functions.

krancour:
I've reviewed about half of this. It looks pretty good so far, but the comments I've made so far might give a general sense of where I think we should make some changes. Slack me and lmk what you think.

jeremyrickard:
@krancour handled the comments above. Not sure I like the naming of the service mangers, they are kind of wordy. I made an attempt to shrink them but I think they are still too wordy. Take a look and let me know what you think.

jeremyrickard:
I also plumbed in stubs for the dB only

jeremyrickard:
Going to close this and redo the work in a cleaner PR.

PR #163, opened 2017-12-19
s/context/details/g

### LINKED ISSUE #150: s/context/details/
For both instances and bindings, it seems "details" (e.g. "instance details" or "binding details") would be clearer language than "context" (e.g. "provisioning context" or "binding context"). If this simple change in language makes the code easier to comprehend, it's a win.

I'm prioritizing this change _only_ because it's a breaking change to the storage schema and I want to handle any such changes we know about sooner than later, while we're still in an alpha state and allowed to break things.

krancour:
Fixes #150 

This might present a challenging merge for the ongoing work in the #124 feature branch, but I'd prefer to see _these_ changes merged sooner than later because they include breaking changes to the storage schema.

Besides addressing #150, this contains a related change that flattens the "standard provisioning parameters / context" onto the instances. This means `location`, `resourceGroup`, and `tags` are now directly part of the instance and not held in some subordinate object.

Collectively, these changes make the data model much simpler to comprehend.


pkg/api/provision.go
```diff
@@ -116,41 +116,61 @@ func (s *server) provision(w http.ResponseWriter, r *http.Request) {
...
+	location := ""
+	locIface, ok := provisioningRequest.Parameters["location"]
+	if ok {
+		location, ok = locIface.(string)
+		if !ok {
+			logFields["location"] = locIface
+			log.WithFields(logFields).Debug(
+				"bad provisioning request: location parameter was not a string",
+			)
+			s.writeResponse(w, http.StatusBadRequest, responseInvalidPlanID)
```
krancour: Whoops... the body of the response is wrong. I will fix this.

jeremyrickard:
I like the Context -> Details change. Easier to understand things.

Issue #100, opened 2017-12-01
Create dockerhub repository for 'open-service-broker-azure'


arschles:
After https://github.com/Azure/open-service-broker-azure/pull/91, the name of this project will have changed from `azure-service-broker' to 'open-service-broker-azure'. 

We need to create the newly named repo and update to push to it. We'll also need to change the chart in [Azure/helm-charts](https://github.com/azure/helm-charts) to point to the image in the new repository.

cc/ @jeremyrickard

krancour:
Deferring this since we don't yet have the repo name we wanted. i.e. Just removing it from the v0.5.0 milestone.

krancour:
I'm closing this for now. We'll re-open if we need to.

PR #165, opened 2017-12-19
Updating docker image name

jeremyrickard:
Migrates name from azure-service-broker to azure-open-service-broker

Fixes: #100

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,11 +3,7 @@ GIT_VERSION = $(shell git describe --always --abbrev=7 --dirty)
 BINARY_DIR := bin
 BINARY_NAME := osba
 
-# This is left as 'azure-service-broker' because we don't yet have a docker repo
-# for 'open-service-broker-azure'
-#
-# See https://github.com/Azure/open-service-broker-azure/issues/100
-BASE_IMAGE_NAME        = azure-service-broker
+BASE_IMAGE_NAME        = azure-open-service-broker
 
 RC_IMAGE_NAME          = $(REGISTRY)$(BASE_IMAGE_NAME):$(GIT_VERSION)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -3,11 +3,7 @@ GIT_VERSION = $(shell git describe --always --abbrev=7 --dirty)
...
 BINARY_NAME := osba
 
-# This is left as 'azure-service-broker' because we don't yet have a docker repo
-# for 'open-service-broker-azure'
-#
-# See https://github.com/Azure/open-service-broker-azure/issues/100
-BASE_IMAGE_NAME        = azure-service-broker
+BASE_IMAGE_NAME        = azure-open-service-broker
 
 RC_IMAGE_NAME          = $(REGISTRY)$(BASE_IMAGE_NAME):$(GIT_VERSION)
```
carolynvs: Where did we finally land on setting `REGISTRY`? In circle?

carolynvs: Nevermind just saw your other PR!

krancour:
idk why naming things has been such a challenge for us. I hate having to ask this...

The name we'd _wanted_ for the Docker Hub repo would be one that matches the name of this GitHub repo: `open-service-broker-azure`.

But, after a lot of back and forth, the name they gave us was `azure-open-service-broker`...

Since this isn't the name we asked for, I have to ask whether there's any _actual_ benefit of switching from the old Docker Hub repo name (`azure-service- broker`) to this...

They're both wrong. So should we keep the nicer of the two until we get the name we wanted?

seanmck:
Agree with @krancour. Hold off on this for now - I didn't have a chance to engage much on this topic in the run-up to KubeCon. Will connect with Steve and see if we can get alignment.

PR #166, opened 2017-12-19
Add REGISTRY variable to makefile

### LINKED ISSUE #145: Master build is failing
I noticed [this build](https://circleci.com/gh/Azure/open-service-broker-azure/700) seems to be failing on the docker push.

jeremyrickard:
Added a Registry variable to the Makefile. It was used in
image names, but never defined. This caused the master build to fail.

Fixes #145

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,7 @@ BINARY_NAME := osba
 #
 # See https://github.com/Azure/open-service-broker-azure/issues/100
 BASE_IMAGE_NAME        = azure-service-broker
+REGISTRY               = microsoft/
 
 RC_IMAGE_NAME          = $(REGISTRY)$(BASE_IMAGE_NAME):$(GIT_VERSION)
 RC_MUTABLE_IMAGE_NAME  = $(REGISTRY)$(BASE_IMAGE_NAME):canary


Makefile
```diff
@@ -8,6 +8,7 @@ BINARY_NAME := osba
 #
 # See https://github.com/Azure/open-service-broker-azure/issues/100
 BASE_IMAGE_NAME        = azure-service-broker
+REGISTRY               = microsoft/
```
carolynvs: nit: I have always defined the components of a fully qualified docker image reference this way:

```
REGISTRY = "quay.io" # leave blank to use the default registry defined in the docker daemon
IMAGE = "carolynvs/lovespuppies"
TAG = "v1"
FULL = (REGISTRY)(IMAGE):(TAG)
```

I realize what you have (setting the registry to a portion of the image name) works, but it seems a bit non-standard?

krancour: @carolynvs ya... it's a little off, but it's a pattern we've used successfully on many projects now.

The thing we called `REGISTRY` is, actually "registry and (potentially) repo."

Left undefined (which was a deliberate default), the image name ends up being an unqualified `azure-service-broker:<version>`-- which was what I'd wanted.

If `REGISTRY` is set to `dockerhub.io/Microsoft/`, for instance, we end up with fully qualified image names like `dockerhub.io/Microsoft/azure-service-broker:<version>`

All that said, the change above is wrong. @jeremyrickard and I have already discussed and we're going to close this one. He'll set the `REGISTRY` var in the Circle config instead.

PR #167, opened 2017-12-19
Updating CircleCI to define DOCKER_REPO env var

jeremyrickard:
We previously set the REGISTRY environment variable in a deploy script.
When that was removed to streamline the release process, we need to
define it in Circle. Added two environment blocks to the publish-rc-images
and publish-release-images jobs.

Fixes: 145

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -82,7 +82,7 @@ jobs:
     working_directory: ~/go/src/github.com/Azure/open-service-broker-azure
     environment:
       GOPATH: ~/go
-      REGISTRY: docker.io/microsoft/
+      DOCKER_REPO: microsoft/
     machine: true
     steps:
       - checkout
@@ -97,7 +97,7 @@ jobs:
           command: make docker-push-rc
   publish-release-images:
     environment:
-      REGISTRY: docker.io/microsoft/
+      DOCKER_REPO: microsoft/
     machine: true
     steps:
       - checkout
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -82,6 +82,7 @@ jobs:
     working_directory: ~/go/src/github.com/Azure/open-service-broker-azure
     environment:
       GOPATH: ~/go
+      REGISTRY: docker.io/microsoft/
```
krancour: We can actually drop the leading `docker.io`, because that's the default registry for Docker anyway.

At that point, what we have is `REGISTRY: microsoft` and that looks funny, because the value doesn't look like a registry. It looks like a _repository_. This is a smell that's telling me we got something wrong here in the first place.

This is a pattern we borrowed from older projects. I don't want to blindly follow it if it doesn't make sense.

Through a conversation with @carolynvs, I think I've come to realize that we should rename this env var as `DOCKER_REPO`. That way an undefined value, `microsoft/` or `$(docker-machine ip my-docker):5000/` all equally make sense-- as even though some contain a host/ip name and some do not, they consistently all do point to _a repository_.

So... can I ask that we switch from `REGISTRY` to `DOCKER_REPO` and make all requisite changes to the `Makefile` as well?

The end result will be much cleaner and easier to understand than where we stand today.

jeremyrickard: SGTM. I'll update the Makefile as well.

carolynvs: I like it!

krancour:
LGTM

PR #172, opened 2017-12-20
Adding user-agent string to requests to ARM

### LINKED ISSUE #170: Use an OSBA-specific user-agent string for calls to ARM RPs
The latest recommendation from the BI v-team is to use user-agent for tracking the provenance of client-side requests to Azure RPs. Unlike tags, it is already present in the ARM Kusto data store, which makes it easy to wire up with PowerBI.

Suggest we do something like:

open-service-broker-azure/version

arschles:
Fixes #170
Requires #173 

This PR requires that we add a version string that is injected at compile-time. I'll do that in another PR and link it to this one.

After #173 is in, this PR needs to pass the version & commit down the stack to the `github.com/Azure/open-service-broker-azure/pkg/azure/arm` package


krancour:
Now requires #185 to be merged first as well.

PR #173, opened 2017-12-20
Adding build flags to inject the version and commit

arschles:
This patch injects the version and commit to the OSBA binary. If there is no version set, then sets the version to “devel”

This needs to go in before #172

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -237,19 +237,16 @@ CLI_BINARY_NAME := broker-cli
 build-mac-broker-cli: check-docker-compose
 	docker-compose run --rm -e GOOS=darwin -e GOARCH=amd64 dev \
 		go build -o ${CONTRIB_BINARY_DIR}/${CLI_BINARY_NAME} \
-		-ldflags '$(LDFLAGS)' \
 		./contrib/cmd/cli
 
 .PHONY: build-linux-broker-cli
 build-linux-broker-cli: check-docker-compose
 	docker-compose run --rm -e GOOS=linux -e GOARCH=amd64 dev \
 		go build -o ${CONTRIB_BINARY_DIR}/${CLI_BINARY_NAME} \
-		-ldflags '$(LDFLAGS)' \
 		./contrib/cmd/cli
 
 .PHONY: build-win-broker-cli
 build-win-broker-cli: check-docker-compose
 	docker-compose run --rm -e GOOS=windows -e GOARCH=amd64 dev \
 		go build -o ${CONTRIB_BINARY_DIR}/${CLI_BINARY_NAME} \
-		-ldflags '$(LDFLAGS)' \
 		./contrib/cmd/cli
--- a/Makefile
+++ b/Makefile
@@ -16,10 +16,12 @@ REL_IMAGE_NAME         = $(DOCKER_REPO)$(BASE_IMAGE_NAME):$(REL_VERSION)
 REL_MUTABLE_IMAGE_NAME = $(DOCKER_REPO)$(BASE_IMAGE_NAME):latest
 
 ifeq ($(REL_VERSION),)
-REL_VERSION="devel"
+BROKER_VERSION="devel"
+else
+BROKER_VERSION=$(REL_VERSION)
 endif
 
-LDFLAGS = -w -X main.commit=$(COMMIT) -X main.commit=$(GIT_VERSION) -X main.version=$(REL_VERSION)
+LDFLAGS = -w -X main.commit=$(COMMIT) -X main.commit=$(GIT_VERSION) -X main.version=$(BROKER_VERSION)
 
 # Checks for the existence of a docker client and prints a nice error message
 # if it isn't present
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ else
 BROKER_VERSION=$(REL_VERSION)
 endif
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -229,16 +235,19 @@ CLI_BINARY_NAME := broker-cli
 build-mac-broker-cli: check-docker-compose
 	docker-compose run --rm -e GOOS=darwin -e GOARCH=amd64 dev \
 		go build -o ${CONTRIB_BINARY_DIR}/${CLI_BINARY_NAME} \
+		-ldflags '$(LDFLAGS)' \
```
krancour: Not sure we need to set this on the CLI builds.

Makefile
```diff
@@ -15,6 +15,12 @@ RC_MUTABLE_IMAGE_NAME  = $(DOCKER_REPO)$(BASE_IMAGE_NAME):canary
 REL_IMAGE_NAME         = $(DOCKER_REPO)$(BASE_IMAGE_NAME):$(REL_VERSION)
 REL_MUTABLE_IMAGE_NAME = $(DOCKER_REPO)$(BASE_IMAGE_NAME):latest
 
+ifeq ($(REL_VERSION),)
+REL_VERSION="devel"
```
krancour: We should use a new variable for this and not re-use `REL_VERSION`. The `docker-push-release` target (used for pushing official, semantically versioned releases) checks whether this var is set and fails if it isn't. Now that check will always pass and the value that we previously expected to always be a semantic version string could now be "devel" instead.

The approach is fine... just use a separate variable.

Makefile
```diff
@@ -15,6 +15,14 @@ RC_MUTABLE_IMAGE_NAME  = $(DOCKER_REPO)$(BASE_IMAGE_NAME):canary
 REL_IMAGE_NAME         = $(DOCKER_REPO)$(BASE_IMAGE_NAME):$(REL_VERSION)
 REL_MUTABLE_IMAGE_NAME = $(DOCKER_REPO)$(BASE_IMAGE_NAME):latest
 
+ifeq ($(REL_VERSION),)
+BROKER_VERSION="devel"
+else
+BROKER_VERSION=$(REL_VERSION)
+endif
+
+LDFLAGS = -w -X main.commit=$(COMMIT) -X main.commit=$(GIT_VERSION) -X main.version=$(BROKER_VERSION)
```
krancour: `-X main.commit=$(COMMIT)` doesn't look right... it seems to write the same property as `-X main.commit=$(GIT_VERSION)`, and appears to use an undefined value. Guessing this was a copy/paste error.

Makefile
krancour: Still think we don't need these flags here.

Makefile
```diff
@@ -229,16 +237,19 @@ CLI_BINARY_NAME := broker-cli
...
 	docker-compose run --rm -e GOOS=darwin -e GOARCH=amd64 dev \
 		go build -o ${CONTRIB_BINARY_DIR}/${CLI_BINARY_NAME} \
+		-ldflags '$(LDFLAGS)' \
 		./contrib/cmd/cli
 
 .PHONY: build-linux-broker-cli
 build-linux-broker-cli: check-docker-compose
 	docker-compose run --rm -e GOOS=linux -e GOARCH=amd64 dev \
 		go build -o ${CONTRIB_BINARY_DIR}/${CLI_BINARY_NAME} \
+		-ldflags '$(LDFLAGS)' \
```
krancour: Don't need these flags here.

Makefile
```diff
@@ -229,16 +237,19 @@ CLI_BINARY_NAME := broker-cli
...
 		./contrib/cmd/cli
 
 .PHONY: build-win-broker-cli
 build-win-broker-cli: check-docker-compose
 	docker-compose run --rm -e GOOS=windows -e GOARCH=amd64 dev \
 		go build -o ${CONTRIB_BINARY_DIR}/${CLI_BINARY_NAME} \
+		-ldflags '$(LDFLAGS)' \
```
krancour: Don't need these flags here.

krancour:
Can we also make the broker print the version at startup so we know this is working as intended?

Makefile
arschles: Even though the CLI is not supposed to be a first-class citizen, it couldn't hurt, right?

arschles:
@krancour I updated/addressed comments. PTAL

cmd/broker/broker.go
```diff
@@ -49,6 +54,7 @@ func init() {
 }
 
 func main() {
+	log.Printf("Broker starting with version %s, commit %s", version, commit)
```
krancour: This will yield an info level log message with version and commit as k/v pairs, easily digestible by any log aggregator. This would be more consistent with most other log messages from OSBA:

```
log.WithFields(
  log.Fields{
    "version": version,
    "commit":  commit,
  },
).Info("Open Service Broker for Azure starting")
```

Makefile
krancour: I'm fine with it. Should this PR include the necessary modification to `contrib/cmd/cli`?

arschles: Repeating what @krancour and I talked about offline: I am going to remove this and submit a low-priority issue to:

- Add the flag back
- Implement the `version` command

arschles:
@krancour I've submitted https://github.com/Azure/open-service-broker-azure/issues/177 and removed the loader flags from the CLI builds. PTAL

krancour:
LGTM. Thanks @arschles!

PR #174, opened 2017-12-20
quickstart fixes

krancour:
Fixes critical issues with the quickstart, which we discussed in Slack.

### DIFF
--- a/docs/quickstart-minikube.md
+++ b/docs/quickstart-minikube.md
@@ -123,19 +123,6 @@ First let's identify your Azure subscription and save it for use later on in the
     > $env:AZURE_SUBSCRIPTION_ID = "<SubscriptionId>"
     ```
 
-## Create a resource group
-We are using a resource group to isolate all the resources created in this quickstart
-for easy cleanup later.
-
-1. List the available Azure regions and select a region, for example `centralus`:
-    ```
-    $ az account list-locations -o table
-    ```
-1. Create a resource group for the quickstart:
-    ```
-    $ az group create --name osba-quickstart --location <RegionName>
-    ```
-
 ## Create a service principal
 This creates an identity for Open Service Broker for Azure to use when provisioning
 resources on your account on behalf of Kubernetes.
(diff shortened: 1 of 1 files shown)


docs/quickstart-minikube.md
```diff
@@ -268,18 +255,51 @@ quickstart-wordpress   1         1         1            1           2m
...
+following command returns no resources, deprovisioning is complete:
+
+```
+$ kubectl get serviceinstances -n osba-quickstart
+No resources found.
+```
+
+## Optional: Further Cleanup
+
+At this point, the Azure MySQL database should have been fully deprovisioned.
```
carolynvs: nit: avoid parentheticals

```
To ensure that you are not billed for unused resources, delete the Azure resource group
that contained the database. The name of the resource group is the namespace used in
the `helm install` command.

File: docs/quickstart-minikube.md [hunk 2]
```diff
@@ -231,7 +218,7 @@ WordPress to Kubernetes and OSBA will handle provisioning an Azure MySQL databas
 and binding it to our WordPress installation.
 
 ```
-$ helm install azure/wordpress --name quickstart
+$ helm install azure/wordpress --name osba-quickstart --namespace osba-quickstart
```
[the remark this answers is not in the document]
krancour: > because helm will set the resource group to the same name as namespace

Practically speaking yes. But that's a feature (or quirk) of _our charts_ specifically.

> I'm not clear on why you changed the name of the wordpress installation though? Was there something wrong with quickstart? Do they have to match?

They don't have to, but I took a guess that requiring readers to mentally juggle "quickstart" (the release) and "osba-quickstart" (the namespace and, implicitly, the resource group) would probably lead to some mistakes.

docs/quickstart-minikube.md
krancour: Will fix.

carolynvs:
Much better! I'm so glad you caught this!

PR #176, opened 2017-12-27
Decomposition of single service into multiple services

jeremyrickard:
1.) This renames the existing service manager into a service to
    implement the all in one scenario (existing)
2.) Next, two new services are added (vm only and db only)
3.) Each lifecycle operation go file (bind, deprovision, provision, unbind)
    now declares a method for each service manager struct
4.) Added new services to the catalog
5.) DB only option is plumbed in, but won't work as of yet (ref instance)
6.) Refactored catalog/service interface to add GetBindable() method
7.) Modified tests to only execute binding related tests if plan
    is bindable

I'm wondering if it would make sense to decompose the lifecycle go files. Right now, we are violating DRY for these, so might also be good to identify commonality and refactor the Provision Params / Instance Details to use embedding to get some reuse.

### DIFF
--- a/pkg/services/sqldb/deprovision.go
+++ b/pkg/services/sqldb/deprovision.go
@@ -24,6 +24,10 @@ func (v *vmOnlyManager) GetDeprovisioner(
 ) (service.Deprovisioner, error) {
 	return service.NewDeprovisioner(
 		service.NewDeprovisioningStep("deleteARMDeployment", v.deleteARMDeployment),
+		service.NewDeprovisioningStep(
+			"deleteMsSQLServer",
+			v.deleteMsSQLServer,
+		),
 	)
 }
 
@@ -126,11 +130,32 @@ func (a *allInOneManager) deleteMsSQLServer(
 	return dt, nil
 }
 
+func (v *vmOnlyManager) deleteMsSQLServer(
+	_ context.Context,
+	instance service.Instance,
+	_ service.Plan,
+	_ service.Instance,
+) (service.InstanceDetails, error) {
+	dt, ok := instance.Details.(*mssqlVMOnlyInstanceDetails)
+	if !ok {
+		return nil, fmt.Errorf(
+			"error casting instance.Details as *mssqlInstanceDetails",
+		)
+	}
+	if err := v.mssqlManager.DeleteServer(
+		dt.ServerName,
+		instance.ResourceGroup,
+	); err != nil {
+		return dt, fmt.Errorf("error deleting mssql server: %s", err)
+	}
+	return dt, nil
+}
+
 func (d *dbOnlyManager) deleteMsSQLDatabase(
 	_ context.Context,
 	instance service.Instance,
 	_ service.Plan,
-	referenceInstance service.Instance, // Reference instance
+	referenceInstance service.Instance,
 ) (service.InstanceDetails, error) {
 	dt, ok := instance.Details.(*mssqlInstanceDetails)
 	if !ok {
--- a/pkg/services/sqldb/deprovision.go
+++ b/pkg/services/sqldb/deprovision.go
@@ -7,31 +7,51 @@ import (
 	"github.com/Azure/open-service-broker-azure/pkg/service"
 )
 
-func (s *serviceManager) GetDeprovisioner(
+func (a *allInOneManger) GetDeprovisioner(
 	service.Plan,
 ) (service.Deprovisioner, error) {
 	return service.NewDeprovisioner(
-		service.NewDeprovisioningStep("deleteARMDeployment", s.deleteARMDeployment),
+		service.NewDeprovisioningStep("deleteARMDeployment", a.deleteARMDeployment),
 		service.NewDeprovisioningStep(
 			"deleteMsSQLServerOrDatabase",
-			s.deleteMsSQLServerOrDatabase,
+			a.deleteMsSQLServerOrDatabase,
 		),
 	)
 }
 
-func (s *serviceManager) deleteARMDeployment(
+func (v *vmOnlyManager) GetDeprovisioner(
+	service.Plan,
+) (service.Deprovisioner, error) {
+	return service.NewDeprovisioner(
+		service.NewDeprovisioningStep("deleteARMDeployment", v.deleteARMDeployment),
+	)
+}
+
+func (d *dbOnlyManager) GetDeprovisioner(
+	service.Plan,
+) (service.Deprovisioner, error) {
+	return service.NewDeprovisioner(
+		service.NewDeprovisioningStep("deleteARMDeployment", d.deleteARMDeployment),
+		service.NewDeprovisioningStep(
+			"deleteMsSQLServerOrDatabase",
+			d.deleteMsSQLServerOrDatabase,
+		),
+	)
+}
+
+func (a *allInOneManger) deleteARMDeployment(
 	_ context.Context,
 	instance service.Instance,
 	_ service.Plan,
 	_ service.Instance, // Reference instance
 ) (service.InstanceDetails, error) {
-	dt, ok := instance.Details.(*mssqlInstanceDetails)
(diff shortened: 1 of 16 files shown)


pkg/services/sqldb/deprovision.go
```diff
@@ -7,19 +7,39 @@ import (
...
 		),
 	)
 }
 
-func (s *serviceManager) deleteARMDeployment(
+func (v *vmOnlyManager) GetDeprovisioner(
+	service.Plan,
+) (service.Deprovisioner, error) {
+	return service.NewDeprovisioner(
+		service.NewDeprovisioningStep("deleteARMDeployment", v.deleteARMDeployment),
```
krancour: This will still need a step for deleting the server (same as in the all-in-one scenario).

pkg/services/sqldb/provision.go
```diff
@@ -95,16 +176,96 @@ func (s *serviceManager) preProvision(
...
+	dt.AdministratorLogin = generate.NewIdentifier()
+	dt.AdministratorLoginPassword = generate.NewPassword()
+	return dt, nil
+}
+
+func (d *dbOnlyManager) preProvision(
+	_ context.Context,
+	instance service.Instance,
+	_ service.Plan,
+	referenceInstance service.Instance, //reference instance
```
krancour: Nit: Remove the comment. The purpose of the argument is clear from its name. Let's just include such a comment where we've used `_` to express our disinterest in an argument and don't want to lose track of what that arg is for, in case we wish to use it in the future.

krancour: Same feedback applies in the other provisioning steps used by this manager.

pkg/services/sqldb/provision.go
```diff
@@ -95,16 +176,96 @@ func (s *serviceManager) preProvision(
...
+) (service.InstanceDetails, error) {
+	dt, ok := instance.Details.(*mssqlInstanceDetails)
+	if !ok {
+		return nil, errors.New(
+			"error casting instance.Details as *mssqlInstanceDetails",
+		)
+	}
+	//Assume refererence instance is a vm only instance?
```
krancour: Yes. Good assumption. We should be failing if the reference instance isn't the kind of instance we expected it to be.

pkg/services/sqldb/provision.go
```diff
@@ -95,16 +176,96 @@ func (s *serviceManager) preProvision(
...
+	if err != nil {
+		return nil, err
+	}
+
+	dt.DatabaseName = generate.NewIdentifier()
+
+	//Build the instance details with the reference instance details
+	dt.ServerName = rdt.ServerName
+	dt.AdministratorLogin = rdt.AdministratorLogin
+	dt.AdministratorLoginPassword = rdt.AdministratorLoginPassword
```
krancour: I'm thinking a db-only instance doesn't need to keep track of the three properties above itself, right?

jeremyrickard: Probably not server name, but we don’t pass refer node instance to bind/unbind and we need to have the admin login/password for creating and removing the binding no?

pkg/services/sqldb/types.go
```diff
@@ -2,13 +2,20 @@ package sqldb
...
+type ServerProvisioningParams struct {
 	FirewallIPStart string `json:"firewallStartIPAddress"`
 	FirewallIPEnd   string `json:"firewallEndIPAddress"`
 }
 
+// DBProvisioningParams encapsulates MSSQL-specific provisioning options
+type DBProvisioningParams struct {
+}
+
+// instance details are common to the DB only and All In One Service.
```
krancour: I think there are a few attributes here that are, perhaps, not relevant to db-only. wdyt?

tests/lifecycle/test_cases_test.go
```diff
@@ -18,17 +18,17 @@ func getTestCases(resourceGroup string) ([]serviceLifecycleTestCase, error) {
...
-		getMysqlCases,
-		getPostgresqlCases,
-		getSearchCases,
-		getServicebusCases,
-		getStorageCases,
+		//getMysqlCases,
+		//getPostgresqlCases,
+		//getSearchCases,
+		//getServicebusCases,
+		//getStorageCases,
```
krancour: Did you intend to leave these commented?

I think I want to continue running the full gamut of tests, because eventually, _before_ we can even consider merging this feature branch back into master, we're going to build images from this branch for interested parties to play with. I want to make sure we still have some strong guarantees that all those other modules still work.

tests/lifecycle/mssql_cases_test.go
```diff
@@ -30,13 +30,24 @@ func getMssqlCases(
...
 			location:    "southcentralus",
-			provisioningParameters: &sqldb.ProvisioningParameters{
+			provisioningParameters: &sqldb.ServerProvisioningParams{
 				FirewallIPStart: "0.0.0.0",
 				FirewallIPEnd:   "255.255.255.255",
 			},
 			bindingParameters: &sqldb.BindingParameters{},
 			testCredentials:   testMsSQLCreds(),
 		},
+		{ //server only scenario
```
krancour: I can't comment on lines too far above what's been modified, so I'm adding it here.

On line 27, can you change the comment to reflect the "all-in-one" scenario?

tests/lifecycle/test_cases_test.go
jeremyrickard: Forgot to revert this before I did a commit -a so I’ll revert that back! I think I want to add the ability to selectively run them via env var or something.

tests/lifecycle/mssql_cases_test.go
```diff
@@ -30,13 +30,24 @@ func getMssqlCases(
...
+			module:      sqldb.New(armDeployer, msSQLManager),
+			description: "new server only",
+			serviceID:   "a7454e0e-be2c-46ac-b55f-8c4278117525",
+			planID:      "24f0f42e-1ab3-474e-a5ca-b943b2c48eee",
+			location:    "southcentralus",
+			provisioningParameters: &sqldb.ServerProvisioningParams{
+				FirewallIPStart: "0.0.0.0",
+				FirewallIPEnd:   "255.255.255.255",
+			},
+		}, //TODO: Add a lifecycle test for database only. Need to provision server first
```
krancour: I think we have an established way of doing setup for these lifecycle test cases, which can include provisioning dependent resource. So we can do that and construct the corresponding reference instance. (i.e. The reference instance will be populated by test code after test code provisions the dependencies.) We'll also have to add a `referenceInstance` field to the `serviceLifecycleTestCase` type.

Don't let this block getting this merged.

pkg/services/sqldb/provision.go
krancour: You're right. But I think right thing to do then is probably to start passing ref instance when we bind.

krancour: You know what... what would be even better is if instances had a ref instance _field_. So we never have to pass ref instances around at all. If you have an instance, you'll already have the _means_ to obtain its ref instance (if it has one).

I'll tackle that.

In the meantime, let's not let this issue block us from merging this.

tests/lifecycle/test_cases_test.go
krancour: > I think I want to add the ability to selectively run them via env var or something.

If you do that, PR it to master because that's change is useful regardless of what goes on in this PoC.

krancour:
Registered a few small bits of feedback, but overall, this is great. There are DRY exceptions we can factor out later. (I'd already said that siloing the service managers for this module was a more immediate priority than avoiding DRY exceptions, insofar as this PoC is informing where we go next.)

Re: any feedback I've left. Act on whatever things you reasonably can, but this PR advances us significantly toward what we're trying to achieve with #124 and I think we should get this merged sooner, rather than later. i.e. Don't let perfection be a blocker for this PR.

PR #183, opened 2017-12-29
fix bug decoding tags

krancour:
There is a bug in how we handle tags. Tags are but one of may possible parameters. Parameters themselves are a `map[string]interface{}`. That means, tags are an `interface{}`. We were trying to make a type assertion that tags was a `map[string]string`, but it is not... it is _also_ a `map[string]interface{}`.

This wasn't caught due to lack of e2e testing. (Something we need to resolve, but is consciously taking a back seat to nailing down core functionality.)

This change uses `mapstructure` (which is ordinarily used to decode `map[string]interface{}` onto a struct) to decode the tags `map[string]interface{}` onto a `map[string]string`.


krancour:
Merging without another CI run. The only change pulled in from master after the initial run was a trivial one.

PR #184, opened 2017-12-29
remove quotes from devel pseudo-version string

krancour:
Quotes were becoming part of the version string.

Before this change:
```
INFO[2017-12-29T18:54:27Z] Open Service Broker for Azure starting        commit=6dbd967-dirty version="\"devel\""
```

After:

```
INFO[2017-12-29T18:54:27Z] Open Service Broker for Azure starting        commit=6dbd967-dirty version=devel
```


krancour:
I see no need for CI to run for this change.

PR #192, opened 2018-01-04
Update lifecycle tests to run mssql db only

jeremyrickard:
1.) Small refactor to test_case_test.go to break provision into a
reusable function
2.) Modified setup function signature to include a *service.Instance
    to use as a parent instance
3.) Added setup function in mssql tests
4.) Added lifecycle test for DB only
5.) Fixed db only privision bug

Implemnts #124

### DIFF
--- a/pkg/services/sqldb/bind.go
+++ b/pkg/services/sqldb/bind.go
@@ -1,7 +1,6 @@
 package sqldb
 
 import (
-	"errors"
 	"fmt"
 
 	"github.com/Azure/open-service-broker-azure/pkg/generate"
@@ -171,7 +170,7 @@ func (d *dbOnlyManager) Bind(
 	}
 	//Parent should be set by the framework, but return an error if it is not set.
 	if instance.Parent == nil {
-		return nil, errors.New("parent instance not set")
+		return nil, fmt.Errorf("parent instance not set")
 	}
 	pdt, ok := instance.Parent.Details.(*mssqlVMOnlyInstanceDetails)
 	if !ok {
--- a/pkg/services/sqldb/bind.go
+++ b/pkg/services/sqldb/bind.go
@@ -1,6 +1,7 @@
 package sqldb
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/Azure/open-service-broker-azure/pkg/generate"
@@ -167,6 +168,10 @@ func (d *dbOnlyManager) Bind(
 		return nil, fmt.Errorf(
 			"error casting instance.Details as *mssqlDBOnlyInstanceDetails",
 		)
+	}
+	//Parent should be set by the framework, but return an error if it is not set.
+	if instance.Parent == nil {
+		return nil, errors.New("parent instance not set")
 	}
 	pdt, ok := instance.Parent.Details.(*mssqlVMOnlyInstanceDetails)
 	if !ok {
--- a/pkg/services/sqldb/provision.go
+++ b/pkg/services/sqldb/provision.go
@@ -204,6 +204,11 @@ func (d *dbOnlyManager) preProvision(
 			"error casting instance.Details as *mssqlDBOnlyInstanceDetails",
 		)
 	}
+
+	//Parent should be set by the framework, but return an error if it is not set.
+	if instance.Parent == nil {
+		return nil, errors.New("parent instance not set")
+	}
 	//Assume refererence instance is a vm only instance. Fail if not
 	pdt, ok := instance.Parent.Details.(*mssqlVMOnlyInstanceDetails)
 	if !ok {
@@ -221,9 +226,8 @@ func (d *dbOnlyManager) preProvision(
 	if err != nil {
 		return nil, err
 	}
-
+	dt.ARMDeploymentName = uuid.NewV4().String()
 	dt.DatabaseName = generate.NewIdentifier()
-
 	sqlDatabaseDNSSuffix := azureEnvironment.SQLDatabaseDNSSuffix
 	dt.FullyQualifiedDomainName = fmt.Sprintf(
 		"%s.%s",
@@ -363,7 +367,12 @@ func (d *dbOnlyManager) deployARMTemplate(
 			"error casting instance.Details as *mssqlDBOnlyInstanceDetails",
 		)
 	}
-	pdt, ok := instance.Details.(*mssqlVMOnlyInstanceDetails)
+
+	//Parent should be set by the framework, but return an error if it is not set.
+	if instance.Parent == nil {
+		return nil, errors.New("parent instance not set")
+	}
+	pdt, ok := instance.Parent.Details.(*mssqlVMOnlyInstanceDetails)
 	if !ok {
 		return nil, errors.New(
 			"error casting instance.Parent.Details as " +
--- a/tests/lifecycle/driver_test.go
+++ b/tests/lifecycle/driver_test.go
@@ -46,7 +46,7 @@ func TestServices(t *testing.T) {
 			t.Run(tc.getName(), func(t *testing.T) {
 				// Run subtests in parallel!
 				t.Parallel()
-				err := tc.execute(resourceGroup)
+				err := tc.execute(t, resourceGroup)
 				assert.Nil(t, err)
 			})
 		}
--- a/tests/lifecycle/mssql_cases_test.go
+++ b/tests/lifecycle/mssql_cases_test.go
@@ -3,11 +3,9 @@
 package lifecycle
 
 import (
-	"context"
 	"database/sql"
 	"fmt"
 	"net/url"
-	"time"
 
 	"github.com/Azure/open-service-broker-azure/pkg/azure/arm"
 	ss "github.com/Azure/open-service-broker-azure/pkg/azure/mssql"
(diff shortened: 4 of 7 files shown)


tests/lifecycle/test_case_test.go
```diff
@@ -106,42 +89,12 @@ func (s serviceLifecycleTestCase) execute(resourceGroup string) error {
 		ResourceGroup:          resourceGroup,
 		Details:                serviceManager.GetEmptyInstanceDetails(),
 		ProvisioningParameters: s.provisioningParameters,
+		Parent:                 parent,
 	}
 
-	// Provision...
-	provisioner, err := serviceManager.GetProvisioner(plan)
-	if err != nil {
+	if _, err = s.provision(ctx, serviceManager, instance, plan); err != nil {
```
jeremyrickard: I refactored this out into a method on serviceLifecyleTest so it could be reused within the setup function on the test itself. Since we don't export the actual Instance Details structs, I'm using the provision method to get an instance that's valid (although the instance details are all we really care about).

tests/lifecycle/test_case_test.go
```diff
@@ -58,30 +58,11 @@ func (s serviceLifecycleTestCase) execute(resourceGroup string) error {
 	go s.showStatus(ctx)
 
 	// Get the service and plan
-	cat, err := s.module.GetCatalog()
+	svc, plan, err := s.getServiceAndPlan()
```
jeremyrickard: refactored out so it could be reused by the setup function in the mssql tests.

tests/lifecycle/test_case_test.go
```diff
@@ -106,42 +89,12 @@ func (s serviceLifecycleTestCase) execute(resourceGroup string) error {
...
-		}
-		stepName, ok = provisioner.GetNextStepName(stepName)
-		// If there is no next step, we're done with provisioning
-		if !ok {
-			break
-		}
-	}
 
 	//Only test the binding operations if the service is bindable
 	if svc.GetBindable() {
```
jeremyrickard: I debated refactoring the bind and deprovision functionality out, but right now it doesn't need to be reused so I left it in place. I think stylistically, execute() might read better if all of it was refactored so execute was really just executing subsequent methods that make up the tests and the blocks related to a given lifecycle are contained.

pkg/services/sqldb/bind.go
```diff
@@ -168,6 +168,10 @@ func (d *dbOnlyManager) Bind(
 			"error casting instance.Details as *mssqlDBOnlyInstanceDetails",
 		)
 	}
+	//Parent should be set by the framework, but return an error if it is not set.
```
jeremyrickard: I added these checks because when I first added the life cycle test (but before I was implementing setup), there was no parent instance and there was no check here (framework SHOULD set it before this gets invoked) so the result was a panic. This seemed safer from a runtime perspective, but I can remove this.

pkg/services/sqldb/provision.go
```diff
@@ -221,9 +226,8 @@ func (d *dbOnlyManager) preProvision(
 	if err != nil {
 		return nil, err
 	}
-
+	dt.ARMDeploymentName = uuid.NewV4().String()
```
jeremyrickard: This was a a bug. arm deployment failed because it wasn't set. Lifecycle test found it.

tests/lifecycle/mssql_cases_test.go
```diff
@@ -23,6 +25,47 @@ func getMssqlCases(
 		return nil, err
 	}
 
+	provisionServerOnly := func() (*service.Instance, error) {
+
+		ctx, cancel := context.WithTimeout(context.Background(), time.Minute*20)
+		defer cancel()
+
+		s := serviceLifecycleTestCase{
```
jeremyrickard: this is repeated code from below when the tests are created. I debated refactoring the test declarations out so this one could be reused, but this seemed more readable in the end.

pkg/services/sqldb/bind.go
krancour: A justifiable degree of defensive coding is rarely a bad thing.

tests/lifecycle/mssql_cases_test.go
krancour: I've heard it suggested more than once before that in tests, readability is more important than avoiding DRY exceptions. 👍

krancour:
So this is different than what I had in mind-- it's better. Originally, I had been thinking that in the setup for a db-only test case (for instance), we didn't need to take vm-only through the (OSBA) provisioning process first. I was thinking we could provision the dependency in the simples possible way (i.e. make the minimal sqldb management API calls needed to provision) and then "manufacture" a counterfeit parent instance that matches.

But this is better because it exercises more OSBA code and relies less on more bespoke dependency provisioning logic.

The flip side... this new and frankly better way of looking at the problem is making me realize there's another opportunity improve these tests that I think I'd regret leaving on the table-- or we'd regret it later when we start to give Postgres and MySQL and other modules makeovers similar to the one that sqldb is getting right now...

As it stands, I think this PR will go through vm-only provisioning _twice_. Once as part of the vm-only test case and the second time as a pre-req for the db-only test case. (Again, remember that other modules will eventually get a similar makeover.) Are we dramatically increasing the time required to test? Are we increasing the resource requirements to test? The cost of the tests?

When if we can codify the relationship between related test cases?

Suppose that we modified `serviceLifecycleTestCase` to have a new field-- `childTestCases []*serviceLifecycleTestCase`. The db-only case could be "nested" in the vm-only case. In this PR, you've extracted the provisioning logic from the test case `execute(...)` function to reuse (and that's fine) but instead of (or in addition to), what we could do is recursively call `execute(...)` on each child at the right point in a parent's lifecycle test.

This would optimize the testing of lifecycles for parents and their children and would probably lead to less test code overall. We could even dump all of the bespoke "setup" hooks because we wouldn't need them anymore.

If you feel this is a huge departure from what's in this PR, we can certainly tackle it in a follow-up, but my sense (and you truly know better than me) is that _maybe_ it wouldn't take much to amend this PR as described above.

Thoughts?

tests/lifecycle/driver_test.go
```diff
@@ -46,7 +46,7 @@ func TestServices(t *testing.T) {
 			t.Run(tc.getName(), func(t *testing.T) {
 				// Run subtests in parallel!
 				t.Parallel()
-				err := tc.execute(resourceGroup)
+				err := tc.execute(t, resourceGroup)
```
krancour: This can be a little bit of a problem, iirc. If you pass `t` around to "test helpers" and make assertions in those helpers, I do not think that the output from the test run will give correct attribution to the real source of the error. (There might be some ways around that that have emerged in the last few months. I'll research that a bit.) It might be safer here to do as we were before-- treat `tc.execute()` and everything _it_ calls as "normal" functions that return errors instead of directly failing the tests using `t`.

tests/lifecycle/mssql_cases_test.go
```diff
@@ -39,15 +39,26 @@ func getMssqlCases(
...
+			description: "new server",
 			serviceID:   "a7454e0e-be2c-46ac-b55f-8c4278117525",
 			planID:      "24f0f42e-1ab3-474e-a5ca-b943b2c48eee",
 			location:    "southcentralus",
 			provisioningParameters: &sqldb.ServerProvisioningParams{
 				FirewallIPStart: "0.0.0.0",
 				FirewallIPEnd:   "255.255.255.255",
 			},
-		}, //TODO: Add a lifecycle test for database only.
+			childTestCases: []*serviceLifecycleTestCase{
```
krancour: This is beautiful.

tests/lifecycle/test_case_test.go
```diff
@@ -19,11 +21,12 @@ import (
 type serviceLifecycleTestCase struct {
 	module                 service.Module
 	description            string
-	setup                  func() error
 	serviceID              string
 	planID                 string
 	location               string
 	provisioningParameters service.ProvisioningParameters
+	parentService          *service.Instance
```
krancour: Can we change the name of this attribute? At a glance, it looks like a service and not a service instance. If I'm more careful about noticing the type, it becomes clear, but I don't want the name of the attribute itself to mislead anyone.

tests/lifecycle/test_case_test.go
```diff
@@ -19,11 +21,12 @@ import (
 type serviceLifecycleTestCase struct {
 	module                 service.Module
 	description            string
-	setup                  func() error
```
krancour: Good riddance.

tests/lifecycle/test_case_test.go
```diff
@@ -172,6 +175,23 @@ func (s serviceLifecycleTestCase) execute(resourceGroup string) error {
...
+			subTestName = fmt.Sprintf("subtest-%v", i)
+		} else {
+			subTestName = strings.Replace(test.description, " ", "_", -1)
+		}
+		t.Run(subTestName, func(t *testing.T) {
+			tErr := test.execute(t, resourceGroup)
+			assert.Nil(t, tErr)
+		})
+	}
+
```
krancour: Ooooh. You weren't passing `t` around as a way of failing tests. You were using it to launch additional subtests.

krancour: I think this is ok as long as we're sure failures are always getting chalked up to the correct case.

krancour:
This LGTM. I know you have one thing you might change re: a certain assertion, but let's assume my LGTM stands. Feel free to merge after you are satisfied with this and CI has passed.

jeremyrickard:
@krancour running the child test cases like this makes them subtests, so the hierarchy of the output ends up being like this:

```
--- FAIL: TestServices (255.77s)
    --- FAIL: TestServices/lifecycle (0.00s)
        --- PASS: TestServices/lifecycle/TestServices/lifecycle/mssql/new_server_and_database_(all_in_one) (184.29s)
        --- FAIL: TestServices/lifecycle/TestServices/lifecycle/mssql/new_server (206.06s)
            --- FAIL: TestServices/lifecycle/TestServices/lifecycle/mssql/new_server/database_on_new_server (0.00s)
        Error Trace:    test_case_test.go:193
        Error:      	Expected nil, but got: &errors.errorString{s:"parent instance not set"}
            --- PASS: TestServices/lifecycle/TestServices/lifecycle/mssql/new_server/subtest-1 (105.25s)
```

I think it's clear *where* the failure is, but it does also fail the parent since they are tied together in the parent/child relationship, so it kind of obscures that the one test *did* pass. In this case, there are two sub tests, one passed and one didn't. Clear *there* which one failed and passed, but not on the "server only" lifecycle test. 

So may need to think of a different approach if we want to fully expose where the failure is. When a test fails, start at the lowest level and fix issues and work up?

krancour:
@jeremyrickard that looks clear _enough_ to me in terms of where the test failures are. I'm going to merge this.

PR #198, opened 2018-01-08
append instead of replacing user-agent

joshgav:
updates https://github.com/Azure/open-service-broker-azure/pull/172

### DIFF
--- a/pkg/azure/arm/deployer.go
+++ b/pkg/azure/arm/deployer.go
@@ -109,7 +109,7 @@ func (d *deployer) Deploy(
 		d.subscriptionID,
 	)
 	deploymentsClient.Authorizer = authorizer
-	deploymentsClient.UserAgent = userAgent()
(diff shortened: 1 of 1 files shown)


krancour:
Thanks @joshgav. This LGTM, but underscores some existing DRY exceptions in this code. We'll merge this PR sooner than later (pending CI) and I'll open a separate PR to fix the DRY exceptions.

Thanks!

krancour:
There's a lint failure here. I will fix.

krancour:
Lint failures are fixed. Holding on LGTM / merge because I think we're waiting on a definite answer re: whether our analytics can query "contains" on the user-agent string instead of querying for an exact match.

seanmck:
Confirmed that telemetry can do regex, so we're good to go.

joshgav:
thanks!

jeremyrickard:
Thanks for confirming @seanmck.

PR #201, opened 2018-01-09
parent child count

krancour:
Needs #200 merged first. This allows us to get a count of existing children for any given instance, as identified by its alias. Note, we have no need to know who the children are-- just how many exist. More precisely, we will eventually care that a parent can only be deprovisioned if 0 children remain.


krancour:
@jeremyrickard, I had to rebase on `sqldb-catalog-poc` after #200 was merged and it cleared your LGTM. Can you re-approve, please?

PR #204, opened 2018-01-11
Provision and Deprovision enhancements to use new async features

jeremyrickard:
Using the updated features of the async engine, pkg/api/provision and deprovision were enhanced to determine if a provision or deprovision action needs to wait for
another operation to finish. If so, they submit delayed tasks, either
wait_for_parent or wait_for_children, whichever is appropriate. Those
new jobs will make the same determination, and either readd themselves
as delayed tasks, or initiate the task that would otherwise originally
have been created. 

Provision was also enhanced to reject provision requests for things have parents that are currently deprovisioning or have failed.


pkg/api/provision.go
```diff
@@ -411,13 +439,65 @@ func (s *server) provision(w http.ResponseWriter, r *http.Request) {
...
 	log.WithFields(logFields).Debug("asynchronous provisioning initiated")
 }
 
+func (s *server) isParentProvisioning(instance service.Instance) (bool, error) {
+	//No parent, so no need to wait
+	if instance.ParentAlias == "" {
+		return false, nil
+	}
+
+	parent, parentFound, err := s.store.GetInstanceByAlias(instance.ParentAlias)
```
krancour: Is this because `instance.Parent` will always be nil at this point in time?

pkg/api/provision.go
```diff
@@ -411,13 +439,65 @@ func (s *server) provision(w http.ResponseWriter, r *http.Request) {
...
+
+	//Parent has was not found, so wait for that that to occur
+	if !parentFound {
+		return true, nil
+	}
+
+	if err != nil {
```
krancour: Knowing how `GetInstanceByAlias` is implemented, this is _probably_ ok, but it might be better to code more defensively here by checking for an error _before_ deciding to trust the value of `parentFound`.

pkg/api/provision.go
jeremyrickard: Yeah, at this point in time the instance.Parent is nil (we just created the instance) as we have not retrieved it from storage.  Just realized that should not need to do that in the tasks.

pkg/api/provision.go
jeremyrickard: Will do.

pkg/api/provision.go
```diff
@@ -411,13 +439,65 @@ func (s *server) provision(w http.ResponseWriter, r *http.Request) {
...
+			"bad provision request: unable to retrieve parent",
+		)
+		return false, err
+	}
+
+	//If parent failed, we should not even attempt to provision this
+	if parent.Status == service.InstanceStateProvisioningFailed {
+		log.WithFields(log.Fields{
+			"error":      "waitforParent",
+			"instanceID": instance.InstanceID,
```
krancour: Let's include the parent ID in these log fields also.

jeremyrickard: 👍  where available, I'll add ID. IF ID is not known, I'll us Alias.

pkg/api/provision.go
```diff
@@ -411,13 +439,65 @@ func (s *server) provision(w http.ResponseWriter, r *http.Request) {
...
+		}).Error(
+			"bad provision request: parent failed provisioning",
+		)
+		return false, fmt.Errorf("error provisioning: parent provision failed")
```
krancour: imo, this should be severity `INFO`; not `ERROR`. My reasoning (which I have applied as consistently as I've been able throughout the rest of the broker) is that we use `ERROR` when the broker itself has experienced a problem. In this case, it has not. The broker refusing to provision a child for a parent that has already failed provisioning is actually the broker working as designed. `INFO` because is _is_ worth knowing about, but not `ERROR` because this isn't evidence that anything is wrong with the broker.

pkg/api/provision.go
```diff
@@ -411,13 +439,65 @@ func (s *server) provision(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	//If parent is deprovisioning, we should not even attempt to provision this
+	if parent.Status == service.InstanceStateDeprovisioning {
+		log.WithFields(log.Fields{
+			"error":      "waitforParent",
+			"instanceID": instance.InstanceID,
+		}).Error(
```
krancour: Same comment as [here](https://github.com/Azure/open-service-broker-azure/pull/204/files#r161841171)

pkg/api/provision.go
jeremyrickard: I removed this from the task (but have not pushed).

pkg/broker/deprovision.go
```diff
@@ -89,7 +89,7 @@ func (b *broker) doDeprovisionStep(
 			"error loading persisted instance",
 		)
 	}
-
+
```
krancour: Nit: Not sure why space got added here.

pkg/broker/provision.go
```diff
@@ -89,7 +89,7 @@ func (b *broker) doProvisionStep(
 			"error loading persisted instance",
 		)
 	}
-
+
```
krancour: Nit: Not sure why space got added here.

pkg/api/provision.go
krancour: > IF ID is not known, I'll use Alias. 👍

pkg/broker/wait_for_children.go
```diff
@@ -0,0 +1,103 @@
...
+	log "github.com/Sirupsen/logrus"
+)
+
+func (b *broker) doWaitForChildrenStep(
+	ctx context.Context,
+	args map[string]string,
+) error {
+
+	_, cancel := context.WithCancel(ctx)
+	defer cancel()
```
krancour: This looks like a mistake. This will cancel the context when this function returns, but this context isn't being passed to any functions. (It can't be, because we've used `_` to refuse assigning it to any variable. If `ctx`, is passed to any functions below [I haven't looked yet], that is the `ctx` _we were passed_ and _that_ `ctx` won't be canceled when this function returns.

pkg/broker/wait_for_parent.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/Azure/open-service-broker-azure/pkg/service"
+	log "github.com/Sirupsen/logrus"
+)
+
+func (b *broker) doWaitForParentStep(
+	ctx context.Context,
+	args map[string]string,
+) error {
+	_, cancel := context.WithCancel(ctx)
+	defer cancel()
```
krancour: Same comment as [here](https://github.com/Azure/open-service-broker-azure/pull/204/files#r161846445)

pkg/api/deprovision.go
```diff
@@ -142,21 +143,46 @@ func (s *server) deprovision(w http.ResponseWriter, r *http.Request) {
...
+		)
+		s.writeResponse(w, http.StatusInternalServerError, responseEmptyJSON)
+	}
+
+	var task model.Task
+	if childCount > 0 {
+		task = model.NewDelayedTask(
+			"waitForChildrenStep",
+			map[string]string{
+				"deprovisionFirstStep": firstStepName,
```
krancour: For consistency with the more immediate provision/deprovision tasks, can this just be "stepName"?

pkg/api/provision.go
```diff
@@ -395,13 +406,30 @@ func (s *server) provision(w http.ResponseWriter, r *http.Request) {
...
-			"stepName":   firstStepName,
-			"instanceID": instanceID,
-		},
-	)
+	var task model.Task
+	if waitForParent {
+		task = model.NewDelayedTask(
+			"waitForParentStep",
+			map[string]string{
+				"provisionFirstStep": firstStepName,
```
krancour: Same as [this comment](https://github.com/Azure/open-service-broker-azure/pull/204/files#r161847903)

pkg/api/deprovision.go
```diff
@@ -142,21 +143,46 @@ func (s *server) deprovision(w http.ResponseWriter, r *http.Request) {
...
+		log.WithFields(logFields).Error(
+			"deprovisioning error: error determining child count",
+		)
+		s.writeResponse(w, http.StatusInternalServerError, responseEmptyJSON)
+	}
+
+	var task model.Task
+	if childCount > 0 {
+		task = model.NewDelayedTask(
+			"waitForChildrenStep",
```
krancour: Semantics: existing task names are things like `deprovisionStep` (emphasis on _step_) because that's what those tasks do-- they execute one _step_ of the deprovisioning process. (Step is a "technical" term here, in a manner of speaking, because the provisioners and deprovisioners are literally composed of "steps.") I'm not sure how applicable "step" is to what we're doing here. `waitForChildren` (sans "step"), might be a more appropriate task name.

Part 2-- more semantics: `waitForChildren` still might not adequately describe the task. The task doesn't do much waiting. It's awake for a moment before either scheduling another task like itself or scheduling the first deprovisioning step. Maybe `checkChildrensStatuses` works?

Sorry to split hairs so much, but the naming of tasks (for instance) does fall into the category of things that we eventually can not change anymore, so I want to get it right.

pkg/api/provision.go
```diff
@@ -395,13 +406,30 @@ func (s *server) provision(w http.ResponseWriter, r *http.Request) {
...
-		"provisionStep",
-		map[string]string{
-			"stepName":   firstStepName,
-			"instanceID": instanceID,
-		},
-	)
+	var task model.Task
+	if waitForParent {
+		task = model.NewDelayedTask(
+			"waitForParentStep",
```
krancour: Same as [this comment](https://github.com/Azure/open-service-broker-azure/pull/204/files#r161850904)

pkg/api/deprovision.go
krancour: Actually-- strike that. We don't need to pass this information through the queue.

When the step that checks children's statuses wakes up, it retrieves the instance by ID. From that instance, we know the service and we can get the service's provisioner, which can tell us what its first step is.

pkg/api/provision.go
krancour: Same as [this comment](https://github.com/Azure/open-service-broker-azure/pull/204/files#r161852255)

pkg/api/deprovision.go
jeremyrickard: Yeah, this is totally doable. I passed this through just to avoid duplicating the lookup logic from the things in pkg/api (get service, get plan, get service manger, get deprovisioner/provisioner), but this is easily changed.

pkg/api/deprovision.go
jeremyrickard: 👍

pkg/api/deprovision.go
krancour: > I passed this through just to avoid duplicating the lookup logic from the things in pkg/api (get service, get plan, get service manger, get deprovisioner/provisioner), but this is easily changed.

I have a todo to look up service and plan when instances are retrieved from the database-- then we can _stop_ doing that in a bunch of other places.

jeremyrickard: 👍

pkg/broker/check_children_status.go
```diff
@@ -0,0 +1,167 @@
...
+		task = model.NewDelayedTask(
+			"checkChildrenStatus",
+			map[string]string{
+				"instanceID": instanceID,
+			},
+			time.Minute*5,
+		)
+		log.WithFields(log.Fields{
+			"instanceID": instanceID,
+		}).Debug("children not deprovisioned, will wait again")
```
krancour: Nice message. Great for debugging.

pkg/broker/check_children_status.go
```diff
@@ -0,0 +1,167 @@
...
+				instance,
+				"checkChildrenStatus",
+				nil,
+				"error: no Service found for serviceID",
+			)
+		}
+		plan, ok := svc.GetPlan(instance.PlanID)
+		if !ok {
+			// If we don't find the Service in the catalog, something is really wrong.
+			// (It should exist, because an instance with this serviceID exists.)
```
krancour: This comment is wrong-- and I realize it was probably wrong where it was copied from too. Can we fix it in just this spot for now?

pkg/broker/check_children_status.go
```diff
@@ -0,0 +1,167 @@
...
+			)
+			return b.handleDeprovisioningError(
+				instance,
+				"checkChildrenStatus",
+				nil,
+				"error: no Plan found for planID",
+			)
+		}
+		serviceManager := svc.GetServiceManager()
+		deprovisioner, dErr := serviceManager.GetDeprovisioner(plan)
```
krancour: Why `dErr` instead of the conventional `err`?

jeremyrickard: Lint complained that it was shadowing the _err_ declared outside of the conditional block (same in the other file).

pkg/broker/check_parent_status.go
```diff
@@ -0,0 +1,193 @@
...
+		)
+	}
+	var task model.Task
+	if waitForParent {
+		task = model.NewDelayedTask(
+			"checkParentStatus",
+			map[string]string{
+				"instanceID": instanceID,
+			},
+			time.Minute*5,
```
krancour: Applicable in a few other spots as well... can we bring this down to one minute (for now)? Speaking from a k8s perspective, there are already so many progressive back-offs in play. I worry how far k8s (which is waiting on us) backs off while we are potentially prolonging a process like this longer than we need to. In the future, we can either make this tunable or implement a progressive backoff of our own (that will require us to track how many times we've looped through this check task and we've already discussed doing that in a follow up).

jeremyrickard:
CI passed, merging per Slack conversation.

PR #206, opened 2018-01-12
more assertions on worker shutdown behavior

krancour:
This tests that if any OTHER component of the worker stops working due to error, returning without error, or having its context canceled, the delayed queue watcher gets its own context canceled.


jeremyrickard:
Skipping CI for now

PR #207, opened 2018-01-12
refactor watcher

krancour:
WIP because this needs tests still, but it's not too early to start reviewing the implementation.

The idea here is to stop the goroutine that watches the main delayed work queue from being _so_ chatty with Redis. As it stand before this PR, a single delayed task could be moved back and forth between the main delayed work queue and worker's delayed work queue multiple times per second for as long as it takes to reach the scheduled execution time. That seems as if it would needlessly consume CPU and network resources.

This change works as follows:

1. Work _already_ (i.e. before this PR) gets plucked from the main delayed work queue onto a worker's delayed work queue. (This is important.)
2. This PR only changes the work performed. Previously the work was to compare the current time to the scheduled execution time and then place the task onto either the main active work queue or the main delayed work queue. _Now_ the work performed involves starting an additional goroutine to watch the clock until the time comes to place the task on the main active work queue.
3. If the broker process dies for any reason, the tasks corresponding to any interrupted timers are sitting on the worker's delayed work queue. The cleaning process _already_ will move such work make to the main delayed work queue. Eventually, it will be picked up by another worker and that worker will start a new timer.

### DIFF
--- a/pkg/async/worker_test.go
+++ b/pkg/async/worker_test.go
@@ -32,7 +32,7 @@ func TestWorkerWorkBlocksUntilHeartErrors(t *testing.T) {
 		return ctx.Err()
 	}
 	watchDelayedTasksStopped := false
-	w.watchDelayedTasks = func(
+	w.handleDelayedTasks = func(
 		ctx context.Context,
 		delayedQueueName string,
 		activeQueueName string,
@@ -68,7 +68,7 @@ func TestWorkerWorkBlocksUntilHeartReturns(t *testing.T) {
 		return ctx.Err()
 	}
 	watchDelayedTasksStopped := false
-	w.watchDelayedTasks = func(
+	w.handleDelayedTasks = func(
 		ctx context.Context,
 		delayedQueueName string,
 		activeQueueName string,
@@ -104,7 +104,7 @@ func TestWorkerWorkBlocksUntilReceiveAndWorkErrors(t *testing.T) {
 		return errSome
 	}
 	watchDelayedTasksStopped := false
-	w.watchDelayedTasks = func(
+	w.handleDelayedTasks = func(
 		ctx context.Context,
 		delayedQueueName string,
 		activeQueueName string,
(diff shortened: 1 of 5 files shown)


jeremyrickard:
Looks like what we discussed offline and should handle the issue.  👍

pkg/async/worker_test.go
```diff
@@ -283,3 +283,78 @@ func TestWorkerReceiveAndWorkBlocksUntilContextCanceled(t *testing.T) {
 	err := w.receiveAndWork(ctx, getDisposableQueueName())
 	assert.Equal(t, ctx.Err(), err)
 }
+
+func TestWorkerDoesNotMoveTaskEarly(t *testing.T) {
```
krancour: I know the names of some of our test functions are unpleasantly long, but I want to make sure they're sufficiently descriptive. A worker is comprised of multiple components, so can this be something like `TestDelayedQueueWatcherDoesNotMoveTaskEarly`? Similar though for the other new test.

pkg/async/worker_test.go
```diff
@@ -283,3 +283,78 @@ func TestWorkerReceiveAndWorkBlocksUntilContextCanceled(t *testing.T) {
...
+			"instanceID": "provision",
+		},
+		time.Minute*10,
+	)
+	taskJSON, _ := task.ToJSON()
+	redisClient.LPush(mainDelayedWorkQueueName, taskJSON)
+	w := newWorker(redisClient).(*worker)
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*10)
+	defer cancel()
+	_ = w.handleDelayedTasks(
```
krancour: I think there's a more "focused" way to test this. This function call invokes the _loop_ that listens to the main delayed work queue and spawns new goroutines that launch timers and do the eventual move. If what we want to test is just that tasks aren't moved prematurely, we can test the `handleDelayedTask()` (singular / non-loop) function directly. Same applies to the other new test in this commit.

krancour: We can write separate tests for the loop. They'll just verify that `handleDelayedTask()` is invoked once per delayed task.

krancour: ^ And I will happily contribute that loop test.

PR #208, opened 2018-01-15
add deferred task execution to async package

krancour:
Achieving this (_cleanly_) involved a significant refactor of the entire package. The good news is that the tests are improved beyond what we had before (and liberally annotated now), so i think we can be more confident in the refactored package than the original.

Note to @jeremyrickard. After this is merged into master, we should merge the latest from master into the branch for #124, then rebase #204. #204 should be updated to deal _only_ with deferred child provisioning / deferred parent deprovisioning, and not concern itself with the internals of the async engine.

### DIFF
--- a/pkg/async/cleaner.go
+++ b/pkg/async/cleaner.go
@@ -9,23 +9,38 @@ import (
 	"github.com/go-redis/redis"
 )
 
-type cleanFunction func(workerSetName, mainWorkQueueName string) error
+type cleanFn func(
+	ctx context.Context,
+	workerSetName string,
+	pendingTaskQueueName string,
+	deferredTaskQueueName string,
+) error
 
-type cleanWorkerFunction func(workerID, mainWorkQueueName string) error
+type cleanWorkerQueueFn func(
+	ctx context.Context,
+	workerID string,
+	pendingTaskQueueName string,
+	deferredTaskQueueName string,
+) error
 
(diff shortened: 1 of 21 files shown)


pkg/async/cleaner.go
```diff
@@ -55,62 +79,92 @@ func (c *cleaner) Clean(ctx context.Context) error {
...
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+		}
 	}
 	return nil
 }
 
-func (c *cleaner) defaultCleanWorker(workerID, mainWorkQueueName string) error {
+func (c *cleaner) defaultCleanWorkerQueue(
```
jeremyrickard: nice refactor.

pkg/async/worker.go
```diff
@@ -87,163 +120,337 @@ func (w *worker) Work(ctx context.Context) error {
...
 	}
 }
 
-// defaultReceiveAndWork synchronously receives and completes work. By combining
-// these two operations, a worker never receives more work than it currently
-// has the capacity to process.
-func (w *worker) defaultReceiveAndWork(
+// defaultReceive receives tasks from a source queue and dispatches them to a
+// to both a destination queue and a return channel.
+func (w *worker) defaultReceiveTasks(
```
jeremyrickard: I like how you split this up.

PR #212, opened 2018-01-17
Add async support for "follow-up" tasks

### LINKED ISSUE #203: Possible race in async provisioning / deprovisioning
Provisioning and deprovisioning are implemented as a series of idempotent (or _should be_ idempotent steps). At the end of each step, the service instance that the step has been operating on is updated. There _should_ be no race condition there because n  > 1 provisioning or deprovisioning steps _should_ never be on the queue at the same time-- meaning only _one_ async worker can ever be working on a given instance at a time.

There's one possible, albeit unlikely, way this can all go off the rails...

At the end of each provisioning / deprovisioning step, the _next_ step in the sequence is enqueued. The problem is that this isn't currently done atomically with the removal of the current step from the queue. If enqueuing the next step succeeds and then dequeuing the current step fails (suppose a network failure were to occur at that moment), the queue would contain _two_ steps pertaining to the same instance. While the failure will result in the worker and the whole process dying, the broker will eventually restart and recover and _two_ workers could pick up the two work items pertaining to the same service instance. At that point there's a race condition. One worker will complete first and whatever modifications is makes to the service instance won't be accounted for by the modifications made by the second worker that finishes after. Worse still, the first worker will re-enqueue the next step-- which is already in process (or possibly complete).

The way to fix this is to build some basic support for async request "pipelining" into the async engine. Instead of allowing steps in the provisioning/deprovisoning processes to directly enqueue the next step, they (or any async task) should be able to _return_ the next task. The async engine itself can then _atomically_ dequeue the current task and enqueue the next task(s) _atomically_.

Although this sounds complex, I think this is a relatively easy fix.

krancour:
Fixes #203

### DIFF
--- a/pkg/async/redis/worker.go
+++ b/pkg/async/redis/worker.go
@@ -27,6 +27,7 @@ type executeTasksFn func(
 	ctx context.Context,
 	inputCh chan []byte,
 	pendingTaskQueueName string,
+	deferredTaskQueueName string,
 	errCh chan error,
 )
 
@@ -151,6 +152,7 @@ func (w *worker) Run(ctx context.Context) error {
 				ctx,
 				pendingReceiverRetCh,
 				pendingTaskQueueName,
+				deferredTaskQueueName,
 				executorErrCh,
 			)
 		}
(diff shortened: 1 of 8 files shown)


pkg/async/redis/worker.go
```diff
@@ -327,7 +330,11 @@ func (w *worker) defaultExecuteTasks(
 				}
 				continue
 			}
-			if err := jobFn(ctx, task.GetArgs()); err != nil {
+			taskSuccess := false
+			followUpTaskJSONs := [][]byte{}
+			hadMarshalingError := false
+			followUpTasks, err := jobFn(ctx, task)
```
jeremyrickard: ❤️

PR #218, opened 2018-01-22
Refactor for generic filter capabilities

### LINKED ISSUE #83: Require X-Broker-API-Version Header?
The Open Service Broker API Specification defines a required [API Version Header](https://github.com/openservicebrokerapi/servicebroker/blob/v2.13/spec.md#api-version-header). The current set of [compliance tests](https://github.com/Haishi2016/osb-checker) assert that a request without the header should be rejected with a 412:

 _should reject requests without X-Broker-API-Version header with 412_

The spec states that requests _MUST_ contain the header but doesn't state what the broker should do in the absence of the header. 

Is this an interpretation of the spec that we want to enforce @krancour @arschles or should we submit a change to the compliance tests?

jeremyrickard:
Fixes #83 

1.) Create new api/filter structure
2.) Refactored authenticator stuff there
3.) Created new interfaces: filters.Filter and filters.Chain
4.) Made basic/always authenticator implement filters.Filter interface
5.) Changed pkg/api/Server.NewServer to accept a filters.Chain instead of just authenticator
6.) Implemented a new filter to validate headers
7.) Modified broker to build a filters.Chain with authenticator  and header validator


pkg/api/filters/headers/validator.go
```diff
@@ -0,0 +1,56 @@
...
+
+//Filter validates that required headers are present
+func (v *validator) Filter(handler http.HandlerFunc) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		apiVersion := r.Header.Get("X-Broker-API-Version")
+		if apiVersion == "" {
+			v.sendError(w, responseMissingAPIVersion)
+			return
+		}
+		if !strings.HasPrefix(apiVersion, "2.") {
```
jeremyrickard: This is the same thing that Pivotal does in their go broker api library. We can refine this to a specific version if we'd like.

pkg/api/server.go
```diff
@@ -52,7 +52,7 @@ func NewServer(
 	port int,
 	store storage.Store,
 	asyncEngine async.Engine,
-	authenticator authenticator.Authenticator,
+	filterChain filters.Chain,
```
jeremyrickard: The server was essentially applying the authenticator as a filter previously. Now we allow a chain of them to be passed to the server. This could be just the authenticator. It could be multiple authenticators later on (like if we add support for opaque bearer tokens). This PR also adds in a header validator.

pkg/api/filters/authenticator/authenticator.go
```diff
@@ -0,0 +1,8 @@
+package authenticator
+
+import (
+	"github.com/Azure/open-service-broker-azure/pkg/api/filters"
+)
+
+// Authenticator is an alias of filters.Filter that represents an authenticator
+type Authenticator filters.Filter
```
jeremyrickard: I haven't refactored the broker interface yet. It still takes in the Authenticator type instead of a filter chain. I wasn't sure if I wanted to contain that in the broker or expose that so the builder of the broker (i.e. cmd/broker/broker.go) should determine the configuration. I'm leaning that way, since we construct other things there. (i.e. the authenticator, redis, etc). Maybe move the filter chain construction here.

pkg/api/filters/filter.go
```diff
@@ -0,0 +1,40 @@
...
+//the Filter(s) provided
+func NewFilterChain(filterChain []Filter) Chain {
+	return chain{
+		filters: filterChain,
+	}
+}
+
+//Filter applies a chain of FilterFunction to a target http.HandlerFun
+//and returns a new http.HandlerFunc that wraps the chain
+func (h chain) Filter(
```
jeremyrickard: The heart of this change is contained in this file. Up above, two new interfaces are defined. A Filter, which can execute on a http.HandlerFunc and return another one. The Chain interface defines a Filter method. The Chain is implemented by the chain struct, which has a slice of Filter(s). When it's Filter method is called with a http.HandlerFunc (i.e. 	filterChain.Filter(s.provision)), it wraps that argument with the filter slice in reverse order.

jeremyrickard:
@krancour we discussed having a more generic filtering capability as we add new cross-cutting concerns. I've taken a stab at doing this here. Take a look and let me know what you think.

pkg/api/filters/authenticator/authenticator.go
krancour: Yes. There are most assuredly spots where we're _not_ doing this, but I think we should generally favor treating the `cmd/broker` code as a factory of sorts. It will instantiate all the right implementations of things (based on configuration) and use "constructor injection" to furnish lower-level code with its dependencies.

pkg/api/filters/filter.go
```diff
@@ -0,0 +1,40 @@
+package filters
```
krancour: Any thoughts about renaming this package to `filter`? There's no hard and fast rule we've observed on package naming so far, but I'm just thinking that `filter.Chain` reads better than `filters.Chain`.

pkg/api/filters/filter.go
```diff
@@ -0,0 +1,40 @@
...
+
+import "net/http"
+
+//Filter is a wrapper around a http.HandlerFunc.
```
krancour: _Total_ style nit. Can we put a space after `//`. Applicable in many other spots, too. If you say no, we can still be friends, so don't feel like you have to.

pkg/api/filters/filter.go
jeremyrickard: Yeah, agree. I forgot to rename it before i submitted the review :)

pkg/api/filters/filter.go
krancour: Q: Do we really need a separate interface for `FilterChain`? Cannot `FilterChain` be a specific _implementation_ of the `Filter` interface that just has the logic to chain a composition of _other_ `Filter` implementations together?

jeremyrickard: I don't think we do, given they are the same. I think that these can be consolidated and FilterChain is just another filter, like you mentioned.

pkg/api/filters/headers/validator.go
```diff
@@ -0,0 +1,56 @@
...
+	`{ "error": "MissingAPIVersion", "description": "The request did not ` +
+		`include the X-Broker-API-Version header"}`,
+)
+
+var responseAPIVersionIncorrect = []byte(
+	`{ "error": "APIVersionIncorrect", "description": "X-Broker-API-Verson ` +
+		`header includes an incompatible version"}`,
+)
+
+type validator struct{}
```
krancour: Should we be more specific about _which_ header this validates? There are conceivably other headers we might wish to validate in the future. Or maybe we could make this one generic, and pluggable with a "constructor" that takes a header name and a `func(value string) bool`? Thoughts?

pkg/api/filters/headers/validator.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+func (v *validator) sendError(
+	w http.ResponseWriter,
+	responseBody []byte,
+) {
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusPreconditionFailed)
+	if _, err := w.Write(responseBody); err != nil {
+		log.WithField("error", err).Error(
```
krancour: Bad request != error from the broker's perspective. (Litmus test is whether an operator would want to know about it.) Can we downgrade this to info or debug?

pkg/api/filters/headers/validator.go
jeremyrickard: Yeah, I think the more generic form with a pluggable constructor is better. It's probable that we will need to fix this later.

pkg/api/filters/headers/validator.go
jeremyrickard: 👍

pkg/api/filters/filter.go
krancour: There might be a very good reason for it, btw. You tell me.

jeremyrickard: I don't think so. It's mainly an artifact of me working through it in pieces. I think conceptually it's the same thing. It's a "filter" that applies other "filters" in the end,

krancour:
This is a really positive direction. I have a question about possibly combining the Filter and FilterChain interfaces. Just about all other feedback is very, very minor. I'm excited for this improvement.

pkg/api/filters/headers/validator.go
jeremyrickard: Actually, in this case it's a failure writing the response, not the filter failure itself. I think this one should still be Error no? In this case, the _broker_ failed to do something. That log only happens when err != nil on the write.

krancour: You're right. I misread.

pkg/api/filter/filter.go
```diff
@@ -0,0 +1,33 @@
...
+	Filter(http.HandlerFunc) http.HandlerFunc
+}
+
+type chain struct {
+	filters []Filter
+}
+
+// NewFilterChain returns a Filter capable of applying
+// the provided Filter(s) to another http.HandlerFunc
+func NewFilterChain(filterChain []Filter) Filter {
```
krancour: Can we possibly change the signature to this:

```
func NewFilterChain(filters ...Filter) Filter
```

That will let callers off the hook for explicitly constructing a slice that they likely do not need otherwise.

pkg/api/filter/filter.go
```diff
@@ -0,0 +1,33 @@
+package filter
+
+import "net/http"
+
+// Filter is a wrapper around a http.HandlerFunc.
+type Filter interface {
+	Filter(http.HandlerFunc) http.HandlerFunc
+}
```
krancour: Can we separate the filter interface and the chain implementation of that interface into separate files? imo it would just help keep things clearer in the mind of the reader.

pkg/api/filter/filter.go
krancour: Also, maybe s/NewFilterChain/NewChain/? "filter" is already specified by the package name. `filter.NewChain(...)` reads nice and clean.

pkg/api/filter/filter.go
```diff
@@ -0,0 +1,33 @@
...
+// the provided Filter(s) to another http.HandlerFunc
+func NewFilterChain(filterChain []Filter) Filter {
+	return chain{
+		filters: filterChain,
+	}
+}
+
+// Filter applies a chain of FilterFunction to a target http.HandlerFun
+// and returns a new http.HandlerFunc that wraps the chain
+func (h chain) Filter(
```
krancour: `Filter` sound like a verb or an imperative to actually _do_ the filtering, but this function really just returns a handler. The handler is a decorated one that _does_ the actual filtering. Maybe this function is better renamed as something like `GetHandler(...)`?

jeremyrickard: No problem.

pkg/api/filter/filter.go
jeremyrickard: Yeah, that's a good idea. Along with this, I'm going to reintroduce the Chain interface I think just to make it extra clear the differences.

pkg/api/filter/filter.go
jeremyrickard: Yep!

PR #220, opened 2018-01-22
add svc and plan to instance during retrieval

### LINKED ISSUE #159: Set service and plan on instances when retrieving
When instances are retrieved from storage, we already look up the service so that we can obtain and unmarshal into  service-specific context and parameter objects. If we added these to new service and plan fields on the instance itself, we'd spare lots of other bits of code throughout the broker from having to explicitly lookup services and plans by id themselves.

krancour:
Fixes #159 

This cuts down on the DRY exceptions for retrieving service and plan.


krancour:
Looks like a legit test failure. Will fix.

PR #221, opened 2018-01-22
gitignore .vscode/

krancour:
Discovered recently that this is a common file that exists, depending on what VS Code (common on our team) plugins are installed / enabled. Best to ensure it's .gitignored.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .glide/
+.vscode/
 bin/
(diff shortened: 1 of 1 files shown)


krancour:
@jeremyrickard, I'm proposing that this change doesn't require CI to run.

jeremyrickard:
Merging w/o CI.

PR #222, opened 2018-01-23
Changes to mysql bind.

### LINKED ISSUE #197: Created MySQL user missing the EXECUTE privilege
The created MySQL user is missing the ALL privilege.

I was under the impression that the user created here would have the same privileges as the default "azureuser". This is not the case as per the line below.
https://github.com/Azure/open-service-broker-azure/blob/60b1bb7940b5093c353dc1ecab3de0bf8386186c/pkg/services/mysqldb/bind.go#L55

For a user that is created with this, they will not have enough permissions to do the below, which is supported scenario in the documentation (https://docs.microsoft.com/en-us/azure/mysql/howto-create-users)

```
CREATE DATABASE testdb;
CREATE USER 'db_user'@'%' IDENTIFIED BY 'StrongPassword!';
GRANT ALL PRIVILEGES ON testdb . * TO 'db_user'@'%'; # <- ERROR 1044 (42000): Access denied for user 'vs9uigqscy'@'%' to database 'testdb'
FLUSH PRIVILEGES;
```

This breaks applications (for example, the Magento Helm chart).

jeremyrickard:
Change grant string in mysql bind to be bound to database
Removed global privs
Added Execute

Fixes #197 

The resulting permission list is still fairly permissive at the database level, but no longer has the global permissions.


krancour:
@jeremyrickard you and I talked through these changes and they make sense to me. Can we possibly validate them against a couple charts like Wordpress or Drupal to be sure these more tightly scoped permissions aren't, practically speaking, going to impede an average application (off the shelf or otherwise) from using this?

jeremyrickard:
@krancour I've already been testing the existing charts and they look good. I asked @sabbour if we could get his in progress Magneto chart as I would like to test the addition of the execute permission. 

As an aside, we should start some documentation about the services themselves and include the permissions granted with each bind. This should make it clear to consumers what is available and might be a good way for us to collect feedback once more people are consuming the broker.

sabbour:
The chart (work in progress) is here https://github.com/sabbour/helm-charts/tree/magento/magento

If you want me to test it against the new code, how should I do so? I can apply the modifications locally to my copy of OSBA code and build it if that's the process.

Otherwise maybe you test against my chart code directly?

jeremyrickard:
@sabbour thanks for sharing that. If you'd like to test your chart against this, I have a build prepared. You can update your broker via helm if you're up for it!

If you want to upgrade, you should be able to helm:

krancour:
> As an aside, we should start some documentation about the services themselves and include the permissions granted with each bind.

We have docs on each service, but they could stand to have this information added.

sabbour:
@jeremyrickard would that installation command be:

```
helm install azure/open-service-broker-azure --name osba --namespace osba \
    --set azure.subscriptionId=$AZURE_SUBSCRIPTION_ID \
    --set azure.tenantId=$AZURE_TENANT_ID \
    --set azure.clientId=$AZURE_CLIENT_ID \
    --set azure.clientSecret=$AZURE_CLIENT_SECRET \
     --set image.repository=jeremyrickard/azure-service-broker
    --set image.tag=canary
```

krancour:
@sabbour that's correct.

PR #223, opened 2018-01-23
[Proposal] Add migration plan for vm/db only manager for sqldb

zhongyi-zhang:
Hi @krancour and @arschles,
This PR is not requesting merging but only to better describe what we propose to add.
Just like what we discussed in the email thread, we can still support the migration from MASB based on your design for https://github.com/Azure/open-service-broker-azure/issues/124 by adding migration plans.
Besides the migration from MASB, users who want to integrate their existing Azure SQL services into the broker also benefit from it. We may consider that the operator already created an advanced configured server (e.g. with auditing output to a storage).
What do you think? I would like to work on it when you finish your works and merge the feature branch into master, if you like the idea.

### DIFF
--- a/pkg/services/sqldb/catalog.go
+++ b/pkg/services/sqldb/catalog.go
@@ -165,6 +165,12 @@ func (m *module) GetCatalog() (service.Catalog, error) {
 				Description: "Azure SQL Server VM Only",
 				Free:        false,
 			}),
+			service.NewPlan(&service.PlanProperties{
+				ID:          "2bff6bfc-8ee4-4893-8a99-db092928436d",
+				Name:        "migration",
+				Description: "Base on existing server on Azure",
+				Free:        false,
+			}),
 		),
 		// db only service
 		service.NewService(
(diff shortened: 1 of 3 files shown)


jeremyrickard:
@zhongyi-zhang if I follow what you propose, @krancour and I had been discussing doing just that.. essentially letting users "register" an existing SQL Server for the cases where customers have a non-standard or need to maintain it outside the context of the broker. We can make this available for later provisioning databases on top of. I aslo think it might be a way to address migration concerns.

zhongyi-zhang:
@jeremyrickard that's great! Expect your works.

jeremyrickard:
It's good that we're thinking in similar ways! To clarify, we haven't started implementing that, we've just been discussing it. We've given a build of #124 for evaluation and we'll wait for feedback on that before we merge anything in and I expect we'll return to this then. Will ping you then and we can work together 👍

zhongyi-zhang:
@jeremyrickard yeah, thanks!

PR #228, opened 2018-01-25
Change how the database name for cosmos db is generated

### LINKED ISSUE #227: Current CosmosDB implementation is incompatible with Azure CosmosDB name limits
There have been provisioning failures with CosmosDB that fail in ARM with a BadRequest and an error message that the name must be between 1 and 50 characters. We currently use a UUID. More info below per Zhongyi

The actual limit of the length of CosmosDB account name is [50 – length of region name – 1 (length of ‘-‘)].
For example,
In EastUS, the max length of the name is [50 – 6 – 1] = 43
In AustraliaSoutheast, the max length of the name is [50 – 18 – 1] = 31
I reported this to CosmosDB team several months ago, and they confirmed.
UUID as the account name in OSBA takes 36 of them, so provisioning could fail in some regions such as AustraliaSoutheast.
In @chzbrgr71's case, the region is "southcentralus"(length=14). The actual length of the account name is 14 + 36 + 1 = 51 > 50.

jeremyrickard:
A limitation in the current ARM based CosmosDB provisoning limits the length of the database name to 50 characters, including the name provided, the location name and a dash. In some situations, this means a UUID is invalid. If the length will be too long, this patch will randomly generate a string of the available length instead of using a UUID. 

Randomly select characters from a const containing 0-9 and a-z (lowercase). 

Fixes #227


krancour:
Buckle your seatbelt. There's a problem here and it's not easy to explain.

Run this example in the playground a few times and take note of the results:

https://play.golang.org/p/MtB7sGaEPpw

Did you get the same result every time? Was it this?

> 6fbbd7p95oe8ut5qrttebiwar88s74donnl9x8lipmar2rysjl

The results are deterministic because the default `rand.Source` (unexported package-level variable named `globalRand`) in the `math/rand` package is seeded with the integer `1`. Every time a Go program executes with the `Source` seeded like so, the results of pseudorandom number generation will be completely predictable.

There _is_ a `rand.Seed(...)` function that permits us to override the default seed of `1` with something sensible like `time.Now().UnixNano()`. The trouble with this is that we have no guarantee that some other package we use (now or in the future) doesn't stupidly set it back to `1` or `5` or any other hard-coded value that would result in predictable behavior.

So... at this point, the savvy Gopher says, "Ok. I won't user the global `rand.Intn(...)` convenience function. I'll just instantiate my _own_ seeded `Rand` like so:"

```
rnd := rand.New(rand.NewSource(time.Now().UnixNano()))
```

No one else can tamper with that, so it should be good, right?

No. 😢 

As it turns out, the above approach is not concurrency-safe. What's really galling is that the global `Rand` that we refused to use for _other_ reasons _is_ concurrency safe. So how do they do that? We can just do what they do, right?

Here's the line from the `math/rand` package where that "global" package-level `Rand` gets set up:

```
var globalRand = New(&lockedSource{src: NewSource(1).(Source64)})
```

So the `Source` is wrapped in a `lockedSource`. Cool.

```
type lockedSource struct {
	lk  sync.Mutex
	src Source64
}
```

Very nice. But you notice it's not exported. We can't use it.

So-- if we want to be adequately pseudorandom _and_ concurrency-safe, our only option is to create out _own_ concurrency-safe wrapper around a `Rand`... and that's exactly what's in our own `rand` package.

See `pkg/rand` and `pkg/generate` for some examples uses.

I've opened an issue against Go to highlight how eminently impractical the two out-of-the-box options (global `Rand` vs non-concurrency-safe `Rand`) both are.

https://github.com/golang/go/issues/21393

While that issue hasn't been closed, and likely will not be until Go 2, TIL that [this new package](https://godoc.org/golang.org/x/exp/rand) was created in response to numerous complaints about `math/rand`. Among the improvements, `LockedSource` is now exported... but its `src` attribute _isn't_, and since there's no exported constructor-like function either, it's _still_ not possible to to wrap a `LockedSource` around another `Source`. 😢 

Stick with our `pkg/rand`, I guess.

pkg/services/cosmosdb/provision.go
```diff
@@ -37,10 +41,32 @@ func (s *serviceManager) preProvision(
...
-	dt.DatabaseAccountName = uuid.NewV4().String()
+	dt.DatabaseAccountName = generateDatabaseName(instance.Location)
 	return dt, nil
 }
 
+func generateDatabaseName(location string) string {
+	databaseName := uuid.NewV4().String()
+	// CosmosDB currently limits database name to 50 characters,
+	// which includes location and a - character. Check if we will
+	// exceed this and truncate.
```
krancour: "truncate" might not be the best word choice for what we're doing.

krancour:
But the logic and the approach lgtm. Just need to account for the random nonsense.

jeremyrickard:
That lesson on Go 's math/rand package was a great way to start the day :) I didn't even see our rand package! I'll fix this up to use that.

pkg/generate/identifier.go
```diff
@@ -9,11 +9,16 @@ const (
 // role name, database name for various database systems-- including, at least,
 // PostgresSQL and MySQL and MSSQL.
 func NewIdentifier() string {
-	b := make([]byte, identifierLength)
+	return NewIdentifierOfLength(identifierLength)
+}
+
+// NewIdentifierOfLength generates an identifier of specified length.
+func NewIdentifierOfLength(length int) string {
```
krancour: This is where I wish Go had function overloading. Oh well.

krancour:
This looks spot on. LGTM. Will merge pending CI.

Issue #66, opened 2017-11-17
Respond with error message if operation fails


zhongyi-zhang:
For example: I didn't specified location in provisioning parameters, it would return:
<pre>
$ cf create-service azure-sqldb basic sqlsql
Creating service instance sqlsql in org IP.xip.io_ORGANIZATION / space azure as admin...
FAILED
Server error, status code: 502, error code: 10001, message: The service broker rejected the request to http://asb.IP.xip.io/v2/service_instances/ace140b2-79dc-4ed3-bacb-c4eba0eb67dc?accepts_incomplete=true.
<b>Status Code: 400 Bad Request, Body: {}</b>
</pre>
Relative line:
https://github.com/Azure/azure-service-broker/blob/master/pkg/api/provision.go#L225

See bold, user only knows that it is a bad request but what's the reason in details? User needs to use `cf logs asb --recent` to find out that.
At worst, some other polling operations are working at the same time, maybe it's a little hard to find the corresponding error message. (PS, such an error is printed to stdout, is it expected?)
<pre>
2017-11-17T02:03:31.02+0000 [APP/PROC/WEB/0] <b>OUT</b> time="2017-11-17T02:03:31Z" level=debug msg="bad provisioning request: validation error" field=location instanceID=ace140b2-79dc-4ed3-bacb-c4eba0eb67dc issue="invalid location: \"\""
</pre>

Anyway, it's better to always respond with error message if operation failed.

carolynvs:
👍  I ran into this as well. This is the catalog response I get when forgetting to set a required param:

```json
{
      "metadata": {
        "name": "noparams",
        "namespace": "default",
        "selfLink": "/apis/servicecatalog.k8s.io/v1beta1/namespaces/default/serviceinstances/noparams",
        "uid": "d4d93910-e509-11e7-926b-0242ac110006",
        "resourceVersion": "1214",
        "generation": 1,
        "creationTimestamp": "2017-12-19T22:13:24Z",
        "finalizers": [
          "kubernetes-incubator/service-catalog"
        ]
      },
      "spec": {
        "clusterServiceClassExternalName": "azure-mysqldb",
        "clusterServicePlanExternalName": "standard800",
        "clusterServiceClassRef": {
          "name": "997b8372-8dac-40ac-ae65-758b4a5075a5"
        },
        "clusterServicePlanRef": {
          "name": "08e4b43a-36bc-447e-a81f-8202b13e339c"
        },
        "parameters": {},
[...]

PR #229, opened 2018-01-25
Replacing empty json responses with error messages.

jeremyrickard:
Issue #66 points out that our error messages don't help determine the
cause of operation failures. This PR adds an informative response to
locations where we are returning a 400 or a 409 so the user has some extra infromation
to debug the failure.

These should be inline with the [Broker Errors](https://github.com/openservicebrokerapi/servicebroker/blob/v2.13/spec.md#broker-errors)  section of the OSB spec.


pkg/api/responses.go
```diff
@@ -69,3 +72,16 @@ var responseOperationInvalid = []byte(
...
+)
+
+var responseIncorrectRequestBody = []byte(`{ "error": "MalformedRequest", ` +
+	`"description": The provided request did not match what was expected for ` +
+	`the service" }`,
+)
+
+var responseValidationFailedTemplate = string(`{ "error": "ValidationError", ` +
+	`"description": The value provided for %s is invalid. %s" }`,
+)
```
krancour: Could we replace this with a function that does the formatting so users of `responseValidationFailedTemplate` aren't all individually responsible for that? Maybe make it that we pass it a validation error and it returns the response body?

If we were to do that, should we use functions to access the _other_ response bodies as well? I'm thinking yes, both on account of consistency and the likelihood that we might want to further parameterize all of these responses.

krancour: Some kind of escaping might also be needed on the invalid value. What if it contains a double quote? It would mess up the well-formedness (is that a word?) of the response.

Could it possibly be better to use a struct (even an anonymous one) to encapsulate the details of this response and then we just marshal it to JSON and count on the marshaler to get all the escaping right?

jeremyrickard: 👍 * 2

PR #232, opened 2018-02-05
merge latest from master to poc branch

krancour:


krancour:
@jeremyrickard fyi, see last two commits. I snuck two bug fixes into here.

PR #243, opened 2018-02-06
Parent child services

jeremyrickard:
This PR brings over our changes for supporting parent child services and a refactor of the SQLDDB service to support three distinct services:

* All In One (server and db)
* Server Only (only the server VM, must define an alias)
* VM only (requires a parent declaration)


pkg/api/deprovision.go
```diff
@@ -119,21 +120,46 @@ func (s *server) deprovision(w http.ResponseWriter, r *http.Request) {
...
+		)
+	}
 	if err = s.asyncEngine.SubmitTask(task); err != nil {
 		logFields["step"] = firstStepName
 		logFields["error"] = err
 		log.WithFields(logFields).Error(
 			"deprovisioning error: error submitting deprovisioning task",
 		)
 		s.writeResponse(w, http.StatusInternalServerError, responseEmptyJSON)
-		return
```
krancour: I don't think this return statement should have been removed.

jeremyrickard: Weird! I'll add that back. I'll scan through the rest of these too

pkg/service/catalog.go
```diff
@@ -39,9 +41,12 @@ type Service interface {
 	ToJSON() ([]byte, error)
 	GetID() string
 	GetName() string
+	GetBindable() bool
```
krancour: Can we change this to `IsBindable()`? That seems more conventional when you're asking a yes/no question and expecting a boolean response.

pkg/service/instance.go
```diff
@@ -22,6 +24,7 @@ type Instance struct {
 	StatusReason                    string                 `json:"statusReason"`
 	Location                        string                 `json:"location"`
 	ResourceGroup                   string                 `json:"resourceGroup"`
+	ParentAlias                     string                 `json:"parentAlias"`
```
krancour: Can we move this up above `Parent`... or else move `Parent` down below this? I just want to keep them together, so it's more obvious they're related.

krancour:
LGTM. Will merge pending CI and some manual e2e on my end.

PR #245, opened 2018-02-07
[docs] remove kube-system serviceaccount cluster-admin role

prydonius:
Looks like this isn't really needed. The only reason it may have been there, from what I can tell, is for Tiller but the following steps install Tiller with a specific service account. Nothing else is installed in the kube-system namespace.

### DIFF
--- a/docs/quickstart-minikube.md
+++ b/docs/quickstart-minikube.md
@@ -178,10 +178,6 @@ Next we will create a local cluster using Minikube. You can also [try OSBA on th
     ```console
     minikube start --extra-config=apiserver.Authorization.Mode=RBAC
     ```
-1. Grant the `cluster-admin` role to the default system account:
-    ```console
-    kubectl create clusterrolebinding cluster-admin:kube-system --clusterrole=cluster-admin --serviceaccount=kube-system:default
-    ```
 
 ### Configure the cluster with Open Service Broker for Azure
 


jeremyrickard:
@prydonius thanks for the PR! I'll run through this from scratch to verify and take care of the PR. We also added a new PR for getting started on AKS, which essentially has the same steps. Could you also update that one?

Thanks!

krancour:
So... @prydonius is correct, because of this imperative later on in the quickstart:

```
helm init --service-account tiller
```

I'm more inclined to ask _why_ we're doing it that way.

@prydonius might know better than we do, but in the average case, don't people initialize Helm/Tiller without using the `--service-account` flag?

If I'm right that that's not usual and we drop that flag from this line, is the cluster role binding specified in the quickstart then correct?

prydonius:
> don't people initialize Helm/Tiller without using the --service-account flag?

Yes, but only for historical (i.e. pre-RBAC) reasons. It's much better to give Tiller it's own SA with the extended privileges it needs, than to give the whole kube-system namespace cluster-admin permissions.

On AKS, you'd probably want an even more secure installation of Tiller (see https://github.com/kubernetes/helm/blob/master/docs/securing_installation.md). I'll add a note of this in the AKS quickstart guide.

prydonius:
@jeremyrickard actually it looks like AKS doesn't yet support RBAC so it doesn't make sense to change the guide until it does.

jeremyrickard:
@prydonius lol yeah, i amended my comment just now. It was too early for me :)

jeremyrickard:
NO CI needed, docs only.

krancour:
Thanks @prydonius!

PR #255, opened 2018-02-08
add support for CF to configure basic auth via tiles

### LINKED ISSUE #129: The environment name of basic authentication is not same with meta-azure-service-broker
In open-service-broker-azure, the environment names are:
```
BASIC_AUTH_USERNAME
BASIC_AUTH_PASSWORD
```

In meta-azure-service-broker, the environment names are:
```
SECURITY_USER_NAME
SECURITY_USER_PASSWORD
```
These names are required for Pivotal Cloud Foundry. Please refer to https://docs.pivotal.io/tiledev/1-12/tile-generator.html#packages.
```
Your broker is automatically registered with the Cloud Controller. The Cloud Controller invokes your broker’s endpoints, and it uses basic authentication to secure those API calls. The credentials it uses are passed to your broker in two environment variables:

SECURITY_USER_NAME
SECURITY_USER_PASSWORD
Your broker is expected to accept those credentials. If it doesn’t, automatic broker registration fails.
```

Is it possible to change to `SECURITY_USER_NAME` and `SECURITY_USER_PASSWORD`? Thanks.

krancour:
Fixes #129 

What I wanted to do here was add this option, as requested, without fully allowing CF to dictate the environment variables we choose to use. As such, I've made it that the original environment variables `BASIC_AUTH_USERNAME` and `BASIC_AUTH_PASSWORD` are respected and take precedence, but when those aren't set, we fall back to `SECURITY_USER_NAME` and `SECURITY_USER_PASSWORD`.

This increases the complexity of the basic auth config struct a bit, since it contains exported fields (`envconfig` requires this) that we would have preferred to be internal and maybe we don't actually want clients of our `config` package reading. Honestly, this was already a problem in a few other spots as well...

So, as a pre-requisite for fixing this, I've added exported interfaces for each type of config struct, while the config structs themselves remain unexported. Bonus-- the `GetX()` functions that need to be added to those interfaces and their respective implementation make this approach tenable _also_ effectively make the config structs immutable by clients that use them. Win, win, win.

### DIFF
--- a/cmd/broker/broker.go
+++ b/cmd/broker/broker.go
@@ -136,8 +136,8 @@ func main() {
 		filterChain,
 		modules,
 		modulesConfig.GetMinStability(),
-		azureConfig.DefaultLocation,
-		azureConfig.DefaultResourceGroup,
+		azureConfig.GetDefaultLocation(),
+		azureConfig.GetDefaultResourceGroup(),
 	)
 	if err != nil {
 		log.Fatal(err)
--- a/cmd/broker/broker.go
+++ b/cmd/broker/broker.go
@@ -105,7 +105,7 @@ func main() {
 	if err != nil {
 		log.Fatal(err)
 	}
-	codec, err := aes256.NewCodec([]byte(cryptoConfig.AES256Key))
+	codec, err := aes256.NewCodec([]byte(cryptoConfig.GetAES256Key()))
 	if err != nil {
 		log.Fatal(err)
 	}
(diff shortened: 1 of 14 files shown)


jeremyrickard:
@zhongyi-zhang check it out.

zhongyi-zhang:
Cool! Thank you a lot @krancour and @jeremyrickard. I'll create a PR with the PCF tile stuff for tile generating later.

PR #263, opened 2018-02-13
Refactor of MySQL to support parent/child

### LINKED ISSUE #215: Allow multiple MySQL databases on single MySQL server
Pending validation of the work done in the feature branch for #124, the same approach can be applied to MySQL.

jeremyrickard:
1.) refactor of existing code to support parent/child
  * Add two new services to catalog.go
  * Renamed existing service manager, added 2 new ones
  * Refactor for naming
  * Added new methods for new ServiceManagers
  * Tried to extract common code to funcs where possible
2.) added bind tests

Closes #215


pkg/services/mysqldb/arm_template.go
```diff
@@ -131,3 +131,165 @@ var armTemplateBytes = []byte(`
...
+				"apiVersion": "2017-04-30-preview",
+				"type": "Microsoft.DBforMySQL/servers/databases",
+				"name": "[concat(parameters('serverName'), '/', parameters('databaseName'))]",
+				"properties": {}
+			}
+		],
+		"outputs": {
+		}
+	}
+	`)
```
krancour: In the sqldb module, we separated these ARM templates into three files. Can we do the same here, just for consistency?

jeremyrickard: yeah, can do.

pkg/services/mysqldb/bind.go
```diff
@@ -7,29 +7,113 @@ import (
...
+func (a *allInOneManager) ValidateBindingParameters(
 	bindingParameters service.BindingParameters,
 ) error {
 	// There are no parameters for binding to MySQL, so there is nothing
 	// to validate
 	return nil
 }
 
-func (s *serviceManager) Bind(
+func (v *vmOnlyManager) ValidateBindingParameters(
```
krancour: This matches what we did in sqldb, but I think we should change the name of this manager (and the service it manages) here (and, separately, in the sqldb module as well). "VM" implies some things that might not actually be true for certain plans.

Ok... we can take an educated guess that there are VMs involved (currently), but in the future, these might be containerized and the end-user would be none the wiser. Also, it's possible (assuming support for geo-redundancy, master/slave, etc.) that there are _multiple_ VMs/containers involved.

Let me know if you hate this name, but I think the right nomenclature to opt for here might be `dbmsOnlyManager`. I feel as if that will accurately convey what it is we're managing without implying anything about how that's realized.

pkg/services/mysqldb/bind.go
```diff
@@ -94,3 +178,42 @@ func (s *serviceManager) GetCredentials(
...
+	}
+
+	return &Credentials{
+		Host:     pdt.FullyQualifiedDomainName,
+		Port:     3306,
+		Database: dt.DatabaseName,
+		Username: fmt.Sprintf("%s@%s", bd.LoginName, pdt.ServerName),
+		Password: bd.Password,
+	}, nil
+}
```
krancour: I know we didn't do this for sqldb, but perhaps should have (and can address in a separate pr), but what would you think about breaking this file (and other similar files) up into three that are each prefixed with the manager name? I feel like the code would be easier to follow (for the purposed of review, for instance) if code for the three managers weren't interspersed in one common set of files.

jeremyrickard: Yeah! I thought we should have done it in sqldb, so this is cool with me!

pkg/services/mysqldb/bind.go
jeremyrickard: That's good with me

pkg/services/mysqldb/catalog.go
```diff
@@ -131,5 +131,164 @@ func (m *module) GetCatalog() (service.Catalog, error) {
...
+			service.NewPlan(&service.PlanProperties{
+				ID:          "3f65ebf9-ac1d-4e77-b9bf-918889a4482b",
+				Name:        "basic50",
+				Description: "Basic Tier, 50 DTUs.",
+				Free:        false,
+				Extended: map[string]interface{}{
+					"skuName":        "MYSQLB50",
+					"skuTier":        "Basic",
+					"skuCapacityDTU": 50,
+					"skuSizeMB":      51200,
```
krancour: fyi only... this bothers me. I feel like the sku name "MYSQLB50" already conveys the tier and the number of DTUs. But, afaik, we _need_ to provide all of these values when provisioning via ARM. (Or that was true at one time.) Separate from this PR, we should experiment with this and maybe even talk to the MySQL team to see if they can / plan to streamline this. Picking a sku for other services isn't this difficult.

krancour:
This looks fantastic.

PR #265, opened 2018-02-13
Fixed deprovision of DB only to not kill the server

### LINKED ISSUE #264: mysql db-only deprovision should not delete the entire dbms
This is a follow-up to #263

The db-only service manager for the mysql module is currently deleting the entire dbms upon deprovision, but it should only delete the database.

jeremyrickard:
Fixes #264 

Replaces use of ServersClient in DB deprovisioning with DatabasesClient. Wired that through from cmd/broker/modules.go

### DIFF
--- a/cmd/broker/modules.go
+++ b/cmd/broker/modules.go
@@ -106,6 +106,13 @@ func initModules(azureConfig config.AzureConfig) error {
 	mysqlServersClient.Authorizer = authorizer
 	mysqlServersClient.UserAgent = getUserAgent(mysqlServersClient.Client)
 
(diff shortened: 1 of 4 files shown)


krancour:
@jeremyrickard can we change this line in `deprovision_db_only.go`?

```
service.NewDeprovisioningStep("deleteMySQLServer", d.deleteMySQLServer),
```

It should probably read like so:

```
service.NewDeprovisioningStep("deleteMySQLDB", d.deleteMySQLDB),
```

krancour:
LGTM. Sequence-wise, let's get this in before #266.

PR #269, opened 2018-02-14
DO NOT MERGE: Experiment with rolling back the SDK upgrades

krancour:
This PR should never be merged. It's opened merely to take advantage of our CI. I want to run CI many times against OSBA as it existed prior to "The Great SDK Upgrade of 2018" in order to determine whether the upgraded SDKs are a factor in recent perceived slowness in provisioning in Azure services, most notably with deprovisioning CosmosDB and Redis.


krancour:
I'm re-running CI on this each time we notice newer builds experiencing issues. My hope is to prove that all issues are unrelated to the upgraded SDKs. So far, based on just a few runs, this does appear to be the case, but I'm not confident enough in that assessment to close this yet.

krancour:
I'm closing this PR. Over the last few weeks, I am quite satisfied that issues integrating with various Azure services are _not_ more common with the newer SDK. If it's called into question again, this PR can always be resurrected, but I'm closing it for now because it's noise.

PR #273, opened 2018-02-15
Refactor of postgres module

### LINKED ISSUE #214: Allow multiple PostgreSQL databases on single PostgreSQL server
Pending validation of the work done in the feature branch for #124, the same approach can be applied to PostgreSQL.

jeremyrickard:
1.) Add two new services (dbms only and db only)
2.) Refactored life cycle things into separate files
3.) Added databases client to cmd/broker/modules.go and wired thorugh
4.) Added binding test
5.) Renamed mysql binding test

Closes #214


cmd/broker/modules.go
```diff
@@ -132,9 +132,18 @@ func initModules(azureConfig config.AzureConfig) error {
 		azureEnvironment.ResourceManagerEndpoint,
 		azureSubscriptionID,
 	)
+
```
krancour: Total nit: Let's remove this blank line and the one a few lines down from here. Visually, it's probably beneficial to keep related clients (e.g. all Postgres clients) grouped together.

Or ignore this suggestion. I realize how anal I'm being.

pkg/services/postgresqldb/types_dbms_only.go
```diff
@@ -0,0 +1,47 @@
+package postgresqldb
+
+import "github.com/Azure/open-service-broker-azure/pkg/service"
+
+// ServerProvisioningParameters encapsulates PostgreSQL-specific dbms
+// provisioning options
+type ServerProvisioningParameters struct {
```
krancour: Can we change this to `DBMSOnlyProvisioningParameters`?

jeremyrickard: Yeah good call. I originally had Server and Database embedded for the AllInOne and thought this read better, forgot to change the name.

pkg/services/postgresqldb/types_all_in_one.go
```diff
@@ -0,0 +1,47 @@
...
+// PostgreSQL-specific provisioning options
+type AllInOneProvisioningParameters struct {
+	SSLEnforcement  string   `json:"sslEnforcement"`
+	FirewallIPStart string   `json:"firewallStartIPAddress"`
+	FirewallIPEnd   string   `json:"firewallEndIPAddress"`
+	Extensions      []string `json:"extensions"`
+}
+
+type allInOnePostgresqlInstanceDetails struct {
+	dbmsOnlyPostgresqlInstanceDetails
```
krancour: Any reason to do this here and _not_ use the same trick for the provisioning params above? (There may be. I'm asking.)

jeremyrickard: No, I think I decided to unroll it for readability and forgot to clean this one up. I can go back to the embedded version if you'd like!

pkg/services/postgresqldb/deprovision_db_only.go
```diff
@@ -0,0 +1,74 @@
...
+			"error casting instance.Details " +
+				"as *dbOnlyPostgresqlInstanceDetails",
+		)
+	}
+	result, err := d.databasesClient.Delete(
+		ctx,
+		instance.Parent.ResourceGroup,
+		pdt.ServerName,
+		dt.DatabaseName,
+	)
```
krancour: Does this work out well even though, after provisioning, we modified this DB some? (We change it to be owned by a certain role.) It probably is ok. Legit asking though because I haven't tried it.

pkg/services/postgresqldb/postgresql.go
```diff
@@ -23,9 +37,20 @@ func New(
 	armDeployer arm.Deployer,
 	checkNameAvailabilityClient postgresSDK.CheckNameAvailabilityClient,
 	serversClient postgresSDK.ServersClient,
+	databasesClient postgresSDK.DatabasesClient,
```
krancour: Why does the dbms-only manager need the database client?

krancour: In fact, it doesn't look like you ever set a value for this attribute farther below when you initialize this manager. I'm guessing this attribute is extraneous.

pkg/services/postgresqldb/provision_common.go
```diff
@@ -31,3 +42,172 @@ func getAvailableServerName(
...
+				extension,
+				err,
+			)
+		}
+	}
+	if err = tx.Commit(); err != nil {
+		return fmt.Errorf("error committing transaction: %s", err)
+	}
+	return nil
+}
```
krancour: Thanks for keeping this all DRY. :)

pkg/services/postgresqldb/postgresql.go
jeremyrickard: I think this is how the diff is rendering it. It's passed into the constructor like thing and used by the *db* service manager because it does need it. 

```
func New(
	armDeployer arm.Deployer,
	checkNameAvailabilityClient postgresSDK.CheckNameAvailabilityClient,
	serversClient postgresSDK.ServersClient,
	databasesClient postgresSDK.DatabasesClient,
) service.Module {
	return &module{
		allInOneManager: &allInOneManager{
			armDeployer:                 armDeployer,
			checkNameAvailabilityClient: checkNameAvailabilityClient,
			serversClient:               serversClient,
		},
		dbOnlyManager: &dbOnlyManager{
			armDeployer:                 armDeployer,
			checkNameAvailabilityClient: checkNameAvailabilityClient,
			databasesClient:             databasesClient,
		},
		dbmsOnlyManager: &dbmsOnlyManager{
			armDeployer:                 armDeployer,
			checkNameAvailabilityClient: checkNameAvailabilityClient,
			serversClient:               serversClient,
		},
	}
}
```

krancour: Yep. You're right. I "unfolded" the lines above for a better point of reference and you're right. I'm looking at the args list for the module's "constructor." Got it.

pkg/services/postgresqldb/provision_dbms_only.go
```diff
@@ -0,0 +1,160 @@
...
+func (d *dbmsOnlyManager) GetProvisioner(
+	service.Plan,
+) (service.Provisioner, error) {
+	return service.NewProvisioner(
+		service.NewProvisioningStep("preProvision", d.preProvision),
+		service.NewProvisioningStep("deployARMTemplate", d.deployARMTemplate),
+	)
+}
+
+func (d *dbmsOnlyManager) preProvision(
```
krancour: It might not work out, but is it possible to DRY this up? There's a lot in common with the all-in-one pre-provisioning here. The thing that might be problematic is that the type of object that the values are being assigned to varies between the two managers, but _maybe_ we can still cope with that if the all-in-one details are composed of dbms details?? Maybe? i.e. Can we extract most of this logic and _safely_ do a type assertion that a all-in-one-details are db-only details?

krancour: If you think it makes things harder to grok, I'm also ok with leaving it be in the name of not overcomplicating things. Dealer's choice here.

jeremyrickard: yeah, let me see if that's doable!

pkg/services/postgresqldb/types_all_in_one.go
krancour: Only if it doesn't complicate things. Sometimes that does. Dealer's choice.

pkg/services/postgresqldb/types_dbms_only.go
jeremyrickard: I kept this as ServerProvisioningParameters, I think it reads better in the embedding.

pkg/services/postgresqldb/provision_all_in_one_test.go
```diff
@@ -16,19 +16,18 @@ func TestValidateNoFirewallConfig(t *testing.T) {
...
 func TestValidateGoodFirewallConfig(t *testing.T) {
 	sm := &allInOneManager{}
-	pp := &AllInOneProvisioningParameters{
-		FirewallIPStart: "192.168.86.1",
-		FirewallIPEnd:   "192.168.86.100",
-	}
+	pp := &AllInOneProvisioningParameters{}
+	pp.FirewallIPStart = "192.168.86.1"
+	pp.FirewallIPEnd = "192.168.86.100"
+
```
krancour: This isn't wrong, but I'm curious about what motivated this change.

pkg/services/postgresqldb/types_all_in_one.go
```diff
@@ -5,10 +5,8 @@ import "github.com/Azure/open-service-broker-azure/pkg/service"
 // AllInOneProvisioningParameters encapsulates both dbms and database
 // PostgreSQL-specific provisioning options
 type AllInOneProvisioningParameters struct {
-	SSLEnforcement  string   `json:"sslEnforcement"`
-	FirewallIPStart string   `json:"firewallStartIPAddress"`
-	FirewallIPEnd   string   `json:"firewallEndIPAddress"`
-	Extensions      []string `json:"extensions"`
+	ServerProvisioningParameters
+	DatabaseProvisioningParameters
```
krancour: This is sweet.

Wondering if we want to change `ServerProvisioningParameters` to `DMSProvisioningParameters`. Dealer's choice.

pkg/services/postgresqldb/provision_all_in_one_test.go
krancour: I see why now.

pkg/services/postgresqldb/types_all_in_one.go
krancour: Oh. I see your rationale for this in another comment.

krancour: I might want to revisit this in the future, but it's certainly not blocking this PR.

PR #274, opened 2018-02-15
only encrypt sensitive details

krancour:
Might need to do something similar for params and bindings, but this PR is already big enough. Will tackle those in follow-ups.


krancour:
I'd like to get #273 merged before this one.

PR #278, opened 2018-02-16
Adding to more info to MySQL and PostgreSQL bind credentials and making lifecycle tests better

### LINKED ISSUE #275: Extend MySQL and PostgreSQL Modules to provide additional info
Currently, the mysql and postgres tests fail if ssl is enabled. The test methods don't have knowledge of ssl being enabled, so they fail. We should enhance this to exercise that. We should enhance bind to return sufficient information in the credential to determine this. Ideally, we should provide additional information to satisfy the [recommendations](https://docs.cloudfoundry.org/services/binding-credentials.html) from CF. This would include a uri, tags, and a flag to indicate if the connection requires SSL. This will help us support [Spring based applications a little better on CF](https://cloud.spring.io/spring-cloud-connectors/spring-cloud-cloud-foundry-connector.html), and will enable us to provide better coverage in lifecycle tests.

jeremyrickard:
1.) Added more info to the MySQL and PostgreSQL bind credentials.
    - uri for compliance with CF recommendations
    - SSLRequired to indicate if the connection should be done
      over SLS
2.) Enhanced lifecycle tests to verify
3.) small DRY refactor to support this

Closes #275

### DIFF
--- a/pkg/services/mysqldb/bind_common.go
+++ b/pkg/services/mysqldb/bind_common.go
@@ -75,7 +75,6 @@ func createCredential(
 	sslRequired bool,
 	serverName string,
 	databaseName string,
-	dnsSuffix string,
 	bindingDetails *mysqlBindingDetails,
 	secureBidningDetails *mysqlSecureBindingDetails,
 ) *Credentials {
@@ -96,6 +95,5 @@ func createCredential(
 		Password:    secureBidningDetails.Password,
 		SSLRequired: sslRequired,
 		URI:         connectionString,
-		DNSSuffix:   dnsSuffix,
 	}
 }
--- a/pkg/services/mysqldb/bind_common.go
+++ b/pkg/services/mysqldb/bind_common.go
@@ -2,6 +2,7 @@ package mysqldb
 
 import (
 	"fmt"
+	"net/url"
 
 	"github.com/Azure/open-service-broker-azure/pkg/generate"
 )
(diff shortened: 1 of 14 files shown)


pkg/services/mysqldb/bind_common.go
```diff
@@ -0,0 +1,95 @@
...
+
+func createCredential(
+	fqdn string,
+	sslRequired bool,
+	serverName string,
+	databaseName string,
+	bindingDetails *mysqlBindingDetails,
+) *Credentials {
+	username := fmt.Sprintf("%s@%s", bindingDetails.LoginName, serverName)
+	connectionTemplate := "mysql://%s:%s@%s:3306/%s?allowNativePasswords=true"
```
jeremyrickard: actually not sure if i should include allowNativePasswords here? I don't think that's a mysql option outside of the go lib?

krancour: Not sure I know what that setting does...

krancour:
@jeremyrickard are similar changes advisable for sqldb?

jeremyrickard:
Re sqldb changes, yeah i would imagine so. I need to look up what's needed so we can make sure we include it all, so i was planning on doing a separate PR for that.

pkg/services/mysqldb/bind_common.go
jeremyrickard: w/ the golang lib at least, if you *don't* include thatt mysql complains. I'll try it with java before i merge this and see what's up.

krancour: Ok. lmk. Apart from this open question, this LGTM.

PR #286, opened 2018-02-20
Better binding sqldb

jeremyrickard:
* Refactor to clean up a little DRY
* Expansion of the binding creds to give a CF style uri and jdbcUrl 

*A note on the addition of jdbcUrl*

I think adding the jdbcUrl makes the [SQL Server CF Cloud Connector](https://cloud.spring.io/spring-cloud-connectors/spring-cloud-cloud-foundry-connector.html#_sql_server) work, but the URI won't be sufficient because it doesn't look like the connector handles the query part of the URI, hence won't get the encrypt/host verification params that seem necessary from the docs. In particular, [SqlServerServiceInfo.buildJdbcUrl](https://github.com/spring-cloud/spring-cloud-connectors/blob/master/spring-cloud-core/src/main/java/org/springframework/cloud/service/common/SqlServerServiceInfo.java) doesn't really handle the query. If we provide a jdbcUrl, it will use that vs parsing it out. I don't think that it's harmful in adding this and I think it will probably make it more useful to any one using the broker from CF using Java. MASB also returns that. Should we make all the credentials align with MASB?


krancour:
@jeremyrickard first, this PR is chaulk full of good stuff. Thanks. LGTM as is.

> I think adding the jdbcUrl makes the SQL Server CF Cloud Connector work, but the URI won't be sufficient because it doesn't look like the connector handles the query part of the URI, hence won't get the encrypt/host verification params that seem necessary from the docs. In particular, SqlServerServiceInfo.buildJdbcUrl doesn't really handle the query. If we provide a jdbcUrl, it will use that vs parsing it out. I don't think that it's harmful in adding this and I think it will probably make it more useful to any one using the broker from CF using Java. MASB also returns that. Should we make all the credentials align with MASB?

Ordinarily, I think I'd say these credentials contain enough detail that _anyone_ should be able to construct _any kind of_ connection string they need themselves. That said, MASB was our forebear and while we can't quite manage to be 100% backwards compatible, anywhere where we can make the transition from MASB to OSBA a tiny bit easier (for anyone who has chosen to pursue that) I'm 👌 on that.

So again... LGTM.

Fine with follow-ups re: the question above if the need becomes clear later.

PR #290, opened 2018-02-21
Update the quickstart to say to use Minikube 0.25 

jeremyrickard:
Update the quick start to add instructions to use minikube v0.25 and to use the kubeadm bootstrapper instead of localkube.

### DIFF
--- a/docs/quickstart-minikube.md
+++ b/docs/quickstart-minikube.md
@@ -176,7 +176,7 @@ Next we will create a local cluster using Minikube. You can also [try OSBA on th
 
 1. Create a Minikube Cluster:
     ```console
-    minikube start --kubernetes-version=v1.9.2 --bootstrapper=kubeadm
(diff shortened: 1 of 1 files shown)


krancour:
Does the kube-adm bootstrapper enable RBAC by default?

jeremyrickard:
Yep, kubeadm enables RBAC by default!

jeremyrickard:
If you don't specify a version, it defaults to k8s 1.9.0, i suppose we could go with that to make the command simpler?

carolynvs:
FYI, I ran this both with v1.9.2 and leaving off the version.

krancour:
If we're not dependent on bug fixes in 1.9.1 or 1.9.2, let use the default kube version to avoid having to update that aspect of this doc too frequently. Otherwise, LGTM.

jeremyrickard:
Version removed! LGTM and I'll merge, sans PR.

PR #304, opened 2018-02-26
Add PCF tile stuff

zhongyi-zhang:
The tile is just wrapping the broker to allow it to be installed by PCF Ops Manager.

The README.md is about how to generate a new tile.
For example, there is no generating history for now. Running `./generate-tile.sh -major` will generate a new tile named `open-service-broker-azure-1.0.0.pivotal` and a generating history file `tile-history.yml` (this file should also be checked in after generating).
Then, `./generate-tile.sh -major` can generate tile **2.0.0**, `./generate-tile.sh -minor` can generate tile **1.1.0**, and `./generate-tile.sh` can generate tile **1.0.1**.

User installation steps:

1. Importing the tile into Ops Manager:
image

1. Find the imported broker app:
image

1. Fill the configuration mapped to the environment variables just like filling the [CF manifest](https://github.com/Azure/open-service-broker-azure/blob/master/contrib/cf/manifest.yml):
image

1. Apply the change and wait for success.

After these steps, users can use CF CLI to create/delete/.. service instances. It is the same as OSS CF.

BTW, I have some questions...
When will we get started to include the tile in new releases? When it reaches 1.0?
@seanmck, would you help to go through the Pivotal process to publish the tile?


contrib/cf/pcf-tile/tile.yml
```diff
@@ -0,0 +1,158 @@
+---
+# The high-level description of your tile.
+# Replace these properties with real values.
+#
+name: open-service-broker-azure
+icon_file: resources/icon.png
+label: Microsoft Open Service Broker of Azure
```
seanmck: This should be Microsoft Open Service Broker *for* Azure, not *of* Azure.

contrib/cf/pcf-tile/README.md
```diff
@@ -0,0 +1,26 @@
...
+  ```
+  ./generate-tile.sh
+  ```
+
+  Note:
+    * Build a major version: ./generate-tile.sh -major
+    * Build a minor version: ./generate-tile.sh -minor
+    * Build a patch version: ./generate-tile.sh
+
+1. You can find open-service-broker-azure-`VERSION`.pivotal under the folder `product`.
```
krancour: Disclaimer: I'm not a CF guy, really, but I feel like these directions are, possibly, more complicated than they need to be. I have a few thoughts about how to streamline the UX for generating a tile.

1. Is there an existing Docker image that contains the tooling required (e.g. python2, tile-generator, and bosh)? If not, could you consider creating one as its own open source project and publishing that image. It seems it would be generally useful.
1. If the above is addressed, could you consider adding a new make target that will mount project source into a container that contains all the tooling and running the tile-generating script _within_ that container?

For a user, this could streamline the UX to something as simple as:

```
$ make generate-pcf-tile
```

This makes an (fair, I think) assumption that Docker is a more ubiquitous tool than the virtualenv, pythons2, tile-generator, BOSH chain.

contrib/cf/pcf-tile/tile.yml
zhongyi-zhang: Thanks! Fixed.

contrib/cf/pcf-tile/README.md
zhongyi-zhang: @krancour OK. I'll try to make it work as what you suggest!

krancour:
Hey, @zhongyi-zhang the Docker stuff seems like a step in the right direction. Thanks for doing that. And I'm really happy to see that there was an existing image with the tool chain for generating tiles! This is great.

What I think might make this easier is if you do not use Docker Compose for this. Compose is useful for scenarios where you need to coordinate the action of multiple containers. For instance, the broker uses it to launch tests (or the broker itself) that are connected to a Redis container...

Since all of your generation tasks all involve only one container, you can skip all the complexity and overhead of Compose and just use plain Docker.

Sound ok? It will make this PR a lot smaller.

zhongyi-zhang:
@krancour thanks for the explanation about Compose! It was refined to use Docker only now.

krancour:
@zhongyi-zhang-- nice work. This looks like ehat I was imagining. Will review more thoroughly in the morning. Thanks!

contrib/cf/pcf-tile/Makefile
```diff
@@ -0,0 +1,38 @@
...
+	@if [ -z $$(which docker) ]; then \
+		echo "Missing \`docker\` client which is required for development"; \
+		exit 2; \
+	fi
+
+.PHONY: pull-tile-generator
+pull-tile-generator: check-docker
+	docker pull cfplatformeng/tile-generator:v11.0.5
+
+SHELL = /bin/bash
```
krancour: Can we move all variable definitions like this one up to the top of the Makefile, please?

krancour:
@zhongyi-zhang this looks really good. Thanks!

I tried this process out and it worked seamlessly! Ya!

It did generate new artifacts that I think should probably be added to `.gitignore` so that someone who has run this locally doesn't accidentally include those artifacts with a commit.

I do have some follow-up questions about this. You will have to forgive my relative degree of inexperience with CF...

_Who_ are these new make targets intended for? Is the expectation that those operating CF clusters will clone this repo and generate the tile themselves? Or should _we_ be generating the tile and publishing it somewhere? My gut tells me it is the latter-- in which case, could you help us identify _where_ it should be pushed to so we can build that into our CI/CD pipeline?

contrib/cf/pcf-tile/Makefile
zhongyi-zhang: Thanks for the suggestion, will do soon.

zhongyi-zhang:
@krancour thanks for the review!

>Who are these new make targets intended for? Is the expectation that those operating CF clusters will clone this repo and generate the tile themselves? Or should we be generating the tile and publishing it somewhere? My gut tells me it is the latter-- in which case, could you help us identify where it should be pushed to so we can build that into our CI/CD pipeline?

Yes, it is the latter. The tile of MASB is published here https://network.pivotal.io/products/azure-service-broker. Let's start to publish OSBA as a new product on Pivotal Network. @bingosummer can help to contact Pivotal and publish the tile.
And yes we can build that into CI/CD pipeline, the new release on Github include an extra .pivotal file.

krancour:
@zhongyi-zhang can you also make appropriate changes to .gitignore so that artifacts generated during this process won't ever be accidentally committed to source control?

zhongyi-zhang:
@krancour sorry I missed it... will do soon.
BTW, an extra file `tile-history.yml` which records the history version numbers will also be generated after the first generation. And the following generations will update it. That is to said, we need to commit it after each release generation. It is OK? Or, we can specify the version to generate and write a mocking `tile-history.yml` by script before the generation.

krancour:
> BTW, an extra file tile-history.yml which records the history version numbers will also be generated after the first generation. And the following generations will update it. That is to said, we need to commit it after each release generation. It is OK to also build into CI/CD? Or, we can specify the version to generate and write a mocking tile-history.yml by script before the generation.

@zhongyi-zhang something similar goes on with Helm charts when we publish those. (That's not in this repo; it's separate from here for now.) What we do is we retrieve an index.yaml from the chart repository (not the github repository) and we merge that with a newly generated index.yaml. This way history is preserved and new details are amended to it. Would there be any way of doing something similar with the tile?

If not, then you may have to work out how to incorporate this into CI/CD.

Is your opinion that this PR is useful / mergeable in its current state? Should we tackle the above in a follow-up PR?

zhongyi-zhang:
>What we do is we retrieve an index.yaml from the chart repository (not the github repository) and we merge that with a newly generated index.yaml.

@krancour is it manual? That's OK we can also manually update the tile history after a release. Release is not so frequent after all.

> Is your opinion that this PR is useful / mergeable in its current state? Should we tackle the above in a follow-up PR?

Yes, I think so.

zhongyi-zhang:
Need some fixes according to the env vars changes in https://github.com/Azure/open-service-broker-azure/pull/333 and https://github.com/Azure/open-service-broker-azure/pull/338.

krancour:
@zhongyi-zhang just letting you know we haven't forgotten about this. We'll be sure to get it into v0.10.0.

krancour:
I want to get this merged, but it's currently failing with the following error:

> ~/Code/go/src/github.com/Azure/open-service-broker-azure/contrib/cf/pcf-tile
docker run --rm -v /Users/kent/Code/go/src/github.com/Azure/open-service-broker-azure/contrib/cf/pcf-tile:/workspace -w /workspace --name=tile-generator cfplatformeng/tile-generator tile build major
  File "/usr/lib/python2.7/site.py", line 1
SyntaxError: Non-ASCII character '\xf5' in file /usr/lib/python2.7/site.py on line 1, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details
make: *** [generate-major-tile] Error 1

contrib/cf/pcf-tile/Makefile
```diff
@@ -0,0 +1,38 @@
...
+
+GENERATE_TILE = docker run --rm -v ${CURDIR}:/workspace -w /workspace --name=tile-generator cfplatformeng/tile-generator tile build
```
krancour: The image name `cfplatformeng/tile-generator` is implicitly interpreted as `cfplatformeng/tile-generator:latest`, where `latest` denotes a mutable tag. (i.e. Over time, this will point to different, newer images.) This impairs the determinism and repeatability of the tasks that use this.

Please update this to reference a specific version of the image, i.e. referenced using a semver tag.

krancour:
I am not certain, but https://github.com/Azure/open-service-broker-azure/pull/304/files#r178948722 might have been a factor in getting the error I showed above. It does seem that there has been a major release and two patch releases of that image since this PR was first opened.

contrib/cf/pcf-tile/Makefile
zhongyi-zhang: @krancour thanks! Fixed it with the same version as pull.

contrib/cf/pcf-tile/Makefile
```diff
@@ -0,0 +1,38 @@
...
+	pushd ../../..; \
+	  rm -f contrib/cf/pcf-tile/resources/open-service-broker-azure.zip; \
+	  zip -r contrib/cf/pcf-tile/resources/open-service-broker-azure.zip cmd pkg vendor; \
+	popd;
+
+.PHONY: prepare-to-generate-tile
+prepare-to-generate-tile: pull-tile-generator zip-broker
+
+.PHONY: generate-major-tile
+generate-major-tile: prepare-to-generate-tile
```
krancour: @zhongyi-zhang something I am struggling with here...

Since there are separate targets for generating major and minor versions of tiles, I need to know how we automate this. How can the CI/CD pipeline know which of these two targets to invoke?

We can could apply some regex to the version number, of course, but what I'm really trying to understand is why the process is _different_ for major vs minor. Should we even _need_ to decide between the two? How do the results of one differ from the other?

krancour: From [docs](https://docs.pivotal.io/tiledev/2-0/tile-generator.html):

> Tile Generator uses semver versioning. By default, tile build generates the next patch release. Major and minor releases can be generated by explicitly specifying tile build major or tile build minor. Or to override the version number completely, specify a valid semver version on the build command, e.g. tile build 3.4.5.

So maybe we should have a single make target for generating tiles and it should learn the version number from an environment variable set by the CD pipeline?

Here's what we do for OSBA Docker images, for instance during CD (publishing to Docker Hub):

```
# Push officially released, semantically versioned image
.PHONY: push-release
push-release:
ifndef REL_VERSION
	$(error REL_VERSION is undefined)
endif
	docker pull $(RC_IMAGE_NAME)
	docker tag $(RC_IMAGE_NAME) $(REL_IMAGE_NAME)
	docker tag $(RC_IMAGE_NAME) $(REL_MUTABLE_IMAGE_NAME)
	docker push $(REL_IMAGE_NAME)
	docker push $(REL_MUTABLE_IMAGE_NAME)
```

I imagine, perhaps, we can do something similar for the tile:

```
.PHONY: generate-tile
generate-tile: prepare-to-generate-tile
@# REL_VERSION will be set by the CD pipeline when publishing a tile (i.e. as part of the release process)
@# When REL_VERSION is not set, we'll generate without specifying a version number. This simply allows
@# this step to be tested during CI so that we know it will work in CD.
ifdef REL_VERSION
	${GENERATE_TILE} ${REL_VERSION}
else
	${GENERATE_TILE}
endif
```

Then specific targets `generate-major-tile` and `generate-minor-tile` go away.

Let me know if you think this works.

krancour:
@zhongyi-zhang I spent some time playing with this and amended the PR with a couple commits that I think will bring this over the finish line.

Highlights:

- Moved make targets into main Makefile (we have a preference to have just one)
- Eliminated targets like `generate-major-tile` and `generate-minor-tile` in favor of _just_ `generate-pcf-tile`, which takes further cues from the environment variable `REL_VERSION`.
- As with many of our other targets, this can be executed inside or outside of a container. "Outside" doesn't _really_ mean outside. But in many CI/CD environments, including Circle (and Jenkins, where we hope to move soon), the CI/CD system manages the containers _for us_. So "outside" really means "I'm already in a container; just run the command and don't worry about wrapping it in `docker run` for me."
- tile-history.yaml is ignored. It's no longer needed because we're not going to let the tile generation tool figure out next version numbers itself, rather we pass it `REL_VERSION` to explicitly specify the desired version number ourselves. If `REL_VERSION` isn't set (any build not happening as part of a formal release), the version of the tile generated will be 0.0.1. This is fine, because in any case where we do not intend to publish the generated tile, we're only just testing our ability to do so. `REL_VERSION` will always be set during a formal release, so the version info on the tile will always be correct before that artifact is published. (Publishing is not currently accounted for by this PR.)
- And obviously, this also fully integrates the tile generation process into CI and (to a lesser extend) CD (lesser because, again, we're not actually publishing tiles yet).

zhongyi-zhang:
@krancour cool! Thanks for these totally nice changes. LGTM.

krancour:
@zhongyi-zhang thanks for reviewing.

krancour:
Closed in error. Re-opened.

PR #308, opened 2018-02-27
misc cleanup of postgresql module

krancour:
This is just a bit of cleanup. There should be no functional changes in here. I'm going to annotate the changes with comments to explain the rationale behind each change.

I have similar PRs for MySQL and MS SQL almost ready as well.


cmd/broker/modules.go
```diff
@@ -26,7 +26,7 @@ import (
 	"github.com/Azure/open-service-broker-azure/pkg/services/eventhubs"
 	"github.com/Azure/open-service-broker-azure/pkg/services/keyvault"
 	"github.com/Azure/open-service-broker-azure/pkg/services/mysqldb"
-	"github.com/Azure/open-service-broker-azure/pkg/services/postgresqldb"
+	"github.com/Azure/open-service-broker-azure/pkg/services/postgresql"
```
krancour: This is the first place you'll spot a pervasive change. I changed the package name for the postgres module. Since the module has managers that handle both dbms and databases, it seemed that the name `postgresqldb` hinted too strongly as just handling databases. I wanted to fix that.

pkg/api/bind.go
```diff
@@ -104,54 +104,62 @@ func (s *server) bind(w http.ResponseWriter, r *http.Request) {
...
-	err = decoder.Decode(bindingRequest.Parameters)
-	if err != nil {
-		log.WithFields(logFields).Debug(
-			"bad binding request: error decoding parameter map",
-		)
-		// krancour: Choosing to interpret this scenario as a bad request since the
-		// probable cause would be disagreement between provided and expected types
-		s.writeResponse(w, http.StatusBadRequest, generateEmptyResponse())
-		return
+	if bindingParameters != nil {
```
krancour: Only change to this whole block of code is the nil guard at the beginning. Why? Because I don't want to force managers that don't use any bindingParameters to have to provide a useless type for that purpose just because.

pkg/api/bind.go
```diff
@@ -104,54 +104,62 @@ func (s *server) bind(w http.ResponseWriter, r *http.Request) {
...
-	err = decoder.Decode(bindingRequest.Parameters)
-	if err != nil {
-		log.WithFields(logFields).Debug(
-			"bad binding request: error decoding parameter map",
-		)
-		// krancour: Choosing to interpret this scenario as a bad request since the
-		// probable cause would be disagreement between provided and expected types
-		s.writeResponse(w, http.StatusBadRequest, generateEmptyResponse())
-		return
+	if secureBindingParameters != nil {
```
krancour: Same rationale as [here](https://github.com/Azure/open-service-broker-azure/pull/308/files#r170923216).

pkg/api/provision.go
```diff
@@ -194,6 +194,7 @@ func (s *server) provision(w http.ResponseWriter, r *http.Request) {
 		}
 		err = decoder.Decode(mapTagsIfaces)
 		if err != nil {
+			logFields["error"] = err
```
krancour: This should have been here all along.

pkg/api/provision.go
```diff
@@ -242,57 +243,63 @@ func (s *server) provision(w http.ResponseWriter, r *http.Request) {
...
-	if err != nil {
-		log.WithFields(logFields).Debug(
-			"bad provisioning request: error decoding parameter map into " +
-				"service-specific parameters",
-		)
-		// krancour: Choosing to interpret this scenario as a bad request since the
-		// probable cause would be disagreement between provided and expected types
-		s.writeResponse(w, http.StatusBadRequest, generateInvalidRequestResponse())
-		return
+	if provisioningParameters != nil {
```
krancour: Same rationale as [here](https://github.com/Azure/open-service-broker-azure/pull/308/files#r170923216).

pkg/api/provision.go
```diff
@@ -242,57 +243,63 @@ func (s *server) provision(w http.ResponseWriter, r *http.Request) {
...
-	if err != nil {
-		log.WithFields(logFields).Debug(
-			"bad provisioning request: error decoding parameter map into " +
-				"service-specific secured parameters",
-		)
-		// krancour: Choosing to interpret this scenario as a bad request since the
-		// probable cause would be disagreement between provided and expected types
-		s.writeResponse(w, http.StatusBadRequest, generateInvalidRequestResponse())
-		return
+	if secureProvisioningParameters != nil {
```
krancour: Same rationale as [here](https://github.com/Azure/open-service-broker-azure/pull/308/files#r170923216).

pkg/api/update.go
```diff
@@ -114,55 +114,61 @@ func (s *server) update(w http.ResponseWriter, r *http.Request) {
...
-	err = decoder.Decode(updatingRequest.Parameters)
-	if err != nil {
-		log.WithFields(logFields).Debug(
-			"bad updating request: error decoding parameter map",
-		)
-		// krancour: Choosing to interpret this scenario as a bad request since the
-		// probable cause would be disagreement between provided and expected types
-		s.writeResponse(w, http.StatusBadRequest, generateEmptyResponse())
-		return
+	if updatingParameters != nil {
```
krancour: Same rationale as [here](https://github.com/Azure/open-service-broker-azure/pull/308/files#r170923216).

pkg/api/update.go
```diff
@@ -114,55 +114,61 @@ func (s *server) update(w http.ResponseWriter, r *http.Request) {
...
-	err = decoder.Decode(updatingRequest.Parameters)
-	if err != nil {
-		log.WithFields(logFields).Debug(
-			"bad updating request: error decoding parameter map",
-		)
-		// krancour: Choosing to interpret this scenario as a bad request since the
-		// probable cause would be disagreement between provided and expected types
-		s.writeResponse(w, http.StatusBadRequest, generateEmptyResponse())
-		return
+	if secureUpdatingParameters != nil {
```
krancour: Same rationale as [here](https://github.com/Azure/open-service-broker-azure/pull/308/files#r170923216).

pkg/services/postgresql/all_in_one_arm_template.go
```diff
@@ -1,7 +1,7 @@
-package postgresqldb
+package postgresql
 
 // nolint: lll
-var armTemplateBytes = []byte(`
+var allInOneARMTemplateBytes = []byte(`
```
krancour: Just trying to make this name more descriptive.

pkg/services/postgresql/all_in_one_bind.go
```diff
@@ -20,18 +20,17 @@ func (a *allInOneManager) Bind(
 	_ service.BindingParameters,
 	_ service.SecureBindingParameters,
 ) (service.BindingDetails, service.SecureBindingDetails, error) {
-	dt, ok := instance.Details.(*allInOnePostgresqlInstanceDetails)
+	dt, ok := instance.Details.(*allInOneInstanceDetails)
```
krancour: Another pervasive change you will notice: I dropped module/service names from types. The package name already provides this context. It's borderline obnoxious that it was ever done like this to begin with, but it's my own fault, because that is a mistake I made early on in broker development and it's been replicated in other modules as well.

pkg/services/postgresql/all_in_one_bind.go
```diff
@@ -20,18 +20,17 @@ func (a *allInOneManager) Bind(
...
 	if !ok {
 		return nil, nil, fmt.Errorf(
-			"error casting instance.Details " +
-				"as *allInOnePostgresqlInstanceDetails",
+			"error casting instance.Details as *postgresql.allInOneInstanceDetails",
 		)
 	}
-	sdt, ok := instance.SecureDetails.(*allInOnePostgresqlSecureInstanceDetails)
+	sdt, ok := instance.SecureDetails.(*secureAllInOneInstanceDetails)
```
krancour: Another pervasive change: the detail "secure" moves out in front of the type name. It reads better.

pkg/services/postgresql/all_in_one_provision.go
```diff
@@ -22,10 +22,7 @@ func (a *allInOneManager) ValidateProvisioningParameters(
 				"*postgresql.AllInOneProvisioningParameters",
 		)
 	}
-	return validateServerParameters(
-		pp.SSLEnforcement,
-		pp.FirewallRules,
-	)
+	return validateDBMSProvisionParameters(&pp.DBMSProvisioningParameters)
```
krancour: This just felt a little cleaner. If the list of params grows, we won't have to touch this code again.

pkg/services/postgresql/all_in_one_provision.go
```diff
@@ -146,13 +141,13 @@ func (a *allInOneManager) deployARMTemplate(
 		sdt,
 	)
 	goTemplateParameters := buildGoTemplateParameters(
-		&pp.ServerProvisioningParameters,
+		&pp.DBMSProvisioningParameters,
```
krancour: Another pervasive change: I've opted for "DBMS" over "Server" wherever it makes sense. The thing we say we'll provision is a DBMS, after all. Where this _doesn't_ change is anywhere where we _are_ specifically talking about a "server." A good example is that even when deprovisioning a DBMS, we use the Azure SDK for Go's `postgresql.ServerClient` (or something like that) to `DeleteServer(...)`. So the deprovisiong step that executes that remains "deprovisionServer".

pkg/services/postgresql/all_in_one_types.go
```diff
@@ -31,41 +27,41 @@ func (
 func (
 	a *allInOneManager,
 ) GetEmptySecureProvisioningParameters() service.SecureProvisioningParameters {
-	return &SecureAllInOneProvisioningParameters{}
+	return nil
```
krancour: Another pervasive change: these sort of functions can now return `nil` if they'd otherwise be returning a useless struct type that has no fields.

pkg/services/postgresql/catalog.go
```diff
@@ -105,33 +102,31 @@ func (m *module) GetCatalog() (service.Catalog, error) {
 				},
 			}),
 		),
-		//db only
+		// database only
```
krancour: Pervasive: I'm opting to use "database" in lieu of "db" wherever I can because this is how I want users or contributors to talk-- because "db" is too easily confused with "dbms" when written or read, spoken or heard.

pkg/services/postgresql/database_bind.go
```diff
@@ -0,0 +1,95 @@
+package postgresql
```
krancour: Something else you may notice-- I renamed many files so that their names denote manager first and function second. Visually, this keeps things that are more related grouped together in the tree.

pkg/services/postgresql/database_provision.go
```diff
@@ -81,30 +79,29 @@ func (d *dbOnlyManager) deployARMTemplate(
 	return dt, instance.SecureDetails, nil
 }
 
-func (d *dbOnlyManager) setupDatabase(
+func (d *databaseManager) setupDatabase(
```
krancour: I've dropped "only" wherever I could because I think the role of these managers is clear without "only" in their name. I _did_ keep "only" in relevant catalog descriptions, since that is seen by the end user.

pkg/services/postgresql/all_in_one_types.go
```diff
@@ -1,25 +1,21 @@
...
+package postgresql
 
 import "github.com/Azure/open-service-broker-azure/pkg/service"
 
 // AllInOneProvisioningParameters encapsulates non-sensitive dbms AND database
 // PostgreSQL-specific provisioning options
 type AllInOneProvisioningParameters struct {
-	ServerProvisioningParameters
-	DatabaseProvisioningParameters
+	DBMSProvisioningParameters     `json:",squash"`
```
krancour: The `mapStructure` library that we use gets confused if this annotation isn't present wherever we use composition like this.

pkg/services/postgresql/all_in_one_provision.go
jeremyrickard: 👍

pkg/services/postgresql/all_in_one_types.go
jeremyrickard: :+1:

pkg/services/postgresql/database_bind.go
jeremyrickard: 👍

PR #310, opened 2018-02-27
misc cleanup of mysql module

krancour:
@jeremyrickard similar cleanup to what was in #308


krancour:
Thanks for the review @jeremyrickard.

PR #311, opened 2018-02-27
cli: short-circuit param parsing if there are none

krancour:
Currently, provision and bind fail if you _don't_ provide any params as JSON. (So at minimum, you must currently provide `'{}'`.) This fixes that.

No CI needed on this since tests don't use the CLI.


krancour:
Gonna hold off on merging this until #310 is in just to avoid the appearance of #310 requiring a rebase after it's tests have passed.

PR #312, opened 2018-02-27
Updates to the SQLDB, MySQL and PostgreSQL Docs.

### LINKED ISSUE #271: Module documentation needs to be updated
The existing module documentation needs to be updated. Some current issues:

* Firewall rules were changed from a default of allow all to default of only Azure IPs. Previously there was no documentation about Firewall rules in any of the modules. The default behavior, if no firewall is specified and the broker is run outside of Azure, prevents binding from occuring
* MySQL, PostgreSQL and SQLDB now support dependent services (i.e. provision dbms then provision database). Documentation does not reflect this

Need to do a complete examination of current module docs.

jeremyrickard:
Closes #271

For MySQL, PostgreSQL and SQLDB, the docs have been updated to reflect the fact that
we now provide three services for each module and support parent/child provisioning between
the DBMS and Database services.

For PostgreSQl, the extensions param was also added. It was previously missing.


docs/modules/mssql.md
```diff
@@ -0,0 +1,180 @@
...
+###### Provisioning Parameters
+
+| Parameter Name | Type | Description | Required | Default Value |
+|----------------|------|-------------|----------|---------------|
+| `location` | `string` | The Azure region in which to provision applicable resources. | Required _unless_ an administrator has configured the broker itself with a default location. | The broker's default location, if configured. |
+| `resourceGroup` | `string` | The (new or existing) resource group with which to associate new resources. | N | If an administrator has configured the broker itself with a default resource group and nonde is specified, that default will be applied, otherwise, a new resource group will be created with a UUID as its name. |
+| `tags` | `map[string]string` | Tags to be applied to new resources, specified as key/value pairs. | N | Tags (even if none are specified) are automatically supplemented with `heritage: open-service-broker-azure`. |
+| `alias` | `string` | Y | Specifies an alias that can be used by later provision actions to create databases on this DBMS. |
```
krancour: Looks like description and required are accidentally reversed on this row.

krancour: The same is true for MySQL.

docs/modules/postgresql.md
```diff
@@ -0,0 +1,170 @@
+# [Azure Database for PostgreSQL](https://azure.microsoft.com/en-us/services/postgresql/)
+
+|![](https://upload.wikimedia.org/wikipedia/commons/thumb/1/17/Warning.svg/50px-Warning.svg.png) | This module is EXPERIMENTAL. It is under heavy development and remains subject to the possibility of breaking changes. |
+|---|---|
+
+Open Service Broker for Azure contains three Azure Database for PostgreSQL services. These services enable you to select the most appropriate provision scenario for your needs. These services are:
+
+| Service Name | Description ||--------------|-------------|
```
krancour: This isn't rendering correctly. Need a line break here.

docs/modules/mssql.md
krancour: And PostgreSQL.

jeremyrickard: Because I copy-pasta'ed them. Fixing!

docs/modules/postgresql.md
jeremyrickard: Crazy, don't see that in code. will fix.

krancour:
The only thing I think is missing is the `tags` field on credentials. That's something you added recently, right?

jeremyrickard:
@krancour  re: tags (yes) and the URI! I added that and forgot to push the commit up. Doing that now

Issue #314, opened 2018-02-28
Add example requests to the MySQL, PostgreSQL and SQL DB modules


jeremyrickard:
Add some sample requests to the various modules to enhance the documentation a little bit.

PR #316, opened 2018-02-28
Lots of changes to the module documentation and the contrib examples

jeremyrickard:
Changes to contrib manifests to reflect plan name changes, also some directory structure updates to make things consistent. 

Changes to the module documentation to provide examples and fix a couple of missed updates. 

Closes: #314


docs/modules/mssql.md
```diff
@@ -92,7 +154,8 @@ Deletes both the database and the SQL Server instance.
 
 ##### Provision
 
-Provisions a new SQL Server DBMS only. Databases can be created through subsequent provision requests using the `azure-sql-db-only` service.
+Provisions a new SQL Server DBMS only. Databases can be created through subsequent provision requests using the `azure-sql-database-only` service.
```
itowlson: Suggestion: consider phrasing such as "an empty SQL Server DBMS".  It may be less technically accurate, but I had to think through what "DBMS only" meant in this context.  Or "a SQL Server DBMS instance containing no databases" perhaps?

jeremyrickard: i like it!

contrib/doc-templates/module.md
```diff
@@ -73,3 +73,17 @@ _Describe what occurs upon unbinding. (i.e. Is a user account deleted?)_
...
+
+##### Examples
+
+###### Kubernetes
+
+_Provide an example with manifests. Manifests should be added to contrib/k8s/examples/{module}_
+
+###### Cloud Foundry
+
+_Provide an example with using cf cli_
```
krancour: "with using" sounds odd.

contrib/k8s/examples/mysql/mysqldb-database-only-binding.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: servicecatalog.k8s.io/v1beta1
+kind: ServiceBinding
+metadata:
+  name: mysql-database-only-binding
```
krancour: Since we can only bind to databases anyway (can't bind to dbms), do we really need to qualify this with "database-only"?

In fact... as I think about it more...

Can't we have one binding example that works with both the db that's provisioned as part of all-in-one _and_ a db that is provisioned as a child of a dbms?

contrib/k8s/examples/mysql/mysqldb-database-only-instance..yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: servicecatalog.k8s.io/v1beta1
```
krancour: There's an extra `.` in this filename.

contrib/k8s/examples/mysql/mysqldb-dbms-only-instance..yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: servicecatalog.k8s.io/v1beta1
```
krancour: There's an extra . in this filename.

Also... any reason we can't put the dbms manifest and db-only manifest in the same file? May be more convenient for the user and it's a good way to show off that our broker is smart enough to sort the dependency out.

contrib/k8s/examples/mysql/mysqldb-database-only-binding.yaml
jeremyrickard: We can only do that if the instance name is the same. Should we make the instance name the same and assume they cannot be provisioned at the same time? I'm fine either way (this applies across all three)

contrib/k8s/examples/mysql/mysqldb-dbms-only-instance..yaml
jeremyrickard: Yep, we can do that. Applies across the three modules.

contrib/k8s/examples/mysql/mysqldb-database-only-binding.yaml
krancour: > We can only do that if the instance name is the same.

I thought I was ok with that, but then...

> they cannot be provisioned at the same time?

That will screw over _someone_ who's just kicking the tires.

Maybe we can can just drop "only" from the binding name?

krancour: I would also suggest prefixing k8s resource names with something like "my-" so we'e making it more clear that there's no specific expectation that resource (instance and binding) names _match_ service names.

jeremyrickard: done!

docs/modules/mssql.md
```diff
@@ -118,7 +181,70 @@ This service is not bindable.
...
+                "endIPAddress" : "255.255.255.255"
+            }
+        ]
+    }
+'
+```
+
+###### cURL
+
+To provision an instance using the broker directly, you must use the ID for both plan and service. Assuming your OSBA is running locally on port 8080 with the default username and password, you can provision the DBMS only Azure SQL Database service with a cUrl command similar to the following example. This example illustrates multiple firewall rules and provides an alias for later database provisioning:
```
mmacy: * DBMS-only
* cURL

docs/modules/mssql.md
```diff
@@ -177,4 +303,52 @@ Drops the applicable user from the SQL Server.
 
 ##### Deprovision
 
-Deletes both the database from the SQL Server instance, but does not delete the DBMS.
\ No newline at end of file
+Deletes both the database from the SQL Server instance, but does not delete the DBMS.
```
mmacy: Should this be "both **of** the database**s**"?

docs/modules/mssql.md
```diff
@@ -177,4 +303,52 @@ Drops the applicable user from the SQL Server.
...
+cf create-service azure-sql-database-only basic azure-sql-db-only -c '{
+    "parentAlias" : "ed9798f2-2e91-4b21-8903-d364a3ff7d12"
+}'
+```
+
+Note: this uses the alias used when provisioning the DBMS only service above.
+
+###### cURL
+
+To provision an instance using the broker directly, you must use the ID for both plan and service. Assuming your OSBA is running locally on port 8080 with the default username and password, you can provision the database only Azure SQL Database service onto a previously provisioned SQL DB DBMS with a cUrl command similar to the following example. Note, this uses the alias provided in the DBMS only example above:
```
mmacy: * database-only
* cURL
* DMBS-only

docs/modules/mysql.md
```diff
@@ -9,9 +9,9 @@ Open Service Broker for Azure contains three Azure Database for MySQL services.
...
-| `azure-mysql-db-only` | Provision a new database only upon a previously provisioned DBMS. |
+| `azure-mysql-database-only` | Provision a new database only upon a previously provisioned DBMS. |
 
-The `azure-mysql` service allows you to provision both a DBMS and a database. This service is ready to use upon successful provisioning. You can not provision additional databases onto an instance provisioned through this service. The `azure-mysql-dbms-only` and `azure-mysql-db-only` services, on the other hand, can be combined to prprovisionovison multiple databases on a single DBMS.  For more information on each service, refer to the descriptions below.
+The `azure-mysql` service allows you to provision both a DBMS and a database. This service is ready to use upon successful provisioning. You can not provision additional databases onto an instance provisioned through this service. The `azure-mysql-dbms-only` and `azure-mysql-database-only` services, on the other hand, can be combined to provision multiple databases on a single DBMS.  For more information on each service, refer to the descriptions below.
```
mmacy: "can not" > can't (or cannot)

docs/modules/mysql.md
```diff
@@ -160,4 +290,52 @@ Drops the applicable user from the MySQL DBMS.
...
+Using the `cf` cli, you can provision the `database-only` plan of this service with the following command:
+
+```console
+cf create-service azure-mysql-database-only database-only mysql-database-only -c '{
+    "parentAlias" : "679aab6d-39e7-4a41-8b45-49975569079c"
+}
+'
+```
+
+Note: this uses the alias provided in the DBMS only example
```
mmacy: DBMS-only

Also, the note above was terminated with a period, but here is not. Suggest standardizing.

docs/modules/mssql.md
```diff
@@ -3,15 +3,15 @@
 |![](https://upload.wikimedia.org/wikipedia/commons/thumb/1/17/Warning.svg/50px-Warning.svg.png) | This module is EXPERIMENTAL. It is under heavy development and remains subject to the possibility of breaking changes. |
 |---|---|
 
-Open Service Broker for Azure contains three Azure SQL Database services. These services enable you to select the most appropriate provision scenario for your needs. These services are:
+Open Service Broker for Azure (OSBA) contains three Azure SQL Database services. These services enable you to select the most appropriate provision scenario for your needs. These services are:
```
arschles: s/provision/provisioning

docs/modules/mssql.md
```diff
@@ -3,15 +3,15 @@
...
-| `azure-sql-db-only` | Provision a new database only upon a previously provisioned DBMS. |
+| `azure-sql-database-only` | Provision a new database only upon a previously provisioned DBMS. |
 
-The `azure-sql` service allows you to provision both a DBMS and a database. This service is ready to use upon successful provisioning. You can not provision additional databases onto an instance provisioned through this service. The `azure-sql-dbms-only` and `azure-sql-db-only` services, on the other hand, can be combined to provision multiple databases on a single DBMS.  For more information on each service, refer to the descriptions below.
+The `azure-sql` service allows you to provision both a DBMS and a database. This service is ready to use upon successful provisioning. You can not provision additional databases onto an instance provisioned through this service. The `azure-sql-dbms-only` and `azure-sql-database-only` services, on the other hand, can be combined to provision multiple databases on a single DBMS.  For more information on each service, refer to the descriptions below.
```
arschles: s/`This service is ready to use upon successful provisioning`/`When the provision operation is successful, the database will be ready to use`

docs/modules/mysql.md
arschles: s/This service is ready to use upon successful provisioning/When the provision operation is successful, the database will be ready to use

docs/modules/postgresql.md
```diff
@@ -11,7 +11,7 @@ Open Service Broker for Azure contains three Azure Database for PostgreSQL servi
...
 | `azure-postgresql-database-only` | Provision a new database only upon a previously provisioned DBMS. |
 
-The `azure-postgresql` service allows you to provision both a DBMS and a database. This service is ready to use upon successful provisioning. You can not provision additional databases onto an instance provisioned through this service. The `azure-postgresql-dbms-only` and `azure-postgresql-db-only` services, on the other hand, can be combined to provision multiple databases on a single DBMS.  For more information on each service, refer to the descriptions below.
+The `azure-postgresql` service allows you to provision both a DBMS and a database. This service is ready to use upon successful provisioning. You can not provision additional databases onto an instance provisioned through this service. The `azure-postgresql-dbms-only` and `azure-postgresql-database-only` services, on the other hand, can be combined to provision multiple databases on a single DBMS.  For more information on each service, refer to the descriptions below.
```
arschles: s/This service is ready to use upon successful provisioning/When the provision operation is successful, the database will be ready to use

contrib/k8s/examples/mysql/mysqldb-dbms-only-instance.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: servicecatalog.k8s.io/v1beta1
+kind: ServiceInstance
+metadata:
+  name: my-my-mysqldb-instance
```
krancour: Extra "my" in here. "Oh, my!"

contrib/k8s/examples/mysql/mysqldb-all-in-one-binding.yaml
```diff
@@ -1,9 +1,9 @@
 apiVersion: servicecatalog.k8s.io/v1beta1
 kind: ServiceBinding
 metadata:
-  name: my-mysqldb-binding
+  name: my-mysql-binding
```
krancour: I want to suggest that across the board, we make the names of the manifest files match the names of the resources contained therein.

jeremyrickard: Argh yeah, forgot to rename the files when i moved them and dropped the db to match. Fixed that.

contrib/k8s/examples/mysql/mysqldb-dbms-only-instance.yaml
jeremyrickard: 😭

docs/modules/mssql.md
jeremyrickard: Should be deletes *ONLY* the database

contrib/k8s/examples/mysql/mysql-dbms-only-instance.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: servicecatalog.k8s.io/v1beta1
+kind: ServiceInstance
+metadata:
+  name: my-mysqldb-instance
```
carolynvs: How come sometimes we are using `my-mysql-database-*` and sometimes `my-mysqldb-*`? If we are trying to make it clear that one is a server and another is a db instance maybe use `mysql-server-*` and `mysql-database-*` instead? Or maybe `dbms` since that's another term you are using elsewhere.

contrib/k8s/examples/sqldb/sqldb-db-only-instance.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: db-only-sqldb-instance
   namespace: default
 spec:
-  clusterServiceClassExternalName: azure-sqldb-db-only
+  clusterServiceClassExternalName: azure-sql-database-only
   clusterServicePlanExternalName: basic
   parameters:
-    parentAlias: dbms-only-sqldb
+    parentAlias: 3e8f59b2-0f73-4a59-9c0b-6277c3e8109e
```
carolynvs: How come this switched from a name to a uuid?

docs/modules/mssql.md
```diff
@@ -82,6 +82,68 @@ Drops the applicable user from the SQL Server.
...
 Deletes both the database and the SQL Server instance.
 
+##### Examples
+
+###### Kubernetes
+
+The `contrib/k8s/examples/sqldb/sqldb-all-in-one-instance.yaml` can be used to provision one of the plans from the all-in-one `azure-sql` service. This can be done with the following example:
+
+```console
+kubectl create -f ../../contrib/k8s/examples/sqldb/sqldb-all-in-one-instance.yaml
```
carolynvs: Would it be possible here (and anywhere else in this doc) to use a URL? It's always annoying that one has to clone a repo just to run some of these commands.

docs/modules/mssql.md
```diff
@@ -82,6 +82,68 @@ Drops the applicable user from the SQL Server.
...
+    }
+'
+```
+
+###### cURL
+
+Assuming your OSBA is running locally on port 8080 with the default username and password, you can provision the all-in-one Azure SQL Database service with a cURL command similar to the following example:
+
+```console
+curl -X PUT \
```
carolynvs: You may want to make it more clear which values a user will need to replace, by setting them as env vars above the curl command, and then referencing them in this command.

contrib/k8s/examples/mysql/mysql-dbms-only-instance.yaml
jeremyrickard: Yeah, that was an omission! Thanks for catching it, I've updated it to my-mysql-dbms-instance to follow the pattern.

contrib/k8s/examples/sqldb/sqldb-db-only-instance.yaml
jeremyrickard: Suggestion from @krancour . We aren't doing validation on the parentAlias being unique, so he suggested changing it to a uuid. No strong feelings either way for me. 

@krancour ?

contrib/k8s/examples/mysql/mysql-dbms-only-instance.yaml
carolynvs: 🎵 _Carolyn just found a bug! 💃  Carolyn just found a bug! 🕺_ 🎵

docs/modules/mssql.md
jeremyrickard: Yeah, I think we could do that.

contrib/k8s/examples/mysql/mysql-dbms-only-instance.yaml
krancour: Ya-- this looks off.

Opinion incoming...

We should name example _resources_ like follows:

1. Prefix all resources with `my-` or `example-` or _something_ to make the name feel a little more "customized." We don't want to fool anyone (mostly people who do not know k8s as well as we do) into thinking that resource names have to match service or plan names. (I was favoring `my-`, before, but now feel `example-` might be more appropriate.
2. Manifest file names should match _resource_ names; not service names.
3. If the instance provisioned by a given manifest is a DBMS, it should say so in the resource name and file name. Likewise, if the thing provisioned is a database, that should be denoted clearly as well. (I think we should avoid "db" because in both writing and speech it's too easily to mis-say/mis-hear or mis-write/mis-read "db" when "dbms" might have been intended.
4. If the instance being provisioned is the "all-in-one" DBMS + single database, I think rather than confuse people with "all-in-one," we should just refer to such instances as "example-mysql-instance"-- not denoting DBMS _or_ database. Since this is unspecific, let's treat this like the default case. Specifying that you want a DBMS or database is advanced usage.
5. Make resource and file names for bindings include the name of the instance they bind to. `example-mysql-database-binding` binds to `example-mysql-instance`.

Bottom line is that we just need naming to be crystal clear. Feel free to push back on these recommendations if you think I am nuts.

krancour: Maybe also if we treat "all-in-one" as the base case and everything else as "advanced usage," perhaps for each set of examples, we can move the "advanced" manifests off into an `avanced/` subdir.

docs/modules/mssql.md
jeremyrickard: I will take care of this, along with https://github.com/Azure/open-service-broker-azure/issues/320 in a follow up PR.

contrib/k8s/examples/mysql/advanced/mysql-dbms-instance.yaml
```diff
@@ -1,15 +1,15 @@
 apiVersion: servicecatalog.k8s.io/v1beta1
 kind: ServiceInstance
 metadata:
-  name: dbms-only-sqldb-instance
+  name: my-mysql-dbms-instance
```
carolynvs: Did we want to standardize on either my-* or example-*?

contrib/k8s/examples/postgresql/advanced/postgresql-database-instance.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: servicecatalog.k8s.io/v1beta1
+kind: ServiceInstance
+metadata:
+  name: my-postgresql-database-instance
```
carolynvs: Within the same example, we should stick with the same naming convention, either my- or example-.

contrib/k8s/examples/sqldb/sqldb-db-only-instance.yaml
carolynvs: Within a namespace isn't it unique? Or does alias not refer to the k8s name?

jeremyrickard: It's within the broker, which doesn't do anything w/ k8s related info at the current time, so it's really at the broker level. We could make these names, jsut need to make sure it's unique amongst our examples in case anyone wants to provision multiple examples concurrently.

contrib/k8s/examples/mysql/advanced/mysql-dbms-instance.yaml
jeremyrickard: crap i thought I changed that. hold on.

jeremyrickard: it was changed in code but for some reason I hadn't saved it!

contrib/k8s/examples/sqldb/sqldb-db-only-instance.yaml
carolynvs: Oops! Sorry I live in k8s land all day so I forget sometimes that there is more to life than just k8s. 😀 No, I understand now. It makes more sense to leave as-is.

contrib/k8s/examples/mysql/README.md
```diff
@@ -0,0 +1,27 @@
...
+| `azure-mysql-dbms-only` | DBMS-only |
+| `azure-mysql-database-only` | Database-only |
+
+This directory contains example Kubernetes manifests to exercise these services.
+
+## Basic Usage
+
+The easiest way to use the Azure Database for MySQL service is to use the `azure-mysql` service to provision both a DBMS and a new database.
+
+The `mysql-instance.yaml` manifest will create an instance using the `azure-mysql` service using the `basic50` plan. This will result in a new Azure Database for MySQL that includes both the DBMS and the database. The `mysql-binding.yaml` will create a binding for the to this new database, ultimately resulting in a new Kubernetes secret named `example-mysql-secret`. Once created, you can use this secret in an application to connect to the new Azure Database for MySQL instance.
```
krancour: Replace this:
> will create a binding for the to this new database
With:
> manifest will create a binding to this new database

jeremyrickard:
@arschles changes are applied, take a look and see if it LGTY now.

contrib/k8s/examples/mysql/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Basic Usage
+
+The easiest way to use the Azure Database for MySQL service is to use the `azure-mysql` service to provision both a DBMS and a new database.
+
+The `mysql-instance.yaml` manifest will create an instance using the `azure-mysql` service using the `basic50` plan. This will result in a new Azure Database for MySQL that includes both the DBMS and the database. The `mysql-binding.yaml` manifest will create a binding for the to this new database, ultimately resulting in a new Kubernetes secret named `example-mysql-secret`. Once created, you can use this secret in an application to connect to the new Azure Database for MySQL instance.
+
+## Advanced Usage
+
+The `advanced` directory contains manifests that you can use to provision the `azure-mysql-dbms` and `azure-mysql-databasey` services. These services allow you to independently provision the Azure SQL Database DBMS and the database itself for more advanced use cases, such as running multiple databases on a single DBMS.
```
arschles: a/`azure-mysql-databasey`/`azure-mysql-database`

krancour:
I stopped short of triyng the curl commands, but this LGTM. Nay-- it looks _awesome_ to me. This raises the bar for quality docs, IMO.

jeremyrickard:
DOCS ONLY, NO CI NEEDED!!!!

PR #318, opened 2018-02-28
Quickstart updates

jeremyrickard:
Updates to the quickstart to include the use of the etcd storage persistence flag.

### DIFF
--- a/docs/quickstart-aks.md
+++ b/docs/quickstart-aks.md
@@ -190,7 +190,9 @@ You should also ensure that the `Microsoft.Compute` and `Microsoft.Network` prov
 1. Deploy Service Catalog on the cluster:
     ```console
     helm repo add svc-cat https://svc-catalog-charts.storage.googleapis.com
-    helm install svc-cat/catalog --name catalog --namespace catalog --set rbacEnable=false
+    helm install svc-cat/catalog --name catalog --namespace catalog \
+       --set rbacEnable=false \
+       --set storage.etcd.persistence.enabled=true
     ```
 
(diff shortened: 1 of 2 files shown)


docs/quickstart-aks.md
```diff
@@ -190,10 +190,12 @@ You should also ensure that the `Microsoft.Compute` and `Microsoft.Network` prov
...
     ```console
     helm repo add svc-cat https://svc-catalog-charts.storage.googleapis.com
-    helm install svc-cat/catalog --name catalog --namespace catalog --set rbacEnable=false
+    helm install svc-cat/catalog --name catalog --namespace catalog \
+       --set rbacEnable=false \
+       --set apiserver.storage.etcd.persistence.enabled=true
     ```
 
-    Note: the AKS preview does not _currently_ support RBAC, so you must disable RBAC as shown above.
+    Note: the AKS preview does not _currently_ support RBAC, so you must disable RBAC as shown above. The command above also enables persistence for the embedded etcd used by Service Catalog. This is desirable for evaluation in the event that parts of Service Catalog are restarted. For production use, we recommend a dedicated etcd cluster with appropriate storage.
```
carolynvs: To help out those are who aren't intimately familiar with what that flag does, maybe we can be a bit more clear beyond "enables persistence"? Does it save the svc-cat etcd data on a volume? Does it allow you to reinstall/upgrade svc-cat without losing your data?

jeremyrickard: That's a good idea! Will go ahead and do that now.

PR #319, opened 2018-03-01
bump all rdbms modules to preview level of stability

krancour:
This is the _last_ thing I want to go in before we cut v0.9.0.


arschles:
looks good to me, not that I'm biased or anything :)

Issue #325, opened 2018-03-06
MS SQL: enable optional TDE


krancour:
This issue is split from the broader #3 because we intend to address this in v0.10.0, while other (possibly as-yet-unidentified) features may be implemented later, as they are identified (and if there's value in doing so).

jeremyrickard:
TDE is turned on be default:

```
All newly created SQL databases are encrypted by default by using service-managed transparent data encryption. Existing databases before May 2017 and databases created through restore, geo-replication, and database copy aren't encrypted by default.
```

per https://docs.microsoft.com/en-us/sql/relational-databases/security/encryption/transparent-data-encryption-azure-sql so this is being added with a default value of true for backward compatibility.

jeremyrickard:
@ritazh @zhongyi-zhang  I wanted to follow up on this issue.

TDE is enabled by default on Azure SQL databases. Do we have a really compelling use case for actually disabling it? I'm having issues with ARM disabling it at provision time:

zhongyi-zhang:
@jeremyrickard Indeed, I never heard about a use case for actually disabling it.
Agree not to expose the parameter and just enable it. It would be not too late to expose it if someone shared a reasonable scenario to disable TDE.

jeremyrickard:
Closing this. TDE is enabled by default, if people want to disable it we recommend they go to the portal and do it out of band

PR #329, opened 2018-03-07
Update MS SQL module to allow TDE to be disabled

jeremyrickard:
Adds a new parameter called "disableTransparentDataEncryption" to the all-in-one
and database only MS SQL services. TDE is enabled by default, so this was added in a
backward compatible way. TDE is enabled by default, this allows it to be disabled.

Closes: #325


jeremyrickard:
I added Do Not Merge to this because it is occasionally failing. I've reached out to the the Azure SQL team and @arschles gave me an ARM team contact, so will leave this here until we can figure out if there is a more correct way to disable TDE on a new provision.

jeremyrickard:
Closing this as we don't want to implement this.

Issue #237, opened 2018-02-06
Make number of async worker goroutines configurable


krancour:
The number of goroutines handling pending tasks and deferred tasks should be configurable (independently of one another).

Goroutines are very lightweight. Chances are high we can afford to run _many, many, many_ more than we are now. I'll open a separate PR for profiling resource usage, but for now, just having some knobs to turn is a good start.

PR #330, opened 2018-03-08
move configs around

krancour:
Related to #237 

When I address #237, there will be additional configuration options for the async engine. Keeping in mind that we will very likely extract that package into its own project at some point in the future, I want to make sure those options are defined _within_ the async package- which means the existing async options should probably move to that package as well...

Well... to be consistent, that means options for a lot of other packages should move closer to the components they affect. This make a lot of sense, actually, because as with async, this reduces the effort required to extract broker packages into their own projects (if warranted) in the future.

This is _not_ the last of the refactoring that's going to happen on configs/options, but I want to keep PRs tightly scoped and easier-to-review.

This is ready for review, but let's hold off on merging it for now. I want to let this marinate for a while. I'm building on this work in another PR, and I want to feel more certain that this all works out as I hope before we merge.

### DIFF
--- a/cmd/broker/broker.go
+++ b/cmd/broker/broker.go
@@ -20,6 +20,7 @@ import (
 	"github.com/Azure/open-service-broker-azure/pkg/crypto/noop"
 	"github.com/Azure/open-service-broker-azure/pkg/http/filter"
 	"github.com/Azure/open-service-broker-azure/pkg/http/filters"
+	brokerLog "github.com/Azure/open-service-broker-azure/pkg/log"
 	"github.com/Azure/open-service-broker-azure/pkg/service"
 	"github.com/Azure/open-service-broker-azure/pkg/version"
 	log "github.com/Sirupsen/logrus"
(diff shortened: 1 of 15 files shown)


cmd/broker/modules.go
```diff
@@ -37,11 +36,11 @@ import (
 
 var modules []service.Module
 
-func initModules(azureConfig config.AzureConfig) error {
+func initModules(azureConfig azure.Config) error {
```
jeremyrickard: i like how this all reads now!

cmd/broker/broker.go
```diff
@@ -88,6 +88,12 @@ func main() {
 		DB:         storageRedisConfig.GetDB(),
 		MaxRetries: 5,
 	}
+	if storageRedisConfig.IsTLSEnabled() {
```
jeremyrickard: sorry I missed this when reviewing

krancour: Sorry I missed it when coding!

PR #331, opened 2018-03-09
Adding a preliminary design doc

arschles:
I want to keep the design doc short and succinct, and write up details in supporting documents in follow-ups

cc/ @krancour @jeremyrickard


docs/design.md
```diff
@@ -0,0 +1,59 @@
...
+(Note: testing infrastructure is out of scope for this document)
+
+# REST API Frontend
+
+OSBA runs an HTTP server that provides a REST API that conforms to the
+[Open Service Broker API specification](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md)
+
+TODOs for this section:
+
+* Specific version of the spec?
```
arschles: @krancour @jeremyrickard are you targeting a specific version/range in the spec?

docs/design.md
```diff
@@ -0,0 +1,59 @@
...
+* More detail on the server & its handlers
```
arschles: @jeremyrickard @krancour anything specific you'd like me to put here? Otherwise, I'm gonna do a 2 sentence overview of the routing framework & how handlers are invoked..

docs/design.md
jeremyrickard: We haven't tied to a specific version yet, just 2.x.

docs/design.md
```diff
@@ -0,0 +1,59 @@
+# OSBA Design
+
+We can roughly split OSBA's architecture into three pieces:
+
+1.  REST API frontend
+2.  Stable Storage
+3.  Asynchronous processing
```
jeremyrickard: I think maybe call out the modules here? This design doc is probably for people that will eventually contribute modules so i think it would be good to add a section detailing what the module does vs what the front end does.

jeremyrickard:
no CI needed.

docs/design.md
arschles: Sounds good! I'll add a fourth item here explaining how they interact with the above 3

PR #332, opened 2018-03-09
CosmosDB Refactor

### LINKED ISSUE #143: Refactor cosmosdb module to support multiple service managers
The current cosmosdb module uses a shared service manager between the documentdb and mongodb services. We should refactor this to use different service managers.

jeremyrickard:
Refactor CosmosDB to match the changes made to Sql DB, MySQL etc. 

Removed the DocumentDB wording to match the updated naming found in CosmosDB documentation.  

Closes #143


pkg/services/cosmosdb/catalog.go
```diff
@@ -9,13 +9,13 @@ func (m *module) GetCatalog() (service.Catalog, error) {
...
 				&service.ServiceProperties{
 					ID:   "6330de6f-a561-43ea-a15e-b99f44d183e6",
-					Name: "azure-cosmos-document-db",
-					Description: "Azure DocumentDB (Experimental) provided by CosmosDB " +
-						"and accessible via SQL (DocumentDB), Gremlin (Graph), and Table " +
-						"(Key-Value) APIs",
+					Name: "azure-cosmos-db",
+					Description: "Azure Cosmos DB is a globally distributed, " +
+						"multi-model database service accessible via SQL, " +
+						"Gremlin (Graph), and Table (Key-Value) APIs",
```
krancour: Should we change the name and description to be clear that what this provisions is an _account_?

jeremyrickard: Yeah, good point. Will update that.

pkg/services/cosmosdb/cosmosdb-deprovision.go
```diff
@@ -0,0 +1,41 @@
...
+	instance service.Instance,
+) (service.InstanceDetails, service.SecureInstanceDetails, error) {
+	err := deleteARMDeployment(c.armDeployer, instance)
+	if err != nil {
+		return nil, nil, err
+	}
+	return instance.Details, instance.SecureDetails, nil
+}
+
+func (c *cosmosManager) deleteCosmosDBServer(
```
krancour: Cosmos "server?" Copy/pasta, probably. Or a relic of the first iteration of this module. Can we change this to "account?"

jeremyrickard: Relic of the original module. I'll do a naming scrub

pkg/services/cosmosdb/common-types.go
```diff
@@ -58,47 +56,3 @@ type Credentials struct {
 	PrimaryConnectionString string `json:"primaryConnectionString,omitempty"`
 	PrimaryKey              string `json:"primaryKey,omitempty"`
 }
-
```
krancour: Adding this comment here because github doesn't let me comment unmodified code...

Several lines up from here, we define some types that are just empty structs for binding params and secure binding params. As of about three weeks ago, the broker is tolerant of nil binding params, so you can safely skip defining these types that are unused. Functions that return those types (e.x. `GetEmptyBindingParameters()`) can safely return `nil`.

I do not recall how tolerant the code is of nil `BindingDetails`. You can try. If it doesn't work, we can fix that in a separate PR. But this definitely works for params.

pkg/services/cosmosdb/catalog.go
krancour: @jeremyrickard should we consider changing the name of the manager to `cosmosAccountManager`? Presumably, there are database and/or collection managers (not totally clear yet) that will be added in a near-future PR.

pkg/services/cosmosdb/cosmosdb-deprovision.go
krancour: Ya... I think the original had copy/pasta because the postgres module was likely used as a template for this.

pkg/services/cosmosdb/catalog.go
jeremyrickard: Along those lines, do you think we'll add multiple account managers?

krancour: Probably so.

jeremyrickard: I left the _names_ the same for now, but cleaned up the the description to indicate A.) It's an account only and B, it's the specific API (either SQL or MongoDB). Will add the other APIs in another PR and if we do one to provision a collection on the account, will rename

pkg/services/cosmosdb/catalog.go
```diff
@@ -9,16 +9,16 @@ func (m *module) GetCatalog() (service.Catalog, error) {
 			service.NewService(
 				&service.ServiceProperties{
 					ID:   "6330de6f-a561-43ea-a15e-b99f44d183e6",
-					Name: "azure-cosmos-document-db",
-					Description: "Azure DocumentDB (Experimental) provided by CosmosDB " +
-						"and accessible via SQL (DocumentDB), Gremlin (Graph), and Table " +
-						"(Key-Value) APIs",
+					Name: "azure-cosmos-db-sql",
+					Description: "Azure Cosmos DB is a globally distributed, " +
```
krancour: This description is inconsistent with others. Every existing one I've looked at is just a noun (even if it's long like "Blah blah blah provided by blah blah"), but none of the others are complete sentences describing the service. That which is listed as the description here might be better suited as the `LongDescription` (under `Metadata`), but even then, it seems we haven't used complete sentences.

I know this is super nitpicky. If we can just stay consistent with these existing conventions for now, we can always circle back later and update descriptions and long descriptions across the board to comport with new conventions if we decide they need to change.

pkg/services/cosmosdb/catalog.go
```diff
@@ -9,16 +9,16 @@ func (m *module) GetCatalog() (service.Catalog, error) {
 			service.NewService(
 				&service.ServiceProperties{
 					ID:   "6330de6f-a561-43ea-a15e-b99f44d183e6",
-					Name: "azure-cosmos-document-db",
-					Description: "Azure DocumentDB (Experimental) provided by CosmosDB " +
-						"and accessible via SQL (DocumentDB), Gremlin (Graph), and Table " +
-						"(Key-Value) APIs",
+					Name: "azure-cosmos-db-sql",
```
krancour: I know you opted not to change the name just now, but I think we really need to be clear about _what_ each service provisions. Any thoughts on going with `azure-cosmosdb-sql-account`? (Note also that in addition to qualifying this with "account," I'm proposing to drop the "-" between "cosmos" and "db". I think it's unfortunate that the service has "DB" in its name as that can lead to some confusion. [For the same reason, I dislike "SQL Server" having "server" in its name.] By dropping the "-" between "cosmos" and "db", I hope to emphasize that this is a product name and _not_ mislead anyone into thinking that what they are provisioning is a database and not an account.)

Stop me if I'm talking crazy.

jeremyrickard: nope, that's cool.

pkg/services/cosmosdb/catalog.go
```diff
@@ -70,11 +64,11 @@ func (m *module) GetCatalog() (service.Catalog, error) {
 						"MongoDB",
 					},
 				},
-				m.serviceManager,
```
krancour: Github's not letting me comment on lines that weren't changed, but consistent with [this comment](https://github.com/Azure/open-service-broker-azure/pull/332/files#r174914255), about a dozen lines up from here, could we consider a name change to "azure-cosmosdb-mongo-account"?

pkg/services/cosmosdb/common-provision.go
```diff
@@ -0,0 +1,42 @@
...
+
+	"github.com/Azure/open-service-broker-azure/pkg/generate"
+	"github.com/Azure/open-service-broker-azure/pkg/service"
+	log "github.com/Sirupsen/logrus"
+	uuid "github.com/satori/go.uuid"
+)
+
+func generateDatabaseName(location string) string {
+	databaseName := uuid.NewV4().String()
+	// CosmosDB currently limits database name to 50 characters,
```
krancour: This actually a limitation on _account_ names, I think? Can comment and function name both be updated?

pkg/services/cosmosdb/common-types.go
```diff
@@ -0,0 +1,20 @@
+package cosmosdb
+
+type databaseKind string
+
+const (
+	databaseKindMongoDB          databaseKind = "MongoDB"
+	databaseKindGlobalDocumentDB databaseKind = "GlobalDocumentDB"
```
krancour: I think these are a relic of when we used one manager to provision mongo and non-mongo accounts alike.

Now that there are two separate managers, we can hard-code this bit of each information into each manager, which obviates the need for the `databaseKind` type, variables of that type, and tracking that information in the catalog "extras" and the instance details.

pkg/services/cosmosdb/common-provision.go
jeremyrickard: oh good catch!

contrib/k8s/examples/cosmosdb/cosmosdb-instance.yaml
```diff
@@ -5,7 +5,8 @@ metadata:
   name: my-cosmosdb-instance
   namespace: default
 spec:
-  clusterServiceClassExternalName: azure-cosmos-document-db
+  clusterServiceClassExternalName: azure-cosmosdb-sql-account
+  clusterServicePlanExternalName: account
```
krancour: Only question here is whether the file name should maybe be changed.

krancour: Applicable to other files in this dir also.

contrib/k8s/examples/cosmosdb/cosmosdb-instance.yaml
```diff
@@ -5,7 +5,8 @@ metadata:
   name: my-cosmosdb-instance
```
krancour: Actually... _after_ thinking the file name should change, maybe should this resource name also?

pkg/services/cosmosdb/catalog.go
```diff
@@ -2,23 +2,19 @@ package cosmosdb
...
 	return service.NewCatalog([]service.Service{
 			service.NewService(
 				&service.ServiceProperties{
-					ID:   "6330de6f-a561-43ea-a15e-b99f44d183e6",
-					Name: "azure-cosmos-document-db",
-					Description: "Azure DocumentDB (Experimental) provided by CosmosDB " +
-						"and accessible via SQL (DocumentDB), Gremlin (Graph), and Table " +
-						"(Key-Value) APIs",
+					ID:          "6330de6f-a561-43ea-a15e-b99f44d183e6",
+					Name:        "azure-cosmosdb-sql-account",
```
krancour: 👍

contrib/k8s/examples/cosmosdb/cosmosdb-instance.yaml
jeremyrickard: Let's take care of these in a follow up so we can begin the process of rebasing #347

PR #336, opened 2018-03-12
updated the path  of postgresql example 

vishnoisuresh:
**error**: the path "contrib/k8s/examples/postgresql-binding.yaml" does not exist
Correct Path is "contrib/k8s/examples/postgresql/postgresql-binding.yaml"


jeremyrickard:
Docs Only, no need for CI.

PR #337, opened 2018-03-12
refactor how (where) the catalog is built

krancour:
This builds on #333 and improves how/where we build the catalog.

Previously, we were handing a slice of modules to the broker along with module config that allowed the broker to know which modules to filter out. The broker would then build the catalog itself from the "mini catalogs" provided by each non-filtered-out module. The broker (the component that is) shouldn't be responsible for this.

Now, more of this is done by the bootstrap code in the `cmd`. This means that the resulting catalog can now be treated like a first-class component (it is, really) and passed in an already usable state to other components that need it.

This also makes it easier to programmatically provide a customized / abbreviated catalog during testing.

### DIFF
--- a/cmd/broker/broker.go
+++ b/cmd/broker/broker.go
@@ -59,8 +59,17 @@ func main() {
 		log.Fatal(err)
 	}
 
-	// Initialize modules
-	if err = initModules(azureConfig); err != nil {
+	// Initialize catalog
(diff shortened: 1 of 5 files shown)


krancour:
Note that some of the catalog-building code might reasonably move into the `service` package in the future, but it can't at the moment because there would be dependency cycles to resolve. This is good for now.

PR #340, opened 2018-03-13
Add FAQ

carolynvs:
@chzbrgr71 Ran into this today and I figured it would be helpful and not too much effort to track when things break, and how they were resolved.


jeremyrickard:
Docs only, no CI needed.

PR #344, opened 2018-03-15
[WIP] Removing exclamation marks from docs

arschles:
I want to remove the marks so folks are more likely to try out and use OSBA and we can get more feedback.

This patch keeps the "alpha" wording in.


arschles:
Closing because markdown was linted that I didn't intend (VSCode is too eager). Will reopen

PR #345, opened 2018-03-15
Removing exclamation marks from docs

arschles:
I want to remove the marks so folks are more likely to try out and use OSBA and we can get more feedback.

This patch keeps the "alpha" wording in.

Replaces #344


docs/modules/mssql.md
```diff
@@ -1,7 +1,6 @@
 # [Azure SQL Database](https://azure.microsoft.com/en-us/services/sql-database/)
 
-|![](https://upload.wikimedia.org/wikipedia/commons/thumb/1/17/Warning.svg/50px-Warning.svg.png) | This module is PREVIEW. It is under moderate development, but remains subject to the possibility of breaking changes if they prove necessary. |
-|---|---|
+_Note: this module is PREVIEW. It is under moderate development, but future releases may still break the API._
```
carolynvs: This is a super mega ulta nit but shouldn't it be "in preview"? Also do we define preview/experimental somewhere?

arschles: All the docs capitalize the preview/experimental word so I kept it as-is. Maybe change it wholesale in another PR?

Preview/experimental are defined in https://github.com/Azure/open-service-broker-azure/blob/master/docs/roadmap.md right now, will be moved to a dedicated document in https://github.com/Azure/open-service-broker-azure/pull/346

jeremyrickard:
No CI, docs only.

PR #347, opened 2018-03-15
Refactor integration between core/framework and service modules

krancour:
This PR pays down a lot of debt.

Broker architecture is divided into a core/framework that does most heavy lifting (persistence, coordinating async tasks, etc.) and modules that provide management capabilities for specific services. The core/framework is largely uninterested in service-specific details, and as such, such details have always been treated like a blob by the core/framework when persisting instances or reading them back from storage... with one caveat. In order to successfully unmarshal data into service-specific structs, the core/framework has always needed to be able to _ask_ each service manager for various _kinds_ of structs. This has lead to the proliferation of functions like `GetEmptyInstanceDetails()` (for instance) and many others on the `ServiceManager` interface. This has had some negative consequences...

1. Raises the barrier to entry for module development by increasing the number of functions that must be implemented.
1. Module implementors (myself included) have often felt unsure about how safe it is to return `nil` from these "get empty" functions, and this has lead to the proliferation of a lot of different types that are just aliases for `struct{}` and nothing more.
1. Worst implication-- before an instance or binding can be retrieved from persistent storage, the core/framework needs to identify the correct service manager for the instance or binding so that it can query the _correct_ service for manager for many different kinds of empty structs.

All of this is noise.

This PR updates the touchpoints between the core/framework and the modules so that all input to and output from module functions are _maps_. To the core/framework, these are undifferentiated blobs.

Maps can be cumbersome to work with, so this PR also helps modules preserve the cleanliness of their code by providing utility functions for easily converting between maps and service-specific structs.

This is an overall reduction in code that was high on ceremony and noise and low on essence and signal. It's also much easier to wrap one's head around.

Not that I'd advocate it, but a lazy module implementor could also now wheel and deal exclusively with maps. The conversions to/from service-specific structs are simply a style choice-- albeit a rational one.

Starting with this PR, modules also export _nothing_, which is another improvement.

Merging this should be deferred until #332 is in because it will require a non-trivial rebase.

@jeremyrickard _after_ #332 is in, let's walk these changes together since it's a large diff.


krancour:
Re-applied "do not merge." I might have a superior alternative to this. Hold off on review for now as well.

krancour:
> I might have a superior alternative to this.

I've spent sometime trying to prove out an alternative approach and it ends up not being as clean as what's in this PR. Re-removing "do-not-merge."

PR #351, opened 2018-03-20
refactor lifecycle tests

krancour:
I was starting to copy bits of the lifecycle tests to form the basis of some new e2e tests when I saw some easy cleanup to do.

Highlights:

- Instantiate the modules and catalog all in one place-- same as what the broker itself does on startup-- thus easier to understand
- Lifecycle tests cases are now _just_ data. They don't contain a reference to the module they test. The test framework simply finds the right service manager for each test case by looking it up in the catalog.
- Names of test cases are simplified.
- Functions for testing bindings/credentials eliminate a bit of unnecessary indirection. (We used to have a function return a binding/credential test function. Not sure why we did it that way, but it's not necessary.)

The biggest plus to this all, however, is that these changes will make the lifecycle tests look more like the e2e tests I'm pr'ing shortly. This will reduce the cognitive load required to understand _both_ test suites.

### DIFF
--- a/tests/lifecycle/test_cases_test.go
+++ b/tests/lifecycle/test_cases_test.go
@@ -6,85 +6,27 @@ import (
 	"log"
 	"os"
 	"strings"
-	"time"
-
-	resourcesSDK "github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2017-05-10/resources" // nolint: lll
-	"github.com/Azure/go-autorest/autorest"
-	azureSDK "github.com/Azure/go-autorest/autorest/azure"
-	"github.com/Azure/open-service-broker-azure/pkg/azure"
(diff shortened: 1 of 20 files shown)


tests/lifecycle/test_cases_test.go
```diff
@@ -6,88 +6,28 @@ import (
...
-		azureEnvironment.ResourceManagerEndpoint,
-		azureSubscriptionID,
-	)
-	resourceDeploymentsClient.Authorizer = authorizer
-	resourceDeploymentsClient.PollingDuration = time.Minute * 30
-	armDeployer := arm.NewDeployer(
-		resourceGroupsClient,
-		resourceDeploymentsClient,
-	)
+	testCases := rediscacheTestCases
```
jeremyrickard: 👍

Issue #93, opened 2017-11-30
Proposal: return instance schemas and binding schemas in catalog


Haishi2016:
Although it's not mandated by the spec, I think we should return instance schema and binding schema in catalog so that a downstream service (like a catalog service) can discover how to construct proper provisioning/binding requests. There are currently no way for such services to discover how to make such requests. IMHO schemas should have been mandatory fields in the spec.

carolynvs:
FYI, we are going to start showing these values in svcat, and it would be great if osba supported it. As otherwise I'm not sure how you would even know what to pass in.

https://github.com/kubernetes-incubator/service-catalog/issues/1720

krancour:
> As otherwise I'm not sure how you would even know what to pass in.

Not the ideal answer, but there is documentation for each service. Just wanted you to know there's _something_.

carolynvs:
Yeah... As I was developing svcat, the best way to figure out params was reading our own helm charts. I never found the real doc, so ... if we have it, it's not easy enough to find.

PR #355, opened 2018-03-21
Add JSON Schema to Catalog

jeremyrickard:
This PR adds some types related to producing JSON Schema for service_instance parameters. It defines a common schema type for resource_group, location and tags. Each module can then either use that
or extend it by adding to a map. The catalogs of all the modules were updated to include the schema in each plan: either the default schema or the module specific parameters. 

Some of the structs are modeled after the OSB Lib, but we aren't using that lib so I couldn't just pull it in as is. 

I think a better long term solution would be to define the schema and use a library
to generate the structs, but this is a good stop gap to support running on OpenShift
where parameter schemas are required. It's also a nice to have for anyone else as
it defines the properties that are required or optional. 

Closes: #93


jeremyrickard:
The end result of this is that the catalog is updated to include service_instance schemas for each plan. An example from the azure-cosmosdb-mongo-account service follows: 

```
 {
            "name": "azure-cosmosdb-mongo-account",
            "id": "8797a079-5346-4e84-8018-b7d5ea5c0e3a",
            "description": "Azure Cosmos DB Database Account (MongoDB API)",
            "metadata": {
                "displayName": "Azure Cosmos DB (MongoDB)",
                "imageUrl": "https://azure.microsoft.com/svghandler/cosmos-db/?width=200",
                "longDescription": "Globally distributed, multi-model database service (Experimental)",
                "documentationUrl": "https://docs.microsoft.com/en-us/azure/cosmos-db/",
                "supportUrl": "https://azure.microsoft.com/en-us/support/"
            },
            "tags": [
                "Azure",
                "CosmosDB",
                "Database",
                "MongoDB"
            ],
            "bindable": true,
            "plan_updateable": false,
            "plans": [
                {
                    "id": "86fdda05-78d7-4026-a443-1325928e7b02",
                    "name": "account",
                    "description": "Database Account with the MongoDB API",
                    "free": false,
                    "metadata": {
                        "displayName": "Azure MongoDB"
                    },
                    "schemas": {
                        "service_instance": {
                            "create": {
                                "parameters": {
                                    "$schema": "http://json-schema.org/draft-04/schema#",
                                    "type": "object",
                                    "properties": {
                                        "location": {
                                            "type": "string",
                                            "description": "The Azure region in which to provision applicable resources."
                                        },
                                        "resourceGroup": {
                                            "type": "string",
                                            "description": "The (new or existing) resource group with which to associate new resources."
                                        },
                                        "tags": {
                                            "type": "object",
                                            "description": "Tags to be applied to new resources, specified as key/value pairs.",
                                            "additionalProperties": {
                                                "type": "string"
                                            }
                                        }
                                    }
                                }
                            }
                        }
                    }
                }
            ]
        },
```

pkg/service/catalog.go
```diff
@@ -73,12 +73,13 @@ type service struct {
...
-	Free        bool                   `json:"free"`
-	Metadata    *ServicePlanMetadata   `json:"metadata,omitempty"`
-	Extended    map[string]interface{} `json:"-"`
+	ID               string                 `json:"id"`
+	Name             string                 `json:"name"`
+	Description      string                 `json:"description"`
+	Free             bool                   `json:"free"`
+	Metadata         *ServicePlanMetadata   `json:"metadata,omitempty"`
+	Extended         map[string]interface{} `json:"-"`
+	ParameterSchemas *ParameterSchemas      `json:"schemas,omitempty"`
```
jeremyrickard: From OSB client.

pkg/service/schema.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+// ParametersSchema represents the individual schema for a service
+type ParametersSchema struct {
+	Schema     string                 `json:"$schema"`
+	Type       string                 `json:"type"`
+	Properties map[string]interface{} `json:"properties,omitempty"`
+	Required   []string               `json:"required,omitempty"`
+}
+
```
jeremyrickard: Lines 3-37: OSB client.

pkg/service/schema.go
```diff
@@ -0,0 +1,83 @@
...
+// Parameter represents the individual schema for a given parameter
+type Parameter struct {
```
jeremyrickard: This type encapsulates all the bits of JSON Schema that we need to include AFAICT. The "Items" and "Properties" allow you to embed this object again for arrays. You can find some examples of that in some of the modules (specifically the Firewall Rules where we need to define complex types)

pkg/service/schema.go
```diff
@@ -0,0 +1,83 @@
...
+	p := &ParametersSchema{
+		Schema: "http://json-schema.org/draft-04/schema#",
+		Type:   "object",
+	}
+	return p
+}
+
+// GetCommonSchema builds a default schema object with
+// location, resource group and tags
+func GetCommonSchema() *ParametersSchema {
```
jeremyrickard: Common Schema that everything uses. Tags, Location and RG. Tags, being a freeform object, has no properties defined but defines an "Additional Properties" element that must be string, this basically allows key/value. The other two are simple strings.

pkg/services/aci/types.go
```diff
@@ -9,6 +9,44 @@ type provisioningParameters struct {
...
+	}
+
+	p.Properties["memoryInGb"] = service.Parameter{
+		Type: "integer",
+		Description: "Gigabytes of memory requested for the container. " +
+			"Must be specified in increments of 0.10 GB.",
+		Default: 1.5,
+	}
+
+	p.Properties["ports"] = service.Parameter{
```
jeremyrickard: Here is an example of an array with "Items"

pkg/services/mssql/common_types.go
```diff
@@ -20,3 +24,43 @@ type credentials struct {
...
+	p := service.GetCommonSchema()
+
+	p.Properties["sslEnforcement"] = service.Parameter{
+		Type: "string",
+		Description: "Specifies whether the server requires the use of TLS " +
+			"when connecting. Can be 'enabled', 'disabled' or ''. " +
+			"Left unspecified, SSL will be enforced",
+	}
+
+	firewallRuleSchema := map[string]interface{}{}
```
jeremyrickard: Here is an example of Firewall rules, which is a an array of "complex" objects

pkg/service/schema.go
```diff
@@ -0,0 +1,83 @@
+package service
+
+// ParameterSchemas is a set of optional JSONSchemas that describe
+// the expected parameters for creation and update of instances and
+// creation of bindings.
+type ParameterSchemas struct {
```
krancour: I'm worried about how close the name of this is to that of another type-- `ParametersSchema`.

Any thoughts about renaming this as `PlanSchema` (singular, too)? I feel that accurately reflect what this is

jeremyrickard: Sure! That works for me.

krancour: Actually, I take that back.

krancour: The name was tricky to wrap my head around, but I get it now.

pkg/service/schema.go
```diff
@@ -0,0 +1,83 @@
...
+	ServiceInstances *InstanceSchema `json:"service_instance,omitempty"`
+	ServiceBindings  *BindingSchema  `json:"service_binding,omitempty"`
+}
+
+// InstanceSchema represents a plan's schemas for creation and
+// update of an API resource.
+type InstanceSchema struct {
```
krancour: Can we make this `InstanceSchemas` (plural)? Just as `ParameterSchemas` is plural because its composed of multiple schemas, so to is this.

pkg/service/schema.go
```diff
@@ -0,0 +1,83 @@
...
+	Create *InputParameters `json:"create,omitempty"`
+	Update *InputParameters `json:"update,omitempty"`
+}
+
+// BindingSchema represents a plan's schemas for the parameters
+// accepted for binding creation.
+type BindingSchema struct {
```
krancour: For symmetry with the proposed name change from `InstanceSchema` to `InstanceSchemas` (see above), could we change this as well? `BindingSchemas`?

pkg/service/schema.go
```diff
@@ -0,0 +1,83 @@
...
+	Create *InputParameters `json:"create,omitempty"`
+}
+
+// InputParameters represents a schema for input parameters for creation or
+// update of an API resource.
+type InputParameters struct {
+	Parameters *ParametersSchema `json:"parameters,omitempty"`
+}
```
krancour: I think it might be easier to understand this if this were broken up into a few different types-- `ProvisioningParametersSchema`, `UpdatingParametersSchema`, and `BindingSchema`. Such naming makes things more tractable vis a vis existing broker types.

jeremyrickard: This is partially an artifact of the expected json that needs to get generated for the OSB spec:

pkg/service/catalog.go
krancour: There's an opportunity to simplify things here. First, I should articulate why `PlanProperties` is a separate thing from a `plan`. The short answer is simply that there was need for `ServiceProperties` to be a separate thing from a `service` and I wanted to keep things consistent...

The more complete answer thus requires us to understand why `ServiceProperies` is a separate thing from a `service`. I did not want module implementors directly instantiating `service` objects. This is why it's unexported from the `service` package, has a corresponding interface `Service` (which _is_ exported) and a constructor-like function `NewService(...)`.

Falling farther down the rabbit hole... Why did I not want module implementors instantiating `service` objects directly? Well... there are various things that I wanted to guarantee would happen automatically when a new `service` object is initialized-- namely I wanted to index plans by ID.

So-- coming back to `plan` and `PlanProperties`. The division (currently) exists only to mirror the division between `service` and `ServiceProperties`, _however_, the fact that this division exists means we can take advantage of the opportunity to do things in the constructor-like function `NewPlan(...)`.

What I'd like to propose is that the right properties for the `PlanProperties` type to expose here are `ProvisioningParametersSchema`, `UpdatingParametersSchema`, and `BindingParametersSchema`. (It should be valid for some or all of these to be `nil`.) The constructor-like function `NewPlan(...)` can then take accountability for wrapping these in the correct types like `instanceSchemas`, `bindingSchemas`, and the `parameterSchemas` type that wraps both of those. Note that the above three, then, do not even need to be exported. This reduces the surface area of the SDK (`service` package) that we expect module implementors to use.

The constructor-like function `NewPlan(...)` can also be a good place to supplement module-implementor-provided schema data with common things like `"http://json-schema.org/draft-04/schema#"` so that module-implementors do not need to concern themselves with inconsequential details.

pkg/service/catalog.go
```diff
@@ -260,8 +264,43 @@ func (s *service) GetChildServiceID() string {
...
+	// commonSchema := &service.ParameterSchemas{
+	// 	ServiceInstances: &service.InstancesSchema{
+	// 		Create: &service.ProvisioningParametersSchema{
+	// 			Parameters: service.GetCommonSchema(),
+	// 		},
+	// 	},
+	// }
+	//ProvisionParamSchema *ParametersSchema      `json:"-"`
+	//UpdateParamSchema    *ParametersSchema      `json:"-"`
+	//BindingParamSchema   *ParametersSchema      `json:"-"`
```
krancour: Are these deliberately left here for reference? Or were these meant to be removed?

pkg/service/schema.go
```diff
@@ -0,0 +1,96 @@
...
+// the expected parameters for creation and update of instances and
+// creation of bindings.
+type ParameterSchemas struct {
+	ServiceInstances *InstancesSchema `json:"service_instance,omitempty"`
+	ServiceBindings  *BindingsSchema  `json:"service_binding,omitempty"`
+}
+
+// InstancesSchema represents a plan's schemas for creation and
+// update of an a service instance resource.
+type InstancesSchema struct {
```
krancour: If we don't intend for anyone outside this package to instantiate this directly, we might consider not exporting this.

pkg/service/schema.go
krancour: Reconsidering the naming-- I think `planSchema` is good after all. Let's not export this if we don't need to.

pkg/service/schema.go
```diff
@@ -0,0 +1,94 @@
...
+
+// ParameterSchemas is a set of optional JSONSchemas that describe
+// the expected parameters for creation and update of instances and
+// creation of bindings.
+type ParameterSchemas struct {
+	ServiceInstances *instancesSchema `json:"service_instance,omitempty"`
+	ServiceBindings  *BindingsSchema  `json:"service_binding,omitempty"`
+}
+
+type instancesSchema struct {
```
krancour: s/instancesSchema/instanceSchemas/

pkg/service/schema.go
```diff
@@ -0,0 +1,94 @@
...
+	Create *ProvisioningParametersSchema `json:"create,omitempty"`
```
krancour: Can we rename the field as `ProvisioningParametersSchema`? It seems it's more easily tied back to existing tihngs this way?

pkg/service/schema.go
```diff
@@ -0,0 +1,94 @@
...
+	Update *UpdatingParametersSchema     `json:"update,omitempty"`
```
krancour: Can we rename the field as `UpdatingParametersSchema`? It seems it's more easily tied back to existing things this way?

pkg/service/schema.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+// ProvisioningParametersSchema represents the schema for any parameters that
+// might be needed for provisioning a service instance
+type ProvisioningParametersSchema struct {
```
krancour: Let's not export this if we don't need to.

pkg/service/schema.go
```diff
@@ -0,0 +1,94 @@
...
+	Parameters *ParametersSchema `json:"parameters,omitempty"`
+}
+
+// UpdatingParametersSchema represents the schema for any parameters that
+// might be needed for updating a service instance
+type UpdatingParametersSchema struct {
```
krancour: Let's not export this if we don't need to.

pkg/service/schema.go
```diff
@@ -0,0 +1,94 @@
...
+
+// BindingsSchema represents a plan's schemas for the parameters
+// accepted for binding creation.
+type BindingsSchema struct {
+	Create *BindingSchema `json:"create,omitempty"`
+}
+
+// BindingSchema represents the schema for any parameters that
+// might be needed for creating a binding
+type BindingSchema struct {
```
krancour: Let's not export this if we don't need to.

pkg/service/schema.go
```diff
@@ -0,0 +1,94 @@
...
+// ParametersSchema represents the individual schema for a service
+type ParametersSchema struct {
+	Schema     string               `json:"$schema"`
+	Type       string               `json:"type"`
+	Properties map[string]Parameter `json:"properties,omitempty"`
+	Required   []string             `json:"required,omitempty"`
+}
+
+// Parameter represents the individual schema for a given parameter
+type Parameter struct {
```
krancour: Can we change this to `ParameterSchema`?

pkg/service/schema.go
```diff
@@ -0,0 +1,94 @@
...
+	Type        string                 `json:"type"`
+	Description string                 `json:"description,omitempty"`
+	Default     interface{}            `json:"default,omitempty"`
+	Items       interface{}            `json:"items,omitempty"`
+	Properties  map[string]interface{} `json:"properties,omitempty"`
```
krancour: Wondering if the type of this field should be `map[string]* Parameter`. (Or `map[string]* ParameterSchema` if that type is indeed renamed as suggested in a previous comment.)

pkg/service/catalog.go
```diff
@@ -73,12 +73,15 @@ type service struct {
...
-	Extended    map[string]interface{} `json:"-"`
+	ID                    string                 `json:"id"`
+	Name                  string                 `json:"name"`
+	Description           string                 `json:"description"`
+	Free                  bool                   `json:"free"`
+	Metadata              *ServicePlanMetadata   `json:"metadata,omitempty"`
+	Extended              map[string]interface{} `json:"-"`
+	ProvisionParamsSchema *ParametersSchema      `json:"-"`
+	UpdateParamsSchema    *ParametersSchema      `json:"-"`
+	BindingParamsSchema   *ParametersSchema      `json:"-"`
```
krancour: It is possible that each of the above three attributes would be better off as type `map[string]* Parameter` (or `map[string]ParameterSchema` if renamed as suggested in another comment)?

pkg/service/schema.go
krancour: This also probably does not need to be exported if the suggestion about `PlanProperites` having `map[string]Parameter` field instead of `ParameterSchema` fields is taken.

pkg/service/catalog.go
jeremyrickard: I thought about that. In most cases where we want to provide a custom schema, we need to fetch the base properties. In that case, not sure it saves much. You need to fetch them because there _are_ some scenarios where we don't want to blindly apply them. The database only instances, for example, don't need the resource group and location. 

Can go either way with this, so if you'd like it changed to map[string]*ParameterSchema can make that change.

krancour: > In most cases where we want to provide a custom schema, we need to fetch the base properties.

Rather than have module implementors have to retrieve the base properties and weave in their own, can we let them provide their own and `NewPlan(...)` can supplement those properties with the base properties?

> You need to fetch them because there are some scenarios where we don't want to blindly apply them. The database only instances, for example, don't need the resource group and location.

I _think_ we should be able to figure out in `NewPlan(...)` when we do and do not need to weave in the base properties. We would base it on whether the `ServiceProperties` express that the `Service` has a parent service...

The problem with that approach, however, is probably that `NewPlan(...)` doesn't have any knowledge of the `ServiceProperties`. I guess what we could do is actually make it the `NewService(...)` function iterate over all plans and augment their properties if and only if the `ServiceProperties` express that the service has a parent.

I think this would completely obviate the need for module implementors to ever concern themselves with a default / common set of properties.

Thoughts?

jeremyrickard: Can do. This will be a little bit more involved, so probably check on this tomorrow :)

pkg/service/catalog.go
```diff
@@ -176,8 +180,19 @@ func NewService(
...
 	}
-	for _, plan := range s.plans {
-		s.indexedPlans[plan.GetID()] = plan
+	for _, planIfc := range s.plans {
+		p := planIfc.(*plan)
+		if serviceProperties.ParentServiceID == "" {
+			if p.ParameterSchemas == nil {
+				p.ParameterSchemas = &planSchemas{}
+			}
+			p.ParameterSchemas.addParameters(
```
jeremyrickard: In schema.go, I've created two "addParameters()" methods. This one is on a *planSchemas receiver and takes in the three types of param schemas we can have now. I wanted to contain/centralize all that logic so it was reuable in different spots and we can mostly replace the unexported structs with maps later and just update those methods.

pkg/service/catalog.go
```diff
@@ -271,6 +299,28 @@ func NewPlanFromJSON(jsonBytes []byte) (Plan, error) {
 	if err := json.Unmarshal(jsonBytes, p); err != nil {
 		return nil, err
 	}
+	if p.ParameterSchemas.ServiceInstances != nil {
```
jeremyrickard: the following lines are for loading the catalog from json and repopulating the PlanProperties correctly.

krancour: I'm wondering if it's really necessary to straighten out the properties. The original intention was that properties would only be used as input to `NewPlan(...)`. After that, they shouldn't matter...

I realize, as I type this, that the fact that a `plan` is _composed_ of `PlanProperties` makes the statement above less obvious, or perhaps not entirely true.

Two things:

1. I'd like to see if we can reasonably mitigate the "wrong" composition about in a separate PR. I have some idea of how it can be done, but it might not pan out.

2. The `NewPlanFromJSON (...)` function will be dead code after #357 is addressed.

Bottom line is I don't think we need this code here, but there's also no harm in it staying for now.

pkg/services/aci/types.go
```diff
@@ -9,6 +9,45 @@ type provisioningParameters struct {
 	Ports       []int   `json:"ports"`
 }
 
+// GetProvisionParametersSchema generates the schema for instance
+// provisioning parameters
+func GetProvisionParametersSchema() map[string]*service.ParameterSchema {
```
krancour: Two things about this function...

1. Can we avoid exporting this? As it stands currently, the modules now export nothing except their "constructor" and the functions that the `ServiceModule` or `ServiceManager` interfaces require them to implement. The idea of modules exporting nothing, in my mind, helps keep things nice and tidy.

2. I see that for some modules that expose multiple services (where each service has potentially different parameters), there are multiple functions for returning parameter schemas. I'd like it if, in all cases (e.g. even for modules with a single service manager, we could make these "get schema" helper functions use the appropriate service manager as a receiver. imo, this is another way to keep things tidy and comprehensible, by grouping all functions related to a given service under its service manager.

krancour: The above is applicable to all modules.

jeremyrickard: 👍

pkg/services/mssql/catalog.go
```diff
@@ -518,6 +542,7 @@ func (m *module) GetCatalog() (service.Catalog, error) {
 						"1200 DWUs",
 					},
 				},
+				ProvisionParamsSchema: GetDBaseProvisionParametersSchema(),
```
krancour: Looking at this particular catalog raised a yellow flag for me...

I do realize the OSB spec allows schemas to be defined plan-by-plan and we have to play ball with that, _but_ we've made an architectural decision that all plans of a given service vary only in QoS, and thus have common sets of parameters.

Would it make more sense, for us and for module implementors if we allow the schemas to be added to the `ServiceProperties` instead of the `PlanProperties` and then we just deal with "spraying" those schemas from the services down to their constituent plans behind the scenes?

krancour:
Made a few suggestions, but I think this is almost all the way there.

pkg/services/mssql/catalog.go
jeremyrickard: Yes, I think it would. Implemented as-is because of the broker spec, but I think it would simplify things.

PR #356, opened 2018-03-21
add e2e tests

### LINKED ISSUE #224: Proposal: Kill two birds with one stone-- expand lifecycle tests into e2e tests
The entry point for executing the service lifecycle tests is currently the aync engine. i.e. HTTP requests, request validation, etc. are not involved. These effectively test "from the engine on down," but fall short of being real e2e tests.

Rather than add a separate suite of tests, I want to propose that we amend the current suite of lifecycle tests to _become_ e2e.

@jeremyrickard any thoughts?

krancour:
Fixes #224 

This PR adds happy path e2e tests. I'd like if one day these can replace the lifecycle tests, but I'm stopping short of making that swap right now because I'm not yet confident in our ability to correlate e2e test failures to their causes.

For now, just having this ability is an improvement, as it alleviates the need to conduct e2e manually before a release.

Highlights:

- Moves broker client code from `contrib/pkg/client/` to `pkg/client/` because it's no longer used exclusively by other things in `contrib/`. i.e. The new tests use it.
- Uses docker compose to link the e2e test container to a real, running broker.

To try these:

```
$ make test-e2e
```

You can optionally run a subset of the tests:

```
$ TEST_MODULES=aci,storage make test-e2e
```


krancour:
These require a non-trivial rebase and although they do work, they are not yet a suitable replacement for lifecycle tests. Choosing to defer this to v0.11.0.

krancour:
Closing this PR. These tests proved valuable for asserting no functionality was changed by a recent PR, so I snuck them in along with it.

PR #360, opened 2018-03-23
use docker multistage build to build binary and image

krancour:
We used to use the `build` target to build a binary and then `docker-build` to package that binary and a CA cert bundle into a `FROM scratch` image. Now we can do all of that in one shot using a multistage build and the `build` target alone. (I thought "build" sounded better than "docker-build," so that's the name we kept.)

This is one of several baby steps I want to make toward optimizing our workflows to 1. not require VM executors and 2. be more portable across various CI platforms.


Makefile
```diff
@@ -205,21 +198,25 @@ stop-broker-redis: check-docker-compose
...
+		--build-arg LDFLAGS='$(LDFLAGS)' \
+		-t $(RC_IMAGE_NAME) \
+		.
 	docker tag $(RC_IMAGE_NAME) $(RC_MUTABLE_IMAGE_NAME)
 
 # Push the release candidate Docker images
-.PHONY: docker-push-rc
-docker-push-rc: check-docker docker-build
+.PHONY: push-rc
+push-rc: check-docker docker-build
```
krancour: There's a bug on this line. Will fix in the morning.

krancour: Fixed

PR #361, opened 2018-03-23
Update ComsosDB to have ip range filters.

### LINKED ISSUE #326: CosmosDB: Support ipRangeFilter
Our current ARM template/provision parameters don't support firewall rules for for CosmosDB. We should add this for parity with with the RDBMS offerings.

jeremyrickard:
Updated CosmosDB to have optional ip range filters at provision time. Also did little DRY cleanup by embedding the cosmosAccountManager into the mongo service and removing unneeded extra methods and code.

Closes #326


jeremyrickard:
@krancour note this doesn't have the schema changes, so i'll rebase this off that if it merges first, or update the other one to introduce them. Your call.

pkg/services/cosmosdb/common-provision.go
```diff
@@ -57,11 +95,48 @@ func (c *cosmosAccountManager) preProvision(
...
+
+	if pp.IPFilterRules != nil {
+		filters := []string{}
+		if pp.IPFilterRules.AllowAzure != "disable" {
+			filters = append(filters, "0.0.0.0")
+		} else if pp.IPFilterRules.AllowPortal != "disable" {
+			// Azure Portal IP Addresses per:
+			// https://aka.ms/Vwxndo
+			// Region	IP address
+			// All regions except those specified below
```
krancour: I don't understand this... it says all regions _except those specified below_, but _all_ the IPS in the comment appear in the list of filters. This comment needs some formatting fixes, too.

jeremyrickard: Sorry, formatting messed this up. 

All Regions (except german, china, us gov) are covered by: 
40.76.54.131, 52.176.6.30, 52.169.50.45, 52.187.184.26

Germany is 51.4.229.218
China is 139.217.8.252
US Gov is 52.244.48.71

I added them *all* to the list since we don't readily have access to the cloud we're deployed into at this point.  The comment was mostly to document which is which, so i'll reformat it into maybe a tablular format?

jeremyrickard: This is also reproducing the behavior you get in the portal. You can allow access to all azure OR you can allow access only to the portal IPs if you've restricted access to some other list of IPs.

pkg/services/cosmosdb/cosmosdb.go
```diff
@@ -23,8 +23,7 @@ type sqlAccountManager struct {
 }
 
 type mongoAccountManager struct {
-	armDeployer            arm.Deployer
-	databaseAccountsClient cosmosSDK.DatabaseAccountsClient
+	cosmosAccountManager
```
krancour: I was momentarily concerned that the polymorphic behavior of certain functions having a `mongoAccountManager ` receiver wouldn't properly "override" the functions of the same name having the composed `cosmosAccountManager` as a receiver, but I read up and also crafter an example in the playground and it all checks out. This is a nice improvement.

docs/modules/cosmosdb.md
```diff
@@ -23,6 +23,10 @@ Provisions a new CosmosDB database account that can be accessed through any of t
 | `location` | `string` | The Azure region in which to provision applicable resources. | Required _unless_ an administrator has configured the broker itself with a default location. | The broker's default location, if configured. |
 | `resourceGroup` | `string` | The (new or existing) resource group with which to associate new resources. | N | If an administrator has configured the broker itself with a default resource group and nonde is specified, that default will be applied, otherwise, a new resource group will be created with a UUID as its name. |
 | `tags` | `map[string]string` | Tags to be applied to new resources, specified as key/value pairs. | N | Tags (even if none are specified) are automatically supplemented with `heritage: open-service-broker-azure`. |
+| `ipFilters` | `object` | IP Range Filter to be applied to new CosmosDB account | N | No Filters Applied |
+| `ipFilters.allowAccessFromAzure` | `string` | Specifies if Azure Services should be able to access the CosmosDB account.Valid valued are `""` (unspecified), `enabled`, or `disabled`. | N | If left unspecified, defaults to enabled. |
```
krancour: > If left unspecified, defaults to enabled.

Is that contrary to what you told me current behavior is? (fwiw, I do think this is what we should default to.)

jeremyrickard: Nope, so I probably need to update this comment.

Default behavior, if you don't provide *any* filters is that nothing is applied and it's open to the world.

*If* you provide any filters, it defaults to allow access from all of azure in addition to anything else you provide.

krancour:
I believe you need to update the `SplitProvisioningParameters(...)` function in `common-types.go`. Without doing so, the framework is using `nil` as the non-secure provisioning parameters that get persisted and passed to all subsequent provisioning steps.

krancour:
Indeed, given these params:

```
{"location":"eastus","resourceGroup":"cosmos-demo","ipFilters":{"filters":["1.2.3.4"]}}
```

This is what I see persisted in Redis:

```
{
    "instanceId": "f5e912c8-1183-4eab-b749-8db06fc3356a",
    "alias": "",
    "serviceId": "6330de6f-a561-43ea-a15e-b99f44d183e6",
    "planId": "71168d1a-c704-49ff-8c79-214dd3d6f8eb",
    "provisioningParameters": null,
    "secureProvisioningParameters": "bnVsbA==",
    "updatingParameters": null,
    "secureUpdatingParameters": "bnVsbA==",
    "status": "PROVISIONING",
    "statusReason": "",
    "location": "eastus",
    "resourceGroup": "cosmos-demo",
    "parentAlias": "",
    "tags": {},
    "details": {
        "armDeployment": "a56b70c6-4af1-40e0-a6f0-86b5a41361d4",
        "fullyQualifiedDomainName": "",
        "ipFilters": "",
        "name": "23e68311-5ab4-4cdf-b54e-9556ee5f4c2f"
    },
    "secureDetails": "bnVsbA==",
    "created": "2018-03-26T15:34:43.161975636Z"
}
```

pkg/services/cosmosdb/common-provision.go
krancour: Got it. So just fix formatting, I guess.

docs/modules/cosmosdb.md
krancour: imo, we should be more secure by default. So I like the doc as it is. Can we just bring the code in-line with the doc?

jeremyrickard:
Thanks for catching the parameter thing @krancour!

docs/modules/cosmosdb.md
```diff
@@ -23,7 +23,7 @@ Provisions a new CosmosDB database account that can be accessed through any of t
 | `location` | `string` | The Azure region in which to provision applicable resources. | Required _unless_ an administrator has configured the broker itself with a default location. | The broker's default location, if configured. |
 | `resourceGroup` | `string` | The (new or existing) resource group with which to associate new resources. | N | If an administrator has configured the broker itself with a default resource group and nonde is specified, that default will be applied, otherwise, a new resource group will be created with a UUID as its name. |
 | `tags` | `map[string]string` | Tags to be applied to new resources, specified as key/value pairs. | N | Tags (even if none are specified) are automatically supplemented with `heritage: open-service-broker-azure`. |
-| `ipFilters` | `object` | IP Range Filter to be applied to new CosmosDB account | N | No Filters Applied |
+| `ipFilters` | `object` | IP Range Filter to be applied to new CosmosDB account | N | A default filter is created that allows only Azure service access |
```
krancour: 👍

pkg/services/cosmosdb/common-provision.go
```diff
@@ -97,28 +97,36 @@ func (c *cosmosAccountManager) preProvision(
...
+			//||===================||================||
+			//|| All other regions || 104.42.195.92  ||
+			//||                   || 40.76.54.131   ||
+			//||                   || 52.176.6.30    ||
+			//||                   || 52.169.50.45   ||
+			//||                   || 52.187.184.26  ||
+			//=======================================||
+			// Given that we don't really have context of the cloud
+			// we are provisioning with right now, use all of the above
+			// addresses.
```
krancour: This is not a blocker, but I am now wondering if we can get this stuff from the Azure `Environment` object.

pkg/services/cosmosdb/common-types.go
```diff
@@ -3,7 +3,7 @@ package cosmosdb
 import "github.com/Azure/open-service-broker-azure/pkg/service"
 
 type provisioningParameters struct {
-	IPFilterRules *ipFilterRule `json:"ipFilters"`
+	IPFilterRules *ipFilterRule `json:"ipRangeFilters"`
```
krancour: I think the docs didn't get this same change.

pkg/services/cosmosdb/common-types.go
```diff
@@ -39,7 +39,13 @@ func (c *cosmosAccountManager) SplitProvisioningParameters(
...
 	error,
 ) {
-	return nil, nil, nil
+
+	pp := &provisioningParameters{}
+	if err := service.GetStructFromMap(cpp, &pp); err != nil {
+		return nil, nil, err
+	}
+	ppMap, err := service.GetMapFromStruct(pp)
+	return ppMap, nil, err
```
krancour: 👍

pkg/services/cosmosdb/common-provision.go
```diff
@@ -57,11 +95,59 @@ func (c *cosmosAccountManager) preProvision(
...
+				"51.4.229.218",
+				"139.217.8.252",
+				"52.244.48.71",
+			)
+		}
+		for _, filter := range pp.IPFilterRules.Filters {
+			filters = append(filters, filter)
+		}
+
+	} else {
```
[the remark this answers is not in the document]
krancour: Do we really want to default this here, close to the back end? Or do we want to default it in the params? I think the latter would be my preference.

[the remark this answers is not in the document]
jeremyrickard: Yeah, I think that we _should_ put them in the params. I think there are some other modules that are also defaulting (i think). I'll verify, but I think I'd like to do that in one swoop to make sure it's consistent across modules.

pkg/services/cosmosdb/common-types.go
krancour: I think I like the original name of `ipFilters` better...

My rationale is that this object also include options `allowAccessFromAzure` and `allowAccessFromPortal`, which (as a user) don't feel like ranges.

jeremyrickard: ok, will revert :)

krancour: In a similar vein, I wonder if the `filters` property of this object should maybe be renamed to `allowedIPRanges`

pkg/services/cosmosdb/common-provision.go
```diff
@@ -110,7 +110,7 @@ func (c *cosmosAccountManager) buildGoTemplateParams(
 		} else if pp.IPFilterRules.AllowPortal != "disable" {
 			// Azure Portal IP Addresses per:
 			// https://aka.ms/Vwxndo
-			//|| Region	           || IP address(es) ||
+			//|| Region            || IP address(es) ||
```
krancour: 👍

krancour:
Needs a non-trivial rebase since #355 was just merged.

pkg/services/cosmosdb/common-provision.go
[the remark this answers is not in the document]
jeremyrickard: @krancour I opened up https://github.com/Azure/open-service-broker-azure/issues/366 to do this here and with the RDBMS modules w/ firewall rules. I've already started on it in another branch.

pkg/services/cosmosdb/common-provision.go
```diff
@@ -57,11 +94,58 @@ func (c *cosmosAccountManager) preProvision(
...
+	kind string,
 ) map[string]interface{} {
 	p := map[string]interface{}{}
 	p["name"] = dt.DatabaseAccountName
-	p["kind"] = "GlobalDocumentDB"
+	p["kind"] = kind
+
+	filters := []string{}
+	if pp.IPFilterRules != nil {
+		if pp.IPFilterRules.AllowAzure != "disable" {
```
krancour: When you validated this, that check was case insensitive because you downcased input before comparing, but here you haven't done that. This could lead to some surprising behavior.

jeremyrickard: Oooooh! Will fix!

PR #364, opened 2018-03-26
refactor compliance checks

krancour:
Whether we stay on Circle or move to Jenkins, it's a safe assumption either way that we want to reduce our pipeline's dependency on (comparatively) heavy VMs and maximize the use of container executors. One consequence of that is that we want to avoid "D-in-D" scenarios as much as possible, so it's not desirable for OSBA's pipeline to build the OSB checker image like it does today.

Over the weekend, I extracted the OSB checker into its own image so that the OSBA CI pipeline doesn't need to build it itself.


krancour:
Our pipeline is configured (for now) to _not_ fail on compliance failures, so it's a good idea to look at the output to see that this refactor works as expected:

https://circleci.com/gh/Azure/open-service-broker-azure/2749

We have 11 failures-- same as before.

krancour:
This failed lifecycle tests on a known flake-- deleting Redis timed out. I think it's quite clear that the scope of this PR has no bearing on those particular tests, so I'm making an executive decision to merge this.

PR #365, opened 2018-03-26
refactor Makefile

krancour:
This is a refactoring of the Makefile to allow (for most targets) the option to set an environment variable `SKIP_DOCKER` that signals to the target(s) that the target is already being executed from within an appropriate containerized environment and there is no need to wrap the command in Docker / Docker Compose `run` commands.

Note that I strongly believe that the containerized development environment is an essential tool for ensuring consistent builds / tests across all contributors. This goal of this "skip docker" option is _not_ to officially support "native" builds. Rather, the goal is for all these make targets to remain usable in cases where a CI system (whether Circle, Jenkins, or other) is using a container executor and shouldn't do "D-in-D."

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,48 @@
+################################################################################
+# Version details                                                              #
+################################################################################
+
 GIT_VERSION = $(shell git describe --always --abbrev=7 --dirty)
 
+ifeq ($(REL_VERSION),)
+	BROKER_VERSION := devel
+else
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,7 +1,49 @@
+################################################################################
+# Version details                                                              #
+################################################################################
```
jeremyrickard: thanks for adding these comments!

PR #368, opened 2018-03-28
Fix the broken doc links

bingosummer:


jeremyrickard:
This is a docs only change, no CI required. Thanks @bingosummer!

PR #369, opened 2018-03-28
use container executors only in circle

krancour:
This PR transitions us off of machine (VM) executors in CIrcle and onto container executors. Why? Because Circle has been unreliable lately and acquiring VM executors (a limited resource) seems definitely to have been one of the problems. Also, adapting the build to occur _exclusively_ within containers is another step toward portability to some other CI system where we will assume _only_ container-based executors are available.

Most of this PR is very straightforward. The one part that bears careful explanation is the OSB API compliance tests. See the comments in tests.sh for further details.

@jeremyrickard I have heavily annotated the changes in-line to make the review a bit easier.


.circleci/config.yml
```diff
@@ -1,65 +1,142 @@
 version: 2
 jobs:
   lint:
-    machine: true
+    docker:
+      - image: quay.io/deis/lightweight-docker-go:v0.2.0
+    environment:
+      SKIP_DOCKER: true
+      CGO_ENABLED: 0
+    working_directory: /go/src/github.com/Azure/open-service-broker-azure
```
krancour: You'll see these previous 6 lines recur throughout this file. We want to use a specific Docker image for a container executor, instruct any make targets used by the job that they're _already_ running in an acceptable container so the command won't be "wrapped" in `docker run`, and we disable cgo, cause ain't nobody like cgo, amiright?

krancour: It would be nice if I could DRY this up, but there's no way to dry up YAML. C'est la vie.

.circleci/config.yml
```diff
@@ -1,65 +1,142 @@
...
+      # linter works flawlessly in all environments (containers included)
+      # EXCEPT a container in Circle. I've remoted into running containers in
+      # Circle and have attempted, at great length, to diagnose the issue, but
+      # am unable to do so. What I did discover, however, was that by
+      # re-installing gometalinter (and the linters it, in turn, uses), I can
+      # make the problem go away. ODDLY, when comparing the binaries that were
+      # included in the image to the NEW binaries, diff says they are identical.
+      # I'm at a complete loss for why this happens. I can live with this
+      # workaround for now because I anticipate our days using Circle for CI
+      # are numbered anyhow.
```
krancour: I spent nearly a whole day trying to pin down the cause of this. I give up. This workaround works. I couldn't tell you why.

docker-compose.yml
```diff
@@ -1,6 +1,6 @@
 version: "3.2"
 services:
-  dev: # 
+  dev:
```
krancour: I'm not going to open a separate PR to remove one errant character, so that's that.

docker-compose.yml
```diff
@@ -10,20 +10,26 @@ services:
     network_mode: host
   test: # Like dev, but linked to redis
     image: quay.io/deis/lightweight-docker-go:v0.2.0
+    environment:
+      STORAGE_REDIS_HOST: redis
+      ASYNC_REDIS_HOST: redis
```
krancour: Added the two vars above because when we're using compose to cobble together the environment for testing (i.e. locally) or Circle (i.e. CI), the Redis container we're linked to is addressed differently. In Circle, it's `localhost`. (Multiple containers involved in a single job there share a local network, much as do two containers in the same pod in k8s.) In compose, we have to name the link instead. These new variables let us assert come control over whether we talk to `localhost` or `redis`.

pkg/async/redis/engine_test.go
```diff
@@ -436,7 +437,7 @@ func TestRunRespondsToCanceledContext(t *testing.T) {
 // amend these overrides to test specific scenarios.
 func getTestEngine() *engine {
 	config := NewConfigWithDefaults()
-	config.RedisHost = "redis"
+	config.RedisHost = os.Getenv("ASYNC_REDIS_HOST")
```
krancour: See [this comment](https://github.com/Azure/open-service-broker-azure/pull/369/files#r177846967) for context.

pkg/storage/store_test.go
```diff
@@ -32,8 +33,7 @@ func init() {
 	}
 	fakeServiceManager = fakeModule.ServiceManager
 	config := NewConfigWithDefaults()
-	config.RedisHost = "redis"
-	config.RedisDB = 1
+	config.RedisHost = os.Getenv("STORAGE_REDIS_HOST")
```
krancour: See [this comment](https://github.com/Azure/open-service-broker-azure/pull/369/files#r177846967) for context.

tests/api-compliance/localhost-config.json
```diff
@@ -0,0 +1,81 @@
+{
```
krancour: This is nearly identical to `broker-config.json` (formerly named `config.json`). The only difference is the host that each of these address-- `broker` (compose) or `localhost` (circle).

The rationale for this is the same as explained in [this comment](https://github.com/Azure/open-service-broker-azure/pull/369/files#r177846967).

tests/api-compliance/scripts/test.sh
```diff
@@ -0,0 +1,70 @@
...
+# one application, in this one place, we will install node and the OSB checker
+# just-in-time into the running, PRIMARY go container. We can live with this
+# short term whilst we avoid consuming as-of-late unreliable Circle machine
+# executors, whilst knowing that a more elegant solution to this exists if/when
+# we move to Jenkins.
+#
+# If the workaround seems to do some inefficient things, its deliberate for the
+# sake of keeping these edits isolated to one spot in the script and easily
+# reversed if and when the constraints that led us here are ever lifted.
+
```
krancour: So to be clear, when running in Circle, we are _not_ using the osb-checker Docker image I created recently. It just doesn't work out. I've copied over the scripts (this file, for instance) from that image and made minimal, well-documented changes to them. These modifications install the osb checker into our go dev environment and this allows us to run the (dummy) broker _and_ the osb-checker in the same container-- which is all that Circle will let us get away with. See comment above for the rest of the context.

This stuff can go away when we kiss Circle goodbye.

tests/api-compliance/scripts/wupiao.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
```
krancour: Another script copied from the osb-checker image.

Makefile
```diff
@@ -216,7 +216,7 @@ LINT_CMD := gometalinter ./... \
 	--enable unparam \
 	--enable lll \
 	--line-length 80 \
-	--deadline 120s \
+	--deadline 240s \
```
krancour: Probably a resource issue-- lint tests seem to take a little longer in a container executor. nbd

pkg/async/redis/cleaner_test.go
```diff
@@ -21,35 +22,44 @@ func TestDefaultCleanCleansDeadWorkers(t *testing.T) {
 	}
 
 	// Override the default cleanActiveTaskQueue function to just count how many
-	// times it is invoked
+	// times it is invoked and then notify us when its been invoked workerCount
+	// times.
```
krancour: This whole test was re-written ti mitigate a known flake. Why is it part of this PR? Because the flake is manifesting more frequently with the unit tests running in a container executor. I'll make it a separate PR if anyone wants to push back, but I'd rather not.

.circleci/config.yml
```diff
@@ -1,65 +1,142 @@
...
+          name: Copy Modified Test Scripts
+          command: |
+            mkdir -p /app
+            cp tests/api-compliance/scripts/* /app/
       - run:
           name: Run API Compliance Tests
-          command: make test-api-compliance
+          command: |
+            go run cmd/compliance-test-broker/compliance-test-broker.go &
+            make test-api-compliance
```
krancour: These need to be in the same step. I tried to break them up, but even with `nohup`, I couldn't get the broker to keep running between steps.

Issue #370, opened 2018-03-29
Add "enum" and "default" to schemas where appropriate


jeremyrickard:
The use of the "enum" attribute allows you to specify valid values in JSON Schema. We should add this for things like the sslEnabled, and new things in the cosmosdb ip range.

PR #372, opened 2018-04-02
Refactor of schema code

jeremyrickard:
Refactor of the schema code to better handle complex, nested
objects that have required fields.

Closes: #370


pkg/service/schema.go
```diff
@@ -26,47 +32,143 @@ type bindingSchema struct {
...
-) {
-	if s.Properties == nil {
-		s.Properties = newParams
-		for key, param := range newParams {
-			if param.Required {
-				s.Required = append(s.Required, key)
-			}
+// ParameterSchema defines an interface representing a given Parameter schema.
+// This could be a provision or binding parameter.
+type ParameterSchema interface {
```
krancour: Can you explain the shift to this being an interface. I'm sure there is a good reason, but it's lost on me at the moment.

pkg/services/postgresql/dbms_types.go
```diff
@@ -9,15 +9,14 @@ type dbmsProvisioningParameters struct {
...
-
+) getProvisionParametersSchema() map[string]service.ParameterSchema {
 	p := getDBMSCommonProvisionParamSchema()
-
-	p["alias"] = &service.ParameterSchema{
-		Type:        "string",
-		Description: "Alias to use when provisioning databases on this DBMS",
-		Required:    true,
-	}
+	aliasSchema := service.NewSimpleParameterSchema(
```
krancour: Should have caught this in the first iteration of this feature...

I see the above repeated in a few places. Shouldn't "alias" be dealt with once wherever location and resource group are dealt with? (Those are conditionally required if a service doesn't have a parent. Alias should be conditionally required if a service doesn't have any children.)

pkg/service/schema.go
jeremyrickard: There were two reasons:

1.) I thought we wanted to not export any of it in the first pass at adding the schema stuff
2.) I found it useful to force people to create these through the constructor like methods (i.e. an object has properties, but an array has a field that defines what the items in the array are) so the structure is a little complicated. I thought that the constructor like methods hid some of that complexity and took some of the conceptual burden off the module writer. 

Can go back to exposing the struct if you'd prefer though, I think the struct had benefits as well (like reading the code was a little clearer because of the struct literals)

pkg/services/postgresql/dbms_types.go
jeremyrickard: Yeah, this is a good idea! Should have handled that in the first pass through here. Moving that up to the service is a great idea.

krancour:
https://github.com/Azure/open-service-broker-azure/pull/372/commits/b0562e302167ac3d0735c9c9a9ef6cfff1d11f7e looks good. Just waiting to hear what you want to do about interfaces vs structs.

pkg/service/schema.go
```diff
@@ -26,47 +30,152 @@ type bindingSchema struct {
...
-) {
-	if s.Properties == nil {
-		s.Properties = newParams
-		for key, param := range newParams {
-			if param.Required {
-				s.Required = append(s.Required, key)
+// ObjectParameterSchema represents the attributes of a complicated schema type
+// that can have nested properties
+type ObjectParameterSchema struct {
+	Type               string                     `json:"type"`
```
krancour: I get why `Type` is a field on `SimpleParameterSchema`, because in that case, the field could be a string, an integer, float, etc., but on `ObjectParameterSchema` and `ArrayParameterSchema`, shouldn't `Type` be implied by... the type?

I know it's required by the JSON, but maybe there's a way to hide it here? i.e. Hide it from users who are using these types to craft a schema?

I'm imagining that you could do something like remove `Type` here and then introduce a new, non-exported type `objectParameterSchema` that is composed of the exported `ObjectParameterSchema` and adds in the `Type` field. The same goes for `RequiredProperties`, which I believe you might like to also hide from module implementors.

pkg/service/schema.go
```diff
@@ -52,20 +54,50 @@ type SimpleParameterSchema struct {
...
+	return json.Marshal(struct {
+		Type               string                     `json:"type"`
+		Description        string                     `json:"description,omitempty"`
+		RequiredProperties []string                   `json:"required,omitempty"`
+		Properties         map[string]ParameterSchema `json:"properties,omitempty"`
+		Additional         ParameterSchema            `json:"additionalProperties,omitempty"` // nolint: lll
+	}{
+		Type:               "object",
+		Description:        o.Description,
+		RequiredProperties: o.requiredProperties,
```
krancour: I'm wondering if this is a better place than `NewService(...)` to walk the schemas for the constituent properties to build the `requiredProperties` array.

Any thoughts? I don't feel strongly one way or the other. It's just something that occurred to me.

jeremyrickard: Ahh yeah. I meant to move that _back_ into the `addProperties` method and had forgotten! I've moved that back in there.  It was orignally in the `addProperties` on the interface and when i nuked it I forgot to just make the `addProperties(..)` recursively handle the new things being added to the top level schema.

krancour:
LGTM. Will merge pending CI.

PR #376, opened 2018-04-04
fix chart publishing issue

krancour:
There's no point letting CI run on this. This change is only applicable to the CD / release pipeline.


krancour:
Will wait for #372 to be merged so that we don't need to rebase it and re-run CI.

Issue #327, opened 2018-03-07
CosmosDB: Support providing consistency policy


jeremyrickard:
CosmosDB provides the ability to select the consistency for the database account. We should add that to the CosmosDB instances provisioned by OSBA.

PR #380, opened 2018-04-06
Adding ability to specify cosmos consistency policy

jeremyrickard:
This PR adds the ability to specify the consistency policy
for a CosmosDB database account. It also adds a new type
to the schemas code, a numeric type with an upper and lower range.

Closes: #327

### DIFF
--- a/docs/modules/cosmosdb.md
+++ b/docs/modules/cosmosdb.md
@@ -141,8 +141,9 @@ Provisions a new CosmosDB database account that can be accessed through any of t
 | `ipFilters.allowedIPRanges` | `array` | Values to include in IP Filter. Can be IP Address or CIDR range. | N | If not specified, no additional values will be included in filters. |
 | `consistencyPolicy` | `object` | The consistency policy for the Cosmos DB account. | N | |
 | `consistencyPolicy.defaultConsistencyLevel` | `string` | The default consistency level and configuration settings of the Cosmos DB account. - Eventual, Session, BoundedStaleness, Strong, ConsistentPrefix | Y | |
-| `consistencyPolicy.maxStalenessPrefix` | `integer` | When used with the Bounded Staleness consistency level, this value represents the number of stale requests tolerated. Accepted range for this value is 1 – 2,147,483,647. Required when defaultConsistencyPolicy is set to 'BoundedStaleness'. | N | |
-| `consistencyPolicy.maxIntervalInSeconds` | `integer` | When used with the Bounded Staleness consistency level, this value represents the time amount of staleness (in seconds) tolerated. Accepted range for this value is 5 - 86400. Required when defaultConsistencyPolicy is set to 'BoundedStaleness'. | N | |
+| `consistencyPolicy.boundedStaleness` | object | Settings for to determine staleness when used with `BoundedStaleness` consistency | Yes - If using `BoundedStaleness` consistency | | 
+| `consistencyPolicy.boundedStaleness.maxStalenessPrefix` | `integer` | When used with the Bounded Staleness consistency level, this value represents the number of stale requests tolerated. Accepted range for this value is 1 – 2,147,483,647. Required when defaultConsistencyPolicy is set to 'BoundedStaleness'. | N | |
+| `consistencyPolicy.boundedStaleness.maxIntervalInSeconds` | `integer` | When used with the Bounded Staleness consistency level, this value represents the time amount of staleness (in seconds) tolerated. Accepted range for this value is 5 - 86400. Required when defaultConsistencyPolicy is set to 'BoundedStaleness'. | N | |
 
 ##### Bind
 
--- a/docs/modules/cosmosdb.md
+++ b/docs/modules/cosmosdb.md
@@ -2,7 +2,6 @@
 
 _Note: This module is EXPERIMENTAL and future releases may break the API._
 
-
 ## Services & Plans
 
 ### Service: azure-cosmosdb-sql-account
--- a/docs/modules/cosmosdb.md
+++ b/docs/modules/cosmosdb.md
@@ -57,7 +57,7 @@ Does nothing.
   
 ##### Deprovision
 
-Deletes the CosmosDB database.
+Deletes the CosmosDB database account.
 
 ### Service: azure-cosmosdb-mongo-account
 
@@ -114,7 +114,7 @@ Does nothing.
 
 ##### Deprovision
 
-Deletes the CosmosDB database.
+Deletes the CosmosDB database account.
 
 ### Service: azure-cosmosdb-graph-account
 
@@ -169,7 +169,7 @@ Does nothing.
   
 ##### Deprovision
 
-Deletes the CosmosDB database.
+Deletes the CosmosDB database account.
 
 ### Service: azure-cosmosdb-table-account
 
@@ -223,4 +223,4 @@ Does nothing.
   
 ##### Deprovision
 
-Deletes the CosmosDB database.
+Deletes the CosmosDB database account.
--- a/docs/modules/cosmosdb.md
+++ b/docs/modules/cosmosdb.md
@@ -27,6 +27,10 @@ Provisions a new CosmosDB database account that can be accessed through any of t
 | `ipFilters.allowAccessFromAzure` | `string` | Specifies if Azure Services should be able to access the CosmosDB account.Valid valued are `""` (unspecified), `enabled`, or `disabled`. | N | If left unspecified, defaults to enabled. |
 | `ipFilters.allowAccessFromPortal` | `string` | Specifies if the Azure Portal should be able to access the CosmosDB account. If `allowAccessFromAzure` is set to enabled, this value is ignored. Valid valued are `""` (unspecified), `enabled`, or `disabled`. | N | If left unspecified, defaults to enabled. |
 | `ipFilters.allowedIPRanges` | `array` | Values to include in IP Filter. Can be IP Address or CIDR range. | N | If not specified, no additional values will be included in filters. |
+| `consistencyPolicy` | `object` | The consistency policy for the Cosmos DB account. | N | |
+| `consistencyPolicy.defaultConsistencyLevel` | `string` | The default consistency level and configuration settings of the Cosmos DB account. - Eventual, Session, BoundedStaleness, Strong, ConsistentPrefix | Y | |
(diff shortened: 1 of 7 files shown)


jeremyrickard:
This lets you specify a CosmosDB consistency policy. 
That looks something like:

```
"consistencyPolicy": {
	"defaultConsistencyLevel": "BoundedStaleness",
	"maxStalenessPrefix":      10,
	"maxIntervalInSeconds":    500,
},
```

This also adds the associated schema to support that. In relation to that, I've added a "Numeric" parameter schema that has an upper and lower bound (doesn't apply to other types). 

New cosmosDB schema looks like:

```
{
                                    "$schema": "http://json-schema.org/draft-04/schema#",
                                    "type": "object",
                                    "properties": {
                                        "consistencyPolicy": {
                                            "type": "object",
                                            "description": "The consistency policy for the Cosmos DB account.",
                                            "required": [
                                                "defaultConsistencyLevel"
                                            ],
                                            "properties": {
                                                "defaultConsistencyLevel": {
                                                    "type": "string",
                                                    "description": "The default consistency level and configuration settings of the Cosmos DB account.",
                                                    "enum": [
                                                        "Eventual",
                                                        "Session",
                                                        "BoundedStaleness",
                                                        "Strong",
                                                        "ConsistentPrefix"
                                                    ]
                                                },
                                                "maxIntervalInSeconds": {
                                                    "type": "integer",
                                                    "description": "When used with the Bounded Staleness consistency level, this value represents the time amount of staleness (in seconds) tolerated. Required when defaultConsistencyPolicy is set to  'BoundedStaleness'.",
                                                    "minimum": 1,
                                                    "maximum": 86400
                                                },
                                                "maxStalenessPrefix": {
                                                    "type": "integer",
                                                    "description": "When used with the Bounded Staleness consistency level, this value represents the number of stale requests toleratedRequired when defaultConsistencyPolicy is set to  'BoundedStaleness'.",
                                                    "minimum": 1,
                                                    "maximum": 2147483647
                                                }
                                            }
                                        },
                                        "ipFilters": {
                                            "type": "object",
                                            "description": "IP Range Filter to be applied to new CosmosDB account",
                                            "properties": {
                                                "allowAccessFromAzure": {
                                                    "type": "string",
                                                    "description": "Specifies if Azure Services should be able to access the CosmosDB account.",
                                                    "default": "",
                                                    "enum": [
                                                        "",
                                                        "enabled",
                                                        "disabled"
                                                    ]
                                                },
                                                "allowAccessFromPortal": {
                                                    "type": "string",
                                                    "description": "Specifies if the Azure Portal should be able to access the CosmosDB account. If `allowAccessFromAzure` is set to enabled, this value is ignored.",
                                                    "default": "",
                                                    "enum": [
                                                        "",
                                                        "enabled",
                                                        "disabled"
                                                    ]
                                                },
                                                "allowedIPRanges": {
                                                    "type": "array",
                                                    "description": "Values to include in IP Filter. Can be an IP Address or CIDR range.",
                                                    "items": {
                                                        "type": "string",
                                                        "description": "Must be a valid IP address or CIDR"
                                                    }
                                                }
                                            }
                                        },
                                        "location": {
                                            "type": "string",
                                            "description": "The Azure region in which to provision applicable resources."
                                        },
                                        "resourceGroup": {
                                            "type": "string",
                                            "description": "The (new or existing) resource group with which to associate new resources."
                                        },
                                        "tags": {
                                            "type": "object",
                                            "description": "Tags to be applied to new resources, specified as key/value pairs.",
                                            "additionalProperties": {
                                                "type": "string"
                                            }
                                        }
                                    }
                                }
```

pkg/service/schema.go
```diff
@@ -100,6 +100,19 @@ func (a *ArrayParameterSchema) MarshalJSON() ([]byte, error) {
...
+// NumericParameterSchema represents a numeric type, either
+// integers or floating point numbers, that can have an upper or lower bound.
+type NumericParameterSchema struct {
+	Type             string      `json:"type"`
+	Description      string      `json:"description,omitempty"`
+	Required         bool        `json:"-"`
+	Default          interface{} `json:"default,omitempty"`
+	Minimum          interface{} `json:"minimum,omitempty"`
+	ExclusiveMinimum bool        `json:"exclusiveMinimum,omitempty"`
+	Maximum          interface{} `json:"maximum,omitempty"`
```
jeremyrickard: These "Exlusive" properties aren't being used at the moment but are in the spec, so I included them.

pkg/services/cosmosdb/common-types.go
```diff
@@ -89,6 +96,49 @@ func (
...
+				AllowedValues: []string{
+					"Eventual",
+					"Session",
+					"BoundedStaleness",
+					"Strong",
+					"ConsistentPrefix",
+				},
+			},
+			"maxStalenessPrefix": &service.NumericParameterSchema{
+				Type: "integer",
```
jeremyrickard: Here is the new type

docs/modules/cosmosdb.md
```diff
@@ -27,6 +27,10 @@ Provisions a new CosmosDB database account that can be accessed through any of t
...
 | `ipFilters.allowAccessFromPortal` | `string` | Specifies if the Azure Portal should be able to access the CosmosDB account. If `allowAccessFromAzure` is set to enabled, this value is ignored. Valid valued are `""` (unspecified), `enabled`, or `disabled`. | N | If left unspecified, defaults to enabled. |
 | `ipFilters.allowedIPRanges` | `array` | Values to include in IP Filter. Can be IP Address or CIDR range. | N | If not specified, no additional values will be included in filters. |
+| `consistencyPolicy` | `object` | The consistency policy for the Cosmos DB account. | N | |
+| `consistencyPolicy.defaultConsistencyLevel` | `string` | The default consistency level and configuration settings of the Cosmos DB account. - Eventual, Session, BoundedStaleness, Strong, ConsistentPrefix | Y | |
+| `consistencyPolicy.maxStalenessPrefix` | `integer` | When used with the Bounded Staleness consistency level, this value represents the number of stale requests tolerated. Accepted range for this value is 1 – 2,147,483,647. Required when defaultConsistencyPolicy is set to 'BoundedStaleness'. | N | |
```
krancour: Putting an idea out there...

For many things I have worked on, if an option is subordinate in some way to some other option, I try to make that scope clear through the keys...

For example, in many k8s charts, you can override `service.type`. Often if type `NodePort` is selected, other options become available (like choosing a specific port), but they'd applicable _only_ when `service.type: NodePort`. e.x.:

```
service:
  ## Type of service; valid values are "ClusterIP", "LoadBalancer", and
  ## "NodePort". "ClusterIP" is sufficient in the average case where a service
  ## catalog installation in the same cluster is the only client that needs to
  ## communicate with this broker. i.e. The broker does not need to receive
  ## inbound requests from outside the cluster.
  type: ClusterIP
  ## Further configuration if service is of type "NodePort"
  nodePort:
    ## Available port in allowable range (e.g. 30000 - 32767 on minikube)
    port: 30080
```

I'm wondering if the same sort of strategy would add some clarity here. What if `consistencyPolicy. maxStalenessPrefix ` were changed to `consistencyPolicy.boundedStateless.maxStalenessPrefix`. Would that make it more clear that the option is applicable only when `consistencyPolicy.defaultConsistencyLevel: BoundedStaleness`?

I'm not strongly encouraging this. I'm only opening this up to discussion. Do not count this as a blocker by any means.

pkg/services/cosmosdb/common-provision.go
```diff
@@ -83,7 +83,70 @@ func (c *cosmosAccountManager) ValidateProvisioningParameters(
...
+		cp := strings.ToLower(pp.ConsistencyPolicy.DefaultConsistency)
+		// treat this map like a set and use lookup to check the
+		// validuty of the given consistency policy
+		validValues := map[string]struct{}{
+			"eventual":         {},
+			"session":          {},
+			"boundedstaleness": {},
+			"strong":           {},
+			"consistentprefix": {},
+		}
```
krancour: Good use for map[string]struct{}-- super efficient. Want to propose that this can be a _little_ bit moreso by defining the `validValues` at the `cosmosAccountManager ` level or even the package level so that we don't need to re-allocate memory for this static info on every invocation of this function.

Still not a blocker tho, imo.

pkg/services/cosmosdb/common-provision.go
```diff
@@ -147,6 +210,35 @@ func (c *cosmosAccountManager) buildGoTemplateParams(
...
+		case "strong":
+			consistencyPolicy["defaultConsistencyLevel"] = "Strong"
+		case "consistentprefix":
+			consistencyPolicy["defaultConsistencyLevel"] = "ConsistentPrefix"
+		}
+
+		if pp.ConsistencyPolicy.MaxInternal != nil {
+			consistencyPolicy["maxIntervalInSeconds"] =
+				*pp.ConsistencyPolicy.MaxInternal
+		}
```
krancour: Should there we a guard on this stanza to only set this if the policy is `BoundedStaleness`? Same question for the next stanza.

jeremyrickard: Yeah, I misread the documentation. 

```

File: docs/modules/cosmosdb.md (hunk 3 above)
Comment (jeremyrickard) [reply docs/modules/cosmosdb.md]: Yeah I think so. See my other comment, I mis-read the docs before!

File: docs/modules/cosmosdb.md [hunk 6]
```diff
@@ -29,8 +29,9 @@ Provisions a new CosmosDB database account that can be accessed through any of t
...
-| `consistencyPolicy.maxIntervalInSeconds` | `integer` | When used with the Bounded Staleness consistency level, this value represents the time amount of staleness (in seconds) tolerated. Accepted range for this value is 5 - 86400. Required when defaultConsistencyPolicy is set to 'BoundedStaleness'. | N | |
+| `consistencyPolicy.boundedStaleness` | `object` | Specifies the settings when using BoundedStaleness consistency. | Y - When Using `BoundedStaleness` | |
+| `consistencyPolicy.maxStalenessPrefix` | `integer` | When used with the Bounded Staleness consistency level, this value represents the number of stale requests tolerated. Accepted range for this value is 1 – 2,147,483,647. Required when defaultConsistencyPolicy is set to 'BoundedStaleness'. | Y | |
+| `consistencyPolicy.maxIntervalInSeconds` | `integer` | When used with the Bounded Staleness consistency level, this value represents the time amount of staleness (in seconds) tolerated. Accepted range for this value is 5 - 86400. Required when defaultConsistencyPolicy is set to 'BoundedStaleness'. | Y | |
```
krancour: I think those last two fields were supposed to have changed to `consistencyPolicy. boundedStaleness.<field>`, no?

krancour: Looks like that's how it is in the code. Docs just need to match it.

jeremyrickard: Yeah good catch!

PR #383, opened 2018-04-06
fix chart release; also publish chart release candidates

krancour:
We already treat the head of master as always (potentially) releasable. The head of master is always, therefore, a release candidate. We already publish "rc" images (tagged with git sha) as commits exit the master pipeline. I think it best if we _also_ publish an "rc" chart that goes along with it. This PR accomplishes that.

This also fixes a problem where the actual chart release process (release pipeline) wasn't getting the correct image tag set in `values.yaml` prior to publishing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -317,18 +317,42 @@ else
 	$(DOCKER_HELM_CMD) $(LINT_CHART_CMD)
 endif
 
-PUBLISH_CHART_CMD := bash -c ' \
+PUBLISH_RC_CHART_CMD := bash -c ' \
 	cd contrib/k8s/charts \
 	&& rm -rf repo \
 	&& mkdir repo \
 	&& cd repo \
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -317,18 +317,42 @@ else
...
 endif
 
-PUBLISH_CHART_CMD := bash -c ' \
+PUBLISH_RC_CHART_CMD := bash -c ' \
 	cd contrib/k8s/charts \
 	&& rm -rf repo \
 	&& mkdir repo \
 	&& cd repo \
-	&& sed -i s/0.0.1/$(REL_VERSION)/g ../open-service-broker-azure/Chart.yaml \
+	&& sed -i "s/^version:.*/version: 0.0.1+$(GIT_VERSION)/g" ../open-service-broker-azure/Chart.yaml \
```
krancour: Helm requires versions to be a legitimate semver string, so we'll assign special meaning to 0.0.1, because we can't get away with using a value like "rc" or "canary". (I tried.)

PR #384, opened 2018-04-06
Jenkinsfile

krancour:
Do not merge this (yet).

I'm opening this PR here so that I can just point a (for now, private) Jenkins instance at this and test out the Jenkinsfile.


krancour:
I'm closing this with an assumption that if we move off of Circle, we're more likely to move to Azure Pipelines than to Jenkins.

PR #385, opened 2018-04-08
Rename PCF tile and its label

zhongyi-zhang:
According to Ning and Pivotal's discussion, we need to rename the tile to **Azure Open Service Broker for PCF** to distinguish from a PKS tile. And we keep using **open-service-broker-azure** as the broker app name.

CC: @arschles


zhongyi-zhang:
Please hold on this PR for now. Looks it is still in discussion.

zhongyi-zhang:
Hi @krancour, could you help to merge? Pivotal confirmed that they accept _Azure Open Service Broker for PCF_ to be the tile name.

PR #386, opened 2018-04-09
Fix bug in cosmos common provisioning

jeremyrickard:
There was an error in cosmosdb common provisioning that resulted in
a nil pointer derefence.

### DIFF
--- a/pkg/services/cosmosdb/common-provision.go
+++ b/pkg/services/cosmosdb/common-provision.go
@@ -103,9 +103,10 @@ func (c *cosmosAccountManager) buildGoTemplateParams(
 	p["kind"] = kind
 
 	filters := []string{}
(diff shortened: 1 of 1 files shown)


pkg/services/cosmosdb/common-provision.go
```diff
@@ -103,9 +103,10 @@ func (c *cosmosAccountManager) buildGoTemplateParams(
 	p["kind"] = kind
 
 	filters := []string{}
-	allowAzure := strings.ToLower(pp.IPFilterRules.AllowPortal)
-	allowPortal := strings.ToLower(pp.IPFilterRules.AllowPortal)
```
jeremyrickard: womp womp, not sure how this ended up outside the nil check.

PR #388, opened 2018-04-09
fix default environment in chart

krancour:
```
$ k logs -f osba-open-service-broker-azure-6d966487d-8qw6p -n osba
time="2018-04-09T19:36:00Z" level=info msg="Open Service Broker for Azure starting" commit=8ed9bff version=devel
time="2018-04-09T19:36:00Z" level=info msg="Setting log level" logLevel=INFO
time="2018-04-09T19:36:00Z" level=fatal msg="autorest/azure: There is no cloud environment matching the name \"AZURECLOUD\""
```

The default environment set in the chart is wrong-- does not exist. Not sure how we got away with this for so long...


krancour:
I am merging this simple bug fix in good faith because of a narrowing window of opportunity to complete the v0.10.0 release.

PR #390, opened 2018-04-10
[WIP] Adding PostgreSQL standard100 plan

arschles:
This patch adds support for `standard100` plans to Azure PostgreSQL. Marked WIP because I am still confirming the SKU properties

### DIFF
--- a/pkg/services/postgresql/catalog.go
+++ b/pkg/services/postgresql/catalog.go
@@ -53,6 +53,17 @@ func (m *module) GetCatalog() (service.Catalog, error) {
 					Bullets:     []string{"100 DTUs"},
 				},
 			}),
+			service.NewPlan(&service.PlanProperties{
+				ID:          "bd588e32-0514-4421-8ef3-f54039914e61",
+				Name:        "standard100",
+				Description: "Standard Tier, 100 DTUs",
+				Free:        false,
+				Extended: map[string]interface{}{
+					"skuName":        "PGSQLS100",
+					"skuTier":        "Standard",
+					"skuCapacityDTU": 100,
+				},
+			}),
 		),
 		// dbms only
 		service.NewService(
--- a/pkg/services/postgresql/catalog.go
+++ b/pkg/services/postgresql/catalog.go
@@ -103,6 +103,17 @@ func (m *module) GetCatalog() (service.Catalog, error) {
 					Bullets:     []string{"100 DTUs"},
 				},
 			}),
+			service.NewPlan(&service.PlanProperties{
+				ID:          "ba653f35-2152-4d76-8641-b21d4636b2e1",
+				Name:        "standard100",
+				Description: "Standard Tier, 100 DTUs",
+				Free:        false,
+				Extended: map[string]interface{}{
+					"skuName":        "PGSQLS100",
+					"skuTier":        "Standard",
+					"skuCapacityDTU": 100,
+				},
+			}),
 		),
 		// database only
 		service.NewService(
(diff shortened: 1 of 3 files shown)


pkg/services/postgresql/catalog.go
```diff
@@ -103,6 +103,17 @@ func (m *module) GetCatalog() (service.Catalog, error) {
...
 			}),
+			service.NewPlan(&service.PlanProperties{
+				ID:          "ba653f35-2152-4d76-8641-b21d4636b2e1",
+				Name:        "standard100",
+				Description: "Standard Tier, 100 DTUs",
+				Free:        false,
+				Extended: map[string]interface{}{
+					"skuName":        "PGSQLS100",
+					"skuTier":        "Standard",
+					"skuCapacityDTU": 100,
```
arschles: note I am still confirming these are the right `sku{Name,Tier,CapacityDTU}`

tests/lifecycle/postgresql_cases_test.go
```diff
@@ -15,7 +15,7 @@ var postgresqlTestCases = []serviceLifecycleTestCase{
 		group:           "postgresql",
 		name:            "all-in-one",
 		serviceID:       "b43b4bba-5741-4d98-a10b-17dc5cee0175",
-		planID:          "b2ed210f-6a10-4593-a6c4-964e6b6fad62",
+		planID:          "ba653f35-2152-4d76-8641-b21d4636b2e1",
```
jeremyrickard: 👍 

I don't see a point in adding a new one right now. When we re-do these, we can add another test for different plans.

pkg/services/postgresql/catalog.go
jeremyrickard: You will need to update the ARM template as well. We constrain valid values in the template right now.

krancour:
This looks good, but I notice this just adds this sku as a plan under the DBMS-only service. I think we should also add a plan using this sku under the "all-in-one" service (DBMS + DB).

pkg/services/postgresql/catalog.go
```diff
@@ -103,6 +103,17 @@ func (m *module) GetCatalog() (service.Catalog, error) {
 					Bullets:     []string{"100 DTUs"},
 				},
 			}),
+			service.NewPlan(&service.PlanProperties{
```
jeremyrickard: You also need to add this to the dbms only service. Not sure if you were waiting for validation on the comment below.

pkg/services/postgresql/catalog.go
arschles: got it, thanks @jeremyrickard - commit incoming

pkg/services/postgresql/catalog.go
krancour: Nope... this _is_ already under DBMS-only. It needs to be added under all-in-one. See [other comment](https://github.com/Azure/open-service-broker-azure/pull/390#issuecomment-380196937)

jeremyrickard: yeah flipped that in my brain :)

arschles: :+1: added to all-in-one in https://github.com/Azure/open-service-broker-azure/pull/390/commits/2bf20588332fe3bb60c0dd105742fca712032697

pkg/services/postgresql/catalog.go
arschles: done in https://github.com/Azure/open-service-broker-azure/pull/390/commits/284e4b7c6e850c9918cad7ce3bcd30e6418b8109

krancour:
@arschles 

```
--- FAIL: TestServices (1250.71s)
    --- FAIL: TestServices/lifecycle (0.00s)
        --- FAIL: TestServices/lifecycle/TestServices/lifecycle/postgresql/all-in-one (0.00s)
	Error Trace:	driver_test.go:46
	Error:      	Expected nil, but got: &errors.errorString{s:"plan \"ba653f35-2152-4d76-8641-b21d4636b2e1\" not found for service \"b43b4bba-5741-4d98-a10b-17dc5cee0175\""}
```

jeremyrickard:
👍

krancour:
Thanks, @arschles!

(He still codes, ladies and gentlemen!)

PR #394, opened 2018-04-16
Added some words about stability and the helm chart

### LINKED ISSUE #393: Update Quickstart to mention the min stability flag.
Now that we've updated the helm chart to install the broker in preview mode, we will hide the experimental services. To help people use them, we should update the quickstart guides with some of that info.

jeremyrickard:
We've recently updated the helm chart to set the broker's minimum stability level
to preview. This means that any of the experimental moduels are not going to be included
in the catalog. If people want to use them, they will need to use that flag. This adds it
to the two quick start guides and the FAQ.

Closes #393

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -23,3 +23,19 @@ time="2018-03-13T12:08:31Z" level=fatal msg="async engine stopped: error sending
 After checking on the health of the kube-system pods, kube-dns was not working. This caused osba 
 to be unable to connect to its Redis instance. The problem was resolved by deleting the
 kube-dns pods, so that Kubernetes would recreate them.
+
+## I don't see all the services
+
+The helm chart is configured to set the minimum stability level of the broker to `preview`,
+meaning that only services that are marked `preview` or `stable` will be available from the broker.
+This is currently Azure Database for MySQL, Azure Database for PostgreSQL and Azure SQL Database.
+If you would like to use other services, you will need to add an additional flag to your helm install command:
+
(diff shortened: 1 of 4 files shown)


docs/faq.md
```diff
@@ -22,4 +22,20 @@ time="2018-03-13T12:08:31Z" level=fatal msg="async engine stopped: error sending
 
 After checking on the health of the kube-system pods, kube-dns was not working. This caused osba 
 to be unable to connect to its Redis instance. The problem was resolved by deleting the
-kube-dns pods, so that Kubernetes would recreate them.
\ No newline at end of file
+kube-dns pods, so that Kubernetes would recreate them.
+
+## I don't see all the services
+
+The helm chart is configured to set the minimum stability level of the broker to `preview`,
```
krancour: Can we lead into this with a brief one or two sentence explanation (in case it isn't well-understood by the reader) that the broker's services are provided by a series of modules and those modules all have their own level of stability? It's hard to put myself in the shoes of someone who doesn't already know OSBA inside and out, but I think one could infer all that from this paragraph if you're good at reading between the lines, but we should be explicit about it.

This bit could throw people off:

> the minimum stability level of the broker

Since it sounds like it refers to the broker itself, rather than constituent pieces of it.

krancour: Same applies in other places where this same text appears, obviously.

jeremyrickard:
No CI required, this is documentation only.

PR #395, opened 2018-04-17
Add "uri" to the MongoDB credential.

### LINKED ISSUE #282: Enhance MongoDB API CosmosDB Service Binding
We should enhance the CosmosDB binding information in line with #275. See https://cloud.spring.io/spring-cloud-connectors/spring-cloud-cloud-foundry-connector.html and https://docs.cloudfoundry.org/services/binding-credentials.html. Also see the azure lib: https://github.com/Microsoft/azure-spring-boot/blob/master/azure-spring-boot/src/main/java/com/microsoft/azure/spring/cloundfoundry/environment/VcapResult.java

The MongoDB API service should return the uri string starting with mongodb://<user>:<pass>@<host>:<port>

jeremyrickard:
This PR adds a duplicate entry to the binding credential with the
"uri" property. This is the same value as the "connectionString" property
but will add better support for the Spring Cloud Foundry connector for
MongoDB. By default, it will look for a "uri" field. This should enable
applications to consume our MongoDB API service offering with no special
handling on CF.

Closes #282

### DIFF
--- a/pkg/services/cosmosdb/mongodb-types.go
+++ b/pkg/services/cosmosdb/mongodb-types.go
@@ -3,10 +3,10 @@ package cosmosdb
 // mongoCredentials encapsulates CosmosDB-specific connection details and
 // credentials for connecting with the MongoDB API.
 type mongoCredentials struct {
-	Host             string `json:"host,omitempty"`
-	Port             int    `json:"port,omitempty"`
-	Username         string `json:"username,omitempty"`
-	Password         string `json:"password,omitempty"`
-	ConnectionString string `json:"connectionString,omitempty"`
-	URI              string `json:"uri,omitempty"`
+	Host             string `json:"host"`
+	Port             int    `json:"port"`
+	Username         string `json:"username"`
+	Password         string `json:"password"`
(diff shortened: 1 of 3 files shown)


pkg/services/cosmosdb/mongodb-types.go
```diff
@@ -8,4 +8,5 @@ type mongoCredentials struct {
 	Username         string `json:"username,omitempty"`
 	Password         string `json:"password,omitempty"`
 	ConnectionString string `json:"connectionString,omitempty"`
+	URI              string `json:"uri,omitempty"`
```
krancour: Let's drop `omitempty` here. We should also drop it on the lines above. I know why it was there...

Back before you split this into multiple service managers, there was a single `GetCredentials(....)` function that had to return credentials of a few different "shapes." That isn't a concern anymore and we can reduce cognitive load on contributors by leaving off `omitempty` on field we know will always be set to non-zero values.

jeremyrickard: I went ahead and cleaned all of these up, both on the mongo and the cosmos generic one. With the specific types, all the things will always be set.

Issue #283, opened 2018-02-17
Enhance Redis Binding


jeremyrickard:
In line with #275, enhance the Redis binding to support better use on CF. https://cloud.spring.io/spring-cloud-connectors/spring-cloud-cloud-foundry-connector.html#_redis

PR #396, opened 2018-04-17
Adding to the Redis binding

jeremyrickard:
Adding to the Redis binding to allow it to work out of the box with the Spring
Cloud Connectors. By default, they look for a thing in the credentials called "uri"
that has the form redis://<username>:<password>@<host>:<port>

We were already exposing this data but this makes it easier.

Tested with https://github.com/jeremyrickard/redis-test-app/

Closes: #283

### DIFF
--- a/pkg/services/rediscache/bind.go
+++ b/pkg/services/rediscache/bind.go
@@ -37,14 +37,13 @@ func (s *serviceManager) GetCredentials(
 		return nil, err
 	}
 
-	uriStringTemplate := "redis://:%s@%s:%d"
 	redisPort := 6379
 	return credentials{
 		Host:     dt.FullyQualifiedDomainName,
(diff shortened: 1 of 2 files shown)


pkg/services/rediscache/bind.go
```diff
@@ -33,9 +36,18 @@ func (s *serviceManager) GetCredentials(
 	if err := service.GetStructFromMap(instance.SecureDetails, &sdt); err != nil {
 		return nil, err
 	}
+
+	uriStringTemplate := "redis://:%s@%s:%d"
```
krancour: Let's put this in-line in the one place it's used. If this was copy/paste from another service, it was done like this in one or two places because of conditional tweaks to the template, but nothing like that is applicable here, so imo, this be more readable if it's in-line.

Issue #399, opened 2018-04-18
CosmosDB SQL API Service Actually Provisions a Table API


jeremyrickard:
They CosmosDB SQL API service is actually provisioning a Table API service. We should fix this ASAP.

PR #400, opened 2018-04-18
High priority fix for sql api service

jeremyrickard:
When we completed the changes to cosmosdb, the sql api
got tagged with the Table API experience, so it was marked as
table in the portal.

Fixes: #399


pkg/services/cosmosdb/sql-api-provision.go
```diff
@@ -33,11 +33,9 @@ func (s *sqlAccountManager) deployARMTemplate(
 	}
 
 	p := s.buildGoTemplateParams(pp, dt, "GlobalDocumentDB")
-	p["capability"] = "EnableTable"
 	if instance.Tags == nil {
```
krancour: Instead of just removing this, is there some other value that should be set instead? tbh, I don't understand the who "capability" aspect of Cosmos like you do.

PR #405, opened 2018-04-23
add ability to EOL services and plans

### LINKED ISSUE #403: Internally track service / plan availability
The current proposal in OSB API for deprecating services and plans doesn't involve any admonishment to platforms that they must not attempt to provision instances of deprecated plans. Initially, I objected to the proposal on the grounds that this would make it difficult for a broker to continue deprovisioning instances of plans that it doesn't wish to (or cannot) provision more of.

It's since been made clear to me that some platforms (hopefully all) do not require an instance's service and plan to _currently_ be in the catalog in order to initiate a deprovisioning request. What this means for OSBA, practically speaking, is that when we are no longer able to provision some plan (imagine, for instance that Azure DB for PostgreSQL were to phase out Postgres 9.5) we simply need to EOL that plan (hopefully having previously deprecated it to give fair warning), _but_ we will need to retain the capability of _deprovisioning_ existing instances the plan in question.

To accommodate this, within our own catalog we need to keep track of which services and/or plans are EOL. Such services and plans should be excluded when serving the catalog to platforms via the `/v2/catalog` endpoint.

krancour:
Fixes #403

### DIFF
--- a/pkg/api/provision.go
+++ b/pkg/api/provision.go
@@ -102,7 +102,7 @@ func (s *server) provision(w http.ResponseWriter, r *http.Request) {
 	}
 
 	svc, ok := s.catalog.GetService(serviceID)
-	if !ok {
+	if !ok || svc.IsEndOfLife() {
 		logFields["serviceID"] = serviceID
 		log.WithFields(logFields).Debug(
 			"bad provisioning request: invalid serviceID",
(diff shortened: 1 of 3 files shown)


pkg/api/provision.go
```diff
@@ -102,7 +102,7 @@ func (s *server) provision(w http.ResponseWriter, r *http.Request) {
 	}
 
 	svc, ok := s.catalog.GetService(serviceID)
-	if !ok {
+	if !ok || svc.IsEndOfLife() {
```
jeremyrickard: This is 👍

PR #407, opened 2018-04-23
qualify service and plan versions in terms of api compatibility

### LINKED ISSUE #402: PostgreSQL and MySQL plans should be specific about what version they provide
Using Postgres as an example, when you provision it through the Azure portal today, you can choose from 9.5 or 9.6. (OSBA, currently always provisions 9.6.)

I believe plans should be _deterministic_ and _transparent_ about what they provide. Provisioning the "latest" PostgreSQL would be bad, because what you get today and what you get tomorrow could be different. Not being clear with the end-user about what version you're giving them would also be bad.

I've thought through how we can be more specific on these fronts. There is a strong impulse to say that Postgres version should be a parameter on our Postgres plans. While that seems easy and intuitive, I believe it's wrong for the following reasons:

1. I anticipate that Azure Database for PostgreSQL will add versions over time and retire older versions as well. _If version were expressed as a parameter_ then the acceptable values for that parameter will change over time. Adding acceptable values may be a non-breaking change, but _removing_ acceptable values _is_ a breaking change. So we won't want to do that.

2. Once again, acknowledging that Azure Database for PostgreSQL will likely add versions over time and retire older versions, the day will come where a given version simply _cannot_ be provisioned. The OSB API is currently adding an ability for brokers to deprecate plans. We'd like to be able to take advantage of that feature.

3. Looking to the future, our optimal view of the future involves one where end users don't request services and plans from their platforms via IDs-- rather, we envision a future where every plan uses semantically meaningful metadata to describe its capabilities and users specify constraints (think k8s label selectors) to describe required capabilities. Their platform will then connect them with an appropriate plan. In light of this, it's important to acknowledge that a plan that provides Postgres 9.5 and plan that provides Postgres 9.6 _are not logically the same thing._

What I propose for now is that we offer _only_ the latest versions of PostgreSQL and MySQL, but we change plan names, descriptions, metadata, etc. to be clear about what specific version they provide. We should be prepared to add sibling plans in the future that offer newer versions and we should be prepared to deprecate and EOL plans as Azure Database for PostgreSQL deprecates and retires support for the underlying versions.

krancour:
Fixes #402

### DIFF
--- a/pkg/service/catalog.go
+++ b/pkg/service/catalog.go
@@ -37,6 +37,7 @@ type ServiceProperties struct { // nolint: golint
 	ProvisionParamsSchema map[string]ParameterSchema `json:"-"`
 	UpdateParamsSchema    map[string]ParameterSchema `json:"-"`
 	BindingParamsSchema   map[string]ParameterSchema `json:"-"`
+	Extended              map[string]interface{}     `json:"-"`
 }
 
 // ServiceMetadata contains metadata about the service classes
@@ -61,6 +62,7 @@ type Service interface {
 	GetPlan(planID string) (Plan, bool)
 	GetParentServiceID() string
 	GetChildServiceID() string
+	GetProperties() *ServiceProperties
 }
 
 type service struct {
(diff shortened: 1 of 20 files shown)


jeremyrickard:
@krancour We should open an issue in the helm-charts repo to update all the charts that use mysql as a result of this. The classes will be incompatible now, so it would be good to have a tracking issue to make sure we get them cleaned up when the release happens.

PR #416, opened 2018-04-26
Refactor mysql postgresql plans!

### LINKED ISSUE #409: Update MySQL and PostgreSQL Plans
MySQL and PostgreSQL have new billing models that allow more fine grained control of things like storage and have introduced a new concept called vCores instead of DTUs. We should update the plans to reflect this and keep them in line with what a user would see in the Azure portal.

jeremyrickard:
This PR updates the mysql and postgreQSL modules to reflect the new Azure Database for MyQSL and Azure Database for PostgreSQL business plans. 

It also introduces necessary changes to the core part of the OSBA framework to allow plans to have individual provision/update/bind schemas and to do validation on a plan by plan level. 

Updates documentation to cover the earlier PR that introduced version into the service names

Closes #409

### DIFF
--- a/contrib/k8s/examples/mysql/advanced/mysql-database-instance.yaml
+++ b/contrib/k8s/examples/mysql/advanced/mysql-database-instance.yaml
@@ -4,7 +4,7 @@ metadata:
   name: example-mysql-database-instance
   namespace: default
 spec:
-  clusterServiceClassExternalName: azure-mysql-database
+  clusterServiceClassExternalName: azure-mysql-5-7-database
   clusterServicePlanExternalName: database
   parameters:
     parentAlias: 71b6b6c4-5271-47d3-8188-51e69fba914e
--- a/docs/modules/mysql.md
+++ b/docs/modules/mysql.md
@@ -6,20 +6,21 @@ Open Service Broker for Azure contains three Azure Database for MySQL services. 
 
 | Service Name | Description |
 |--------------|-------------|
-| `azure-mysql` | Provision both an Azure Database for MySQL Database Management System (DBMS) and a database. |
-| `azure-mysql-dbms` | Provision only an Azure Database for MySQL DBMS. This can be used to provision multiple databases at a later time. |
-| `azure-mysql-database` | Provision a new database only upon a previously provisioned DBMS. |
+| `azure-mysql-5-7` | Provision both an Azure Database for MySQL Database Management System (DBMS) and a database, using MySQL 5.7 |
+| `azure-mysql-5-7-dbms` | Provision only an Azure Database for MySQL DBMS with MySQL 5.7. This can be used to provision multiple databases at a later time. |
+| `azure-mysql-5-7-database` | Provision a new database only upon a previously provisioned DBMS. |
 
-The `azure-mysql` service allows you to provision both a DBMS and a database. When the provision operation is successful, the database will be ready to use. You can't provision additional databases onto an instance provisioned through this service. The `azure-mysql-dbms` and `azure-mysql-database` services, on the other hand, can be combined to provision multiple databases on a single DBMS.  For more information on each service, refer to the descriptions below.
+The `azure-mysql-5-7l` service allows you to provision both a DBMS and a database. When the provision operation is successful, the database will be ready to use. You can't provision additional databases onto an instance provisioned through this service. The `azure-mysql-5-7-dbms` and `azure-mysql-5-7-database` services, on the other hand, can be combined to provision multiple databases on a single DBMS.  For more information on each service, refer to the descriptions below.
 
 ## Services & Plans
 
-### Service: azure-mysql
+### Service: azure-mysql-5-7
 
 | Plan Name | Description |
 |-----------|-------------|
-| `basic50` | Basic Tier, 50 DTUs |
-| `basic100` | Basic Tier, 100 DTUs |
+| `basic` | Basic Tier, Up to 2 vCores, Variable I/O performance |
+| `general-purpose` | General Purporse Tier, Up to 32 vCores, Predictable I/O Performance, Local or Geo-Redundant Backups |
+| `memory-optimized` | Memory Optimized Tier, Up to 16 memory optimized vCores, Predictable I/O Performance, Local or Geo-Redundant Backups |
 
 #### Behaviors
 
@@ -39,6 +40,37 @@ Provisions a new MySQL DBMS and a new database upon it. The new database will be
 | `firewallRules[n].startIPAddress` | `string` | Specifies the start of the IP range allowed by this firewall rule | Y | |
 | `firewallRules[n].endIPAddress` | `string` | Specifies the end of the IP range allowed by this firewall rule | Y | |
 | `tags` | `map[string]string` | Tags to be applied to new resources, specified as key/value pairs. | N | Tags (even if none are specified) are automatically supplemented with `heritage: open-service-broker-azure`. |
+
+The three plans each have additional provisioning parameters with different default and allowed values. See the tables below for details on each.
+
+####### Provisioning Parameters: basic
+
+| Parameter Name | Type | Description | Required | Default Value |
+|----------------|------|-------------|----------|---------------|
+| `hardwareFamily` | `string` | Specifies the underlying hardware type. Valid values are `gen4` or `gen5` | N | `gen5` | 
+| `cores` | `integer` | Specifies vCores, which represent the logical CPU. Valid values are 1 or 2 | N | 1 |
+| `storage` | `integer` | Specifies the amount of storage to allocate in GB. Ranges from 5 to 1048 | N | 10 |
+| `backupRetention` | `integer` | Specifies the number of days to retain backups. Ranges from 7 to 35 | N | 7 |
+
+
+####### Provisioning Parameters: general-purpose
+
+| Parameter Name | Type | Description | Required | Default Value |
+|----------------|------|-------------|----------|---------------|
+| `hardwareFamily` | `string` | Specifies the underlying hardware type. Valid values are `gen4` or `gen5` | N | `gen5` | 
+| `cores` | `integer` | Specifies vCores, which represent the logical CPU. Valid values are 1 or 2 | N | 1 |
+| `storage` | `integer` | Specifies the amount of storage to allocate in GB. Ranges from 5 to 1048 | N | 10 |
+| `backupRetention` | `integer` | Specifies the number of days to retain backups. Ranges from 7 to 35 | N | 7 |
+| `backupRedundancy` | `string` | Specifies the backup redundancy, either `local` or `geo` | N | `local` |
+
+####### Provisioning Parameters: memory-optimized
+
+| Parameter Name | Type | Description | Required | Default Value |
+|----------------|------|-------------|----------|---------------|
+| `cores` | `integer` | Specifies vCores, which represent the logical CPU. Valid values are 1 or 2 | N | 1 |
+| `storage` | `integer` | Specifies the amount of storage to allocate in GB. Ranges from 5 to 1048 | N | 10 |
+| `backupRetention` | `integer` | Specifies the number of days to retain backups. Ranges from 7 to 35 | N | 7 |
+| `backupRedundancy` | `string` | Specifies the backup redundancy, either `local` or `geo` | N | `local` |
 
 ##### Bind
 
@@ -139,12 +171,13 @@ curl -X PUT \
 }'
 ```
 
-### Service: azure-mysql-dbms
+### Service: azure-mysql-5-7-dbms
 
 | Plan Name | Description |
 |-----------|-------------|
-| `basic50` | Basic Tier, 50 DTUs |
-| `basic100` | Basic Tier, 100 DTUs |
+| `basic` | Basic Tier, Up to 2 vCores, Variable I/O performance |
+| `general-purpose` | General Purporse Tier, Up to 32 vCores, Predictable I/O Performance, Local or Geo-Redundant Backups |
+| `memory-optimized` | Memory Optimized Tier, Up to 16 memory optimized vCores, Predictable I/O Performance, Local or Geo-Redundant Backups |
 
 #### Behaviors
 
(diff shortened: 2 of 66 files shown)


contrib/k8s/examples/mysql/advanced/mysql-database-instance.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: example-mysql-database-instance
   namespace: default
 spec:
-  clusterServiceClassExternalName: azure-mysql-database
+  clusterServiceClassExternalName: azure-mysql-5-7-database
```
krancour: Missed this (and other similar updates to the examples) in #407. Thanks for catching this and fixing it.

docs/modules/mysql.md
```diff
@@ -6,20 +6,21 @@ Open Service Broker for Azure contains three Azure Database for MySQL services.
...
+| `azure-mysql-5-7-database` | Provision a new database only upon a previously provisioned DBMS. |
 
-The `azure-mysql` service allows you to provision both a DBMS and a database. When the provision operation is successful, the database will be ready to use. You can't provision additional databases onto an instance provisioned through this service. The `azure-mysql-dbms` and `azure-mysql-database` services, on the other hand, can be combined to provision multiple databases on a single DBMS.  For more information on each service, refer to the descriptions below.
+The `azure-mysql-5-7l` service allows you to provision both a DBMS and a database. When the provision operation is successful, the database will be ready to use. You can't provision additional databases onto an instance provisioned through this service. The `azure-mysql-5-7-dbms` and `azure-mysql-5-7-database` services, on the other hand, can be combined to provision multiple databases on a single DBMS.  For more information on each service, refer to the descriptions below.
```
krancour: `azure-mysql-5-7l` seems to have an errant "l" in it.

jeremyrickard: I _swore_ I had cleaned that one up! Sure enough, hadn't hit save in VS Code :(

docs/modules/postgresql.md
```diff
@@ -43,6 +44,36 @@ name.
 | `tags` | `map[string]string` | Tags to be applied to new resources, specified as key/value pairs. | N | Tags (even if none are specified) are automatically supplemented with `heritage: open-service-broker-azure`. |
 | `extensions` | `string[]` | Specifies a list of PostgreSQL extensions to install | N | |
 
+####### Provisioning Parameters: basic
```
jeremyrickard: We didn't have an established pattern for this, so I thought this might be a good way to handle these plan specific things. We do have the common params for the service as well, so i left them up above.

pkg/services/mysql/all_in_one_arm_template.go
```diff
@@ -78,7 +51,7 @@ var allInOneARMTemplateBytes = []byte(`
 					"type": "firewallrules",
 					"apiVersion": "[variables('DBforMySQLapiVersion')]",
 					"dependsOn": [
-						"[concat('Microsoft.DBforMySQL/servers/', parameters('serverName'))]"
+						"[concat('Microsoft.DBforMySQL/servers/', variables('ServerName'))]"
 					],
 					"location": "[parameters('location')]",
```
jeremyrickard: We currently do tags/location into the map at the deployer layer. We'll need to do a refactor there to switch fully over to go params. should be easy enough to do, but we'll need to bump all the arm templates.

pkg/services/mysql/database_arm_template.go
```diff
@@ -24,11 +24,11 @@ var databaseARMTemplateBytes = []byte(`
 			}
 		},
 		"variables": {
-			"DBforMySQLapiVersion": "2017-04-30-preview"
+			"DBforPostgreSQLapiVersion": "2017-12-01"
```
krancour: Looks like this says "postgres" where it should say "mysql." It's just a variable name, but we should still fix it.

pkg/services/mysql/dbms_arm_template.go
```diff
@@ -9,61 +9,39 @@ var dbmsARMTemplateBytes = []byte(`
...
-				"defaultValue": "Enabled"
-			},
 			"tags": {
 				"type": "object"
 			}
 		},
 		"variables": {
-			"DBforMySQLapiVersion": "2017-04-30-preview"
+			"DBforMySQLapiVersion": "2017-12-01",
+			"ServerName" : "{{ .serverName }}"
```
krancour: This doesn't buy us anything, afaict. Might be easier to read if we just use `{{ .serverName }}` wherever.

pkg/services/postgresql/all_in_one_arm_template.go
```diff
@@ -9,67 +9,39 @@ var allInOneARMTemplateBytes = []byte(`
...
-			"defaultValue": "Enabled"
-		},
 		"tags": {
 			"type": "object"
 		}
 	},
 	"variables": {
-		"DBforPostgreSQLapiVersion": "2016-02-01-privatepreview"
+		"DBforPostgreSQLapiVersion": "2017-12-01",
+		"ServerName" : "{{ .serverName }}"
```
krancour: Let's get rid of this extra bit of indirection.

pkg/services/mysql/dbms_arm_template.go
jeremyrickard: It’s used within the firewall block, which is selecting the .firewallRules so you can’t use {{ .serviceName }} directly. Seemed more readable to just declare an arm variable and use it throughout, but open to suggestions!

pkg/services/mysql/all_in_one_arm_template.go
```diff
@@ -9,66 +9,39 @@ var allInOneARMTemplateBytes = []byte(`
...
-			"defaultValue": "Enabled"
-		},
 		"tags": {
 			"type": "object"
 		}
 	},
 	"variables": {
-		"DBforMySQLapiVersion": "2017-04-30-preview"
+		"DBforMySQLapiVersion": "2017-12-01",
+		"ServerName" : "{{ .serverName }}"
```
krancour: Let's get rid of this extra bit of indirection.

pkg/services/postgresql/dbms_arm_template.go
```diff
@@ -9,62 +9,39 @@ var dbmsARMTemplateBytes = []byte(`
...
-			"defaultValue": "Enabled"
-		},
 		"tags": {
 			"type": "object"
 		}
 	},
 	"variables": {
-		"DBforPostgreSQLapiVersion": "2016-02-01-privatepreview"
+		"DBforPostgreSQLapiVersion": "2017-12-01",
+		"ServerName" : "{{ .serverName }}"
```
krancour: Let's get rid of this extra bit of indirection.

pkg/services/mysql/all_in_one_provision.go
```diff
@@ -11,20 +11,19 @@ import (
...
+	plan service.Plan,
 	provisioningParameters service.ProvisioningParameters,
 	_ service.SecureProvisioningParameters,
 ) error {
 	pp := allInOneProvisioningParameters{}
-	if err := service.GetStructFromMap(provisioningParameters, &pp); err != nil {
+	if err := service.GetStructFromMap(
+		provisioningParameters,
+		&pp,
+	); err != nil {
```
krancour: I'm probably going to undo this. It doesn't bother me, but since this is a big PR, I want to try and limit the changes to only those that are meaningful-- just to keep the PR as a whole more comprehensible.

pkg/services/mysql/all_in_one_provision.go
```diff
@@ -52,7 +51,10 @@ func (a *allInOneManager) preProvision(
 	}
 	pp := allInOneProvisioningParameters{}
 	if err :=
-		service.GetStructFromMap(instance.ProvisioningParameters, &pp); err != nil {
+		service.GetStructFromMap(
+			instance.ProvisioningParameters,
+			&pp,
+		); err != nil {
```
krancour: https://github.com/Azure/open-service-broker-azure/pull/416/files#r184577223

pkg/services/mysql/all_in_one_provision.go
```diff
@@ -117,38 +94,44 @@ func (a *allInOneManager) deployARMTemplate(
 		return nil, nil, err
 	}
 	sdt := secureAllInOneInstanceDetails{}
-	if err := service.GetStructFromMap(instance.SecureDetails, &sdt); err != nil {
+	if err := service.GetStructFromMap(
+		instance.SecureDetails,
+		&sdt,
+	); err != nil {
```
krancour: https://github.com/Azure/open-service-broker-azure/pull/416/files#r184577223

pkg/services/mysql/all_in_one_provision.go
```diff
@@ -117,38 +94,44 @@ func (a *allInOneManager) deployARMTemplate(
...
 		return nil, nil, err
 	}
 	pp := allInOneProvisioningParameters{}
 	if err :=
-		service.GetStructFromMap(instance.ProvisioningParameters, &pp); err != nil {
+		service.GetStructFromMap(
+			instance.ProvisioningParameters,
+			&pp,
+		); err != nil {
```
krancour: https://github.com/Azure/open-service-broker-azure/pull/416/files#r184577223

pkg/services/mysql/all_in_one_provision.go
```diff
@@ -117,38 +94,44 @@ func (a *allInOneManager) deployARMTemplate(
...
 		instance.Tags,
 	)
 	if err != nil {
 		return nil, nil, fmt.Errorf("error deploying ARM template: %s", err)
 	}
 
 	var ok bool
-	dt.FullyQualifiedDomainName, ok = outputs["fullyQualifiedDomainName"].(string)
+	dt.FullyQualifiedDomainName, ok =
+		outputs["fullyQualifiedDomainName"].(string)
```
krancour: https://github.com/Azure/open-service-broker-azure/pull/416/files#r184577223

pkg/service/catalog.go
```diff
@@ -166,40 +169,38 @@ func NewService(
...
-			nil,
-		)
-	}
+	serviceParamSchemas.addCommonSchema(serviceProperties)
+
 	for _, planIfc := range s.plans {
 		p := planIfc.(*plan)
-		p.ParameterSchemas = paramSchemas
+
+		// If the plan has it's own schema, ignore the service schema
```
krancour: s/it's/its/

krancour:
@jeremyrickard I've made a few comments. Please don't feel compelled to address them. Enjoy your day off and I will take action on most of the comments myself.

Note I'm still only about 75% done reviewing. Still trying to wrap my head around schema-related stuff. Will resume in the morning.

pkg/services/mysql/dbms_types.go
```diff
@@ -13,17 +18,16 @@ type firewallRule struct {
...
 type dbmsInstanceDetails struct {
 	ARMDeploymentName        string `json:"armDeployment"`
 	ServerName               string `json:"server"`
 	FullyQualifiedDomainName string `json:"fullyQualifiedDomainName"`
 	EnforceSSL               bool   `json:"enforceSSL"`
+	HardwareFamily           string `json:"hardwareFamily"`
+	Cores                    int    `json:"cores"`
+	Storage                  int    `json:"storage"`
+	BackupRetention          int    `json:"backupRetention"`
+	BackupRedundancy         string `json:"backupRedundancy"`
```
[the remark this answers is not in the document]
krancour: > Added them for later upgrade use so we can do validation of the update request vs what was there before

We can validate the update request against the original provisioning parameters if we need to since we hang on to them. I guess really, I should reframe my comment-- we don't need two copies of this info.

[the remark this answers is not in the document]
jeremyrickard: Yeah good point. And we can compare to previous update parameters for any subsequent update operations.

pkg/services/mysql/dbms_arm_template.go
krancour: @jeremyrickard in cases where `{{range}}` affects what `.` references, you can still reference what `.` referenced outside that loop by using `$.`

i.e. `$.serverName`

Admittedly, I didn't know this until quite recently.

I'm going to amend the PR with this fix.

pkg/services/mysql/dbms_types.go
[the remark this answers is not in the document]
krancour: > And we can compare to previous update parameters for any subsequent update operations.

When an update is complete, imo, the updating parameters should completely replace the original provisioning parameters so that the instance then appears as it if was originally provisioned in its current state. Or at least this has been my plan.

pkg/services/mysql/common_provision.go
```diff
@@ -13,7 +13,28 @@ import (
...
+		return fmt.Errorf("plan invalid")
+	}
+	s, ok := plan.GetProperties().Extended["provisionSchema"]
+	if !ok {
+		return fmt.Errorf("invalid plan, schema not found")
+	}
+	schema := s.(planSchema)
+	if err := schema.validateProvisionParameters(pp); err != nil {
+		return err
+	}
```
krancour: I'm a little concerned with how fragmented this feels. It seems we're delegating most of the validation to this schema object that enumerates permissible values for a host of _new_ provisioning parameters, but we're still handing validation for all the previously existing provisioning parameters over the next several lines here. Not sure why we wouldn't do all of this in one place using a single approach.

docs/modules/postgresql.md
krancour: Seems ok-- at least for now. Docs are easy enough to change later if we find a better way.

krancour:
The plans schemas and the validations felt fragmented to me. There were some params where default values were driven by the schemas and others where the defaults were encoded in logic elsewhere in the modules. Something similar was true for validations. There were functions for validating some params WRT a schema, but others were validated as they always have been.

I refactored a bit to unify the two approaches. Everything seems alright, but the changes are big enough that I think I'm exempt from LGTM'ing this PR (at least by myself) at this point. @jeremyrickard since we have both worked extensively on this, I would suggest that we move forward only if we can mutually LGTM one another's efforts.

Also note one remaining weakness I perceive in here (which can be fixed later in a follow up). The same plan schema struct is used for all tiers of plan, even though some attributes aren't applicable to some tiers.

jeremyrickard:
I know this is my doing, but do you think the name "schema" is appropriate? It's really metadata about the plan, but in a more structured form than the map we used before. Maybe something to change after this?

pkg/slice/contains.go
```diff
@@ -0,0 +1,23 @@
+package slice
```
jeremyrickard: thanks for doing this. good for later reuse.

pkg/services/postgresql/plan_schemas.go
```diff
@@ -55,121 +57,83 @@ func (p *planSchema) generateHardwareFamilyString(
...
-			if v == coresParam {
-				coresValid = true
-			}
-		}
-		if !coresValid {
-			return service.NewValidationError(
-				"cores",
-				fmt.Sprintf(`invalid value : "%d"`, coresParam))
-		}
+	return map[string]service.ParameterSchema{
```
jeremyrickard: yeah, i agree this should have been consolidated. thanks!

jeremyrickard:
@krancour i found this description of the pricing tiers on a website. should we iupdate the descriptions to match?


Pricing tier | Target workloads
-- | --
Basic | Workloads that require light compute and I/O performance. Examples include servers used for development or testing or small-scale infrequently used applications.
General Purpose | Most business workloads that require balanced compute and memory with scalable I/O throughput. Examples include servers for hosting web and mobile apps and other enterprise applications.
Memory Optimized | High-performance database workloads that require in-memory performance for faster transaction processing and higher concurrency. Examples include servers for processing real-time data and high-performance transactional or analytical apps.

krancour:
> I know this is my doing, but do you think the name "schema" is appropriate? It's really metadata about the plan, but in a more structured form than the map we used before. Maybe something to change after this?

I do think we should change it to something more like `planMetadata` or something. But let's definitely do it in a follow-up. I'd also like to see more of our modules adopt this approach.

krancour:
>  i found this description of the pricing tiers on a website. should we iupdate the descriptions to match?

I'll think about that today and probably include it in v0.11.0.

Issue #323, opened 2018-03-06
Update CF example in README


jeremyrickard:
We should update the CF example in the readme to provide firewall rules. We default to only allowing Azure IPs and the user could be using cf dev or a CF cluster off Azure. At a minimum, we should probably point that out.

PR #418, opened 2018-04-27
Updated CF example to include firewall rules

jeremyrickard:
Fixes: #323


jeremyrickard:
This actually needs to have the plan and service updated as well! But probably after we merge the plan changes (CF builds from source, so as soon as that happens)

PR #425, opened 2018-05-01
Add OpenShift template

jimzim:
Add functionality to deploy OSBA on OpenShift using OpenShift Project template.

### DIFF
--- a/contrib/openshift/osba-os-template.yaml
+++ b/contrib/openshift/osba-os-template.yaml
@@ -1,0 +1,238 @@
+kind: Template
+apiVersion: v1
+objects:
+- kind: Secret
+  apiVersion: v1
+  metadata:
+    name: osba-redis
+    labels:
+      app: osba-redis
+      release: "osba"
+  type: Opaque
+  data:
+    redis-password: "cGFzc3dvcmQ="
+- kind: Secret
+  apiVersion: v1
+  metadata:
+    name: osba-open-service-broker-azure-auth
+    labels:
+      app: osba-open-service-broker-azure
+      release: "osba"
+  type: Opaque
+  data:
+    username: dXNlcm5hbWU=
+    password: cGFzc3dvcmQ=
+- kind: Secret
+  apiVersion: v1
+  metadata:
+    name: osba-open-service-broker-azure
+    labels:
+      app: osba-open-service-broker-azure
+      release: "osba"
+  stringData:
+    azure-client-secret: ${AZURE_CLIENT_SECRET}
+    encryption-key: This is a key that is 256 bits!!
+    redis-password: "password"
+    basic-auth-password: "password"
+- kind: Service
+  apiVersion: v1
+  metadata:
+    name: osba-redis
+    labels:
+      app: osba-redis
+      release: "osba"
+  spec:
+    ports:
+    - name: redis
+      port: 6379
+      targetPort: redis
+    selector:
+      app: osba-redis
+- kind: Service
+  apiVersion: v1
+  metadata:
+    name: osba-open-service-broker-azure
+    labels:
+      app: osba-open-service-broker-azure
+      release: "osba"
+  spec:
+    type: ClusterIP
+    selector:
+      app: osba-open-service-broker-azure
+    ports:
+    - name: http
+      protocol: TCP
+      port: 80
+      targetPort: 8080
+
+- apiVersion: v1
+  kind: PersistentVolumeClaim
+  metadata:
+    name: osba-redis-pv-claim
+  spec:
+    accessModes:
+      - ReadWriteOnce
+    resources:
+      requests:
+        storage: 10Gi
+
+- apiVersion: extensions/v1beta1
+  kind: Deployment
+  metadata:
+    name: osba-redis
+    labels:
+      app: osba-redis
+      release: "osba"
+  spec:
+    template:
+      metadata:
+        labels:
+          app: osba-redis
+      spec:
+        containers:
+        - name: osba-redis
+          image: "bitnami/redis:4.0.9"
+          imagePullPolicy: "IfNotPresent"
+          env:
+          - name: REDIS_PASSWORD
+            valueFrom:
+              secretKeyRef:
+                name: osba-redis
+                key: redis-password
+          ports:
+          - name: redis
+            containerPort: 6379
+          livenessProbe:
+            exec:
+              command:
+              - redis-cli
+              - ping
+            initialDelaySeconds: 30
+            timeoutSeconds: 5
+          readinessProbe:
+            exec:
+              command:
+              - redis-cli
+              - ping
+            initialDelaySeconds: 5
+            timeoutSeconds: 1
+          resources:
+            requests:
+              cpu: 100m
+              memory: 256Mi
+          volumeMounts:
+          - name: redis-data
+            mountPath: /bitnami/redis
+        volumes:
+        - name: redis-data
+          persistentVolumeClaim:
+            claimName: osba-redis-pv-claim
+
+- apiVersion: extensions/v1beta1
+  kind: Deployment
+  metadata:
+    name: osba-open-service-broker-azure
+    labels:
+      app: osba-open-service-broker-azure
+      release: "osba"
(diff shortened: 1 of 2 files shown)


krancour:
@jimzim You'll have to forgive my ignorance of OpenShift, but shouldn't the Helm chart work fine on OpenShift?

If it does, that would be my preference for how OSBA is installed. Something that concerns me here-- and again, I don't know much about OpenShift, is that this template doesn't look parameterized. It seems to me that you can only tweak configuration by manually editing what look like plain Kubernetes manifests. Is that right?

jimzim:
@krancour While it is possible to install via Helm, Helm charts are not currently supported in OpenShift.  It is not trivial to get Helm working on OpenShift making it difficult for customers to consume OSBA.  We worked with Red Hat to get this project template working so that OSBA can work seamlessly on both OpenShift 3.7 and 3.9.

krancour:
@jimzim, how are people installing Kubernetes Service Catalog onto OpenShift without Helm?

jeremyrickard:
@krancour service catalog is just installed in OpenShift by default now!

krancour:
Good to know. So follow-up question. What is the technical obstacle to using Helm with OpenShift?

jeremyrickard:
No CI, docs only.

contrib/openshift/osba-os-template.yaml
```diff
@@ -0,0 +1,238 @@
...
+  required: true
+- name: AZURE_CLIENT_ID
+  value: 
+  required: true
+- name: AZURE_DEFAULT_LOCATION
+  value: 
+  required: true
+- name: AZURE_CLIENT_SECRET
+  value: 
+  required: true
```
krancour: It seems to me that there are a lot of missing parameters here. Without more parameters, users cannot do things like run OSBA with a remote Redis or, much more importantly, override the default password or encryption keys. This needs to be addressed as soon as possible.

contrib/openshift/osba-os-template.yaml
```diff
@@ -0,0 +1,238 @@
...
+- kind: Secret
+  apiVersion: v1
+  metadata:
+    name: osba-redis
+    labels:
+      app: osba-redis
+      release: "osba"
+  type: Opaque
+  data:
+    redis-password: "cGFzc3dvcmQ="
```
krancour: This should be parameterized.

contrib/openshift/osba-os-template.yaml
```diff
@@ -0,0 +1,238 @@
...
+  apiVersion: v1
+  metadata:
+    name: osba-open-service-broker-azure-auth
+    labels:
+      app: osba-open-service-broker-azure
+      release: "osba"
+  type: Opaque
+  data:
+    username: dXNlcm5hbWU=
+    password: cGFzc3dvcmQ=
```
krancour: These should be parameterized.

contrib/openshift/osba-os-template.yaml
```diff
@@ -0,0 +1,238 @@
...
+                  key: encryption-key
+            - name: BASIC_AUTH_USERNAME
+              value: username
+            - name: BASIC_AUTH_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  name: osba-open-service-broker-azure
+                  key: basic-auth-password
+            - name: MIN_STABILITY
+              value: EXPERIMENTAL
```
krancour: This should be parameterized and needs to default to `PREVIEW`.

contrib/openshift/osba-os-template.yaml
krancour: I understand now that this is for consumption by the `Broker` resource.

contrib/openshift/osba-os-template.yaml
```diff
@@ -0,0 +1,238 @@
...
+- kind: Secret
+  apiVersion: v1
+  metadata:
+    name: osba-open-service-broker-azure
+    labels:
+      app: osba-open-service-broker-azure
+      release: "osba"
+  stringData:
+    azure-client-secret: ${AZURE_CLIENT_SECRET}
+    encryption-key: This is a key that is 256 bits!!
```
krancour: This should be parameterized.

contrib/openshift/osba-os-template.yaml
```diff
@@ -0,0 +1,238 @@
...
+    redis-password: "password"
+    basic-auth-password: "password"
```
krancour: These should be parameterized.

contrib/openshift/osba-os-template.yaml
```diff
@@ -0,0 +1,238 @@
...
+- kind: Service
```
krancour: This resource should be created conditionally, depending on whether one wants to use on-cluster Redis or a remote Redis (i.e. Azure Redis Cache).

contrib/openshift/osba-os-template.yaml
krancour: This resource should be created conditionally, depending on whether one wants to use on-cluster Redis or a remote Redis (i.e. Azure Redis Cache).

contrib/openshift/osba-os-template.yaml
```diff
@@ -0,0 +1,238 @@
...
+    selector:
+      app: osba-open-service-broker-azure
+    ports:
+    - name: http
+      protocol: TCP
+      port: 80
+      targetPort: 8080
+
+- apiVersion: v1
+  kind: PersistentVolumeClaim
```
krancour: This resource should be created conditionally, depending on whether one wants to use on-cluster Redis or a remote Redis (i.e. Azure Redis Cache).

contrib/openshift/osba-os-template.yaml
```diff
@@ -0,0 +1,238 @@
...
+        app: osba-open-service-broker-azure
+    template:
+      metadata:
+        labels:
+          app: osba-open-service-broker-azure
+          release: "osba"
+      spec:
+        containers:
+          - name: open-service-broker-azure
+            image: microsoft/azure-service-broker:v0.10.0
```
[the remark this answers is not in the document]
krancour: This should probably also be parameterized to facilitate deployment using edge build or custom build.

contrib/openshift/osba-os-template.yaml
```diff
@@ -0,0 +1,238 @@
...
+              value: ${AZURE_CLIENT_ID}
+            - name: AZURE_CLIENT_SECRET
+              valueFrom:
+                secretKeyRef:
+                  name: osba-open-service-broker-azure
+                  key: azure-client-secret
+            - name: AZURE_DEFAULT_LOCATION
+              value: ${AZURE_DEFAULT_LOCATION}
+            - name: REDIS_HOST
+              value: osba-redis
```
krancour: Users should have the option to use remote Redis, so this should only be `osba-redis` if users have elected an on-cluster Redis. Otherwise, the host (and port) should come from parameters.

contrib/openshift/osba-os-template.yaml
```diff
@@ -0,0 +1,238 @@
...
+  kind: ClusterServiceBroker
+  metadata:
+    name: osba
+  spec:
+    url: http://osba-open-service-broker-azure.osba.svc.cluster.local
+    authInfo:
+      basic:
+        secretRef:
+          name: "osba-open-service-broker-azure-auth"
+          namespace: "osba"
```
krancour: Namespace should not be hardcoded.

contrib/openshift/osba-os-template.yaml
```diff
@@ -0,0 +1,238 @@
...
+              initialDelaySeconds: 30
+              periodSeconds: 10
+              successThreshold: 1
+              timeoutSeconds: 2
+- apiVersion: servicecatalog.k8s.io/v1beta1
+  kind: ClusterServiceBroker
+  metadata:
+    name: osba
+  spec:
+    url: http://osba-open-service-broker-azure.osba.svc.cluster.local
```
krancour: This looks like it assumes the broker has been deployed in the `osba` namespace. That isn't a good assumption.

PR #431, opened 2018-05-08
schema-based validations pt 1

krancour:
This PR is a series of discrete commits that incrementally advance us toward driving validations off of schemas.

Highlights:

- Make schemas more intuitive to work with for module authors
  - Can no longer define at service level. Too hard to wrap one's head around the idea of "schema inheritance." Just eliminating this concept.
  - Renamed many schema types
  - Improved marshaling for many types-- don't need to ever specify `Type` field. It's handled transparently.
  - Replaced `SimpleParameterSchema` and `NumericParameterSchema` with more specific types `StringPropertySchema`, `IntPropertySchema`, and `FloatPropertySchema`. Why? These three things validate differently and Go doesn't have generics.
  - Added `DefaultValue` attribute to `ArrayPropertSchema` and `ObjectPropertySchema`. This will allow us to do more complex things in the future like adding a default list of firewall rules for when none are explicitly defined.
 - Less magic in `NewService()` function == more transparency

Do not merge this. It's a WIP and I just want to leverage CI against what is completed so far.

What's still missing is that even though schema types all support validation, validation of input parameters for provisioning and binding operations does _not_ yet leverage that capability-- that's next.


krancour:
> Note to self: Ensure that validations are done using a depth-first traversal of the schemas. This makes logical sense-- validating an object may require looking at multiple attributes of that object and doing that is likely to involve type assertions. It's useful for those individual fields to have been validated first so that correctness of type can be taken for granted.

I've double-checked that custom validator functions for array and object fields are the last validation fired for properties of those types, so this should be all set.

krancour:
Making note that I want #434 and #437 to go in first and this PR will need a non-trivial rebase after that.

pkg/service/schema.go
```diff
@@ -111,6 +173,48 @@ func (i IntPropertySchema) MarshalJSON() ([]byte, error) {
 	)
 }
 
+func (i IntPropertySchema) validate(context string, value interface{}) error {
+	floatVal, ok := value.(float64)
+	if !ok {
+		return NewValidationError(context, "field value is not of type int64")
```
jeremyrickard: should we return int64 as part of the validation error? JSON Schema really only defines "integer" and "number". Might be better to return the generic type integer here.

pkg/service/schema.go
```diff
@@ -136,12 +240,50 @@ func (f FloatPropertySchema) MarshalJSON() ([]byte, error) {
 	)
 }
 
+func (f FloatPropertySchema) validate(context string, value interface{}) error {
+	val, ok := value.(float64)
+	if !ok {
+		return NewValidationError(context, "field value is not of type float64")
```
jeremyrickard: Same comment as with int64. Maybe better to return field is not of type number?

pkg/service/schema.go
jeremyrickard: 👍

Issue #433, opened 2018-05-11
Framework needs some changes to support instance updates and failure cases


jeremyrickard:
The API surface and framework currently have some issues in allowing instance updates to successfully be implemented. 

Some issues:
* Can't deprovision a failed update
* Can't update a failed update
* We don't update the provision params with an update request, we should do this to reflect the actual request so it can be reflected when we implement the OSB GET endpoint

PR #434, opened 2018-05-11
Update to framework to enable instance updates

jeremyrickard:
This PR is the first step in allowing instance updates to happen. 

Summary: 
* Updated pkg/api/deprovision.go to allow an instance that failed to update to be deprovisioned
* Updated pkg/api/update.go to allow you to submit another update request on an instance that had failed. This one will need to be evolved further to possibly add retry logic (along with provision requests) that fail due to Azure timeouts or other non-request related failures. 
* Added a new method to pkg/azure/arm/deployer called Update(...). This has slightly different state checking logic than Deploy(..). They both call the same method, which i've renamed to "doDeployment" to make it clear that it is used for both *new* deployments and additional (update) deployments to an existing deployment. doDeployment(..) calls CreateOrUpdate(..) on the azure sdk so it is a little clearer. 
* Updated pkg/broker/update.go to merge the update parameters into the provision parameters on successful update. OSB spec says we should not change anything that wasn't included in the update request, so this uses a new utility to generically check if the value in the update parameters is a empty value (or empty string) and only copies if that is not the case
* Added types package with IsEmpty(..) method to check if the interface{} value is empty. 
* Also fixes a small bug in schemas.go (which may be OBE based on #431) 

Fixes: #433


jeremyrickard:
Merging with @krancour's previous review that I had to dismiss in order to get the build to go again.

PR #440, opened 2018-05-16
Bump helm-chart-publishing-tools image to v0.1.1

carolynvs:
* Uses az go cli instead of the python cli


krancour:
Going to merge this without CI because the change only affects CD.

PR #445, opened 2018-05-18
improve schema-drive validations

krancour:
Highlights:

- Fixes bug where updating params were allowed that were not in the updating params schema-- this could effectively allow someone to attempt an update of a field that was not meant to be updated. It does this by eliminating support for `additionalProperties` at the "top-level" schema (i.e. `InputParametersSchema` type). `additionalProperties` remain supported on the `ObjectPropertySchema` type.

- On `ObjectPropertySchema`, if additional properties aren't explicitly allowed, `additionalProperties: false` is automatically set upon marshaling to JSON schema.

- Validations are more tolerant of types. This is because our control over what's in a map is somewhat imperfect. If a parameter map has been unmarshaled from JSON, all numeric types (ints and floats alike) are probably `float64`. If, however, you're writing a test where you build a parameter map directly, one is likely to have forgotten that and is likely to use `int` as a type (in cases where they're actually dealing with integers, obviously). Validations are more forgiving of this now and as long as they can make heads or tails out of what they're looking at, they will try to apply validations as codified in the schema.

### DIFF
--- a/pkg/service/catalog_test.go
+++ b/pkg/service/catalog_test.go
@@ -69,53 +69,56 @@ func init() {
 	})
 
 	//nolint: lll
-	testCatalogTemplate := `{
-            "services":[
-                {
-                    "name":"%s",
-                    "id":"%s",
-                    "description":"%s",
-                    "tags":["%s"],
-                    "bindable":%t,
-                    "plan_updateable":%t,
-                    "plans":[
-                        {
-                            "id":"%s",
-                            "name":"%s",
-                            "description":"%s",
-                            "free":%t,
-                            "schemas": {
-                              	"service_instance": {
-                                	"create": {
-                                		"parameters": {
-                                			"$schema": "http://json-schema.org/draft-04/schema#",
-                                			"type": "object",
-                                			"properties": {
-                                				"location": {
-                                					"type": "string",
-                                					"description": "%s"
-                                				},
-                                				"resourceGroup": {
-                                					"type": "string",
-                                					"description": "%s"
-                                                },
-                                                "tags": {
-                                                    "type": "object",
-                                                    "description": "%s",
-                                                    "additionalProperties" : {
-                                                        "type" : "string"
-                                                    }
-                                                }
-                                			}
-                                		}
-                                	}
-                                }
-                            }
-                        }
-                    ]
-                }
-            ]
-        }`
+	testCatalogTemplate := `
+	{
+		"services":[
+			{
+				"name":"%s",
+				"id":"%s",
+				"description":"%s",
+				"tags":["%s"],
+				"bindable":%t,
+				"plan_updateable":%t,
+				"plans":[
+					{
+						"id":"%s",
+						"name":"%s",
+						"description":"%s",
+						"free":%t,
+						"schemas": {
+							"service_instance": {
+								"create": {
+									"parameters": {
+										"$schema": "http://json-schema.org/draft-04/schema#",
+										"type": "object",
+										"properties": {
+											"location": {
+												"type": "string",
+												"description": "%s"
+											},
+											"resourceGroup": {
+												"type": "string",
+												"description": "%s"
+											},
+											"tags": {
+												"type": "object",
+												"description": "%s",
+												"additionalProperties" : {
+													"type" : "string"
+												}
+											}
+										},
+										"additionalProperties": false
+									}
+								}
+							}
+						}
+					}
+        ]
+			}
+		]
+	}
+	`
 
 	testCatalogTemplate = strings.Replace(testCatalogTemplate, " ", "", -1)
 	testCatalogTemplate = strings.Replace(testCatalogTemplate, "\n", "", -1)
--- a/pkg/service/schema.go
+++ b/pkg/service/schema.go
@@ -34,7 +34,6 @@ type BindingSchemas struct {
 type InputParametersSchema struct {
 	RequiredProperties []string                  `json:"required,omitempty"`
 	PropertySchemas    map[string]PropertySchema `json:"properties,omitempty"`
-	Additional         PropertySchema            `json:"additionalProperties,omitempty"` // nolint: lll
 }
 
 // MarshalJSON defines custom JSON marshaling for InputParametersSchema and
@@ -46,6 +45,7 @@ func (i InputParametersSchema) MarshalJSON() ([]byte, error) {
 		Schema string `json:"$schema"`
 		Type   string `json:"type"`
 		inputParametersSchema
+		Additional bool `json:"additionalProperties"`
 	}
 	return json.Marshal(
 		struct {
@@ -55,6 +55,7 @@ func (i InputParametersSchema) MarshalJSON() ([]byte, error) {
 				Schema: jsonSchemaVersion,
 				Type:   "object",
 				inputParametersSchema: inputParametersSchema(i),
+				Additional:            false,
 			},
 		},
 	)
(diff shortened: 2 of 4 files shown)


krancour:
This needs one minor fix. Standby.

krancour:
> This needs one minor fix. Standby.

All set.

pkg/service/catalog_test.go
```diff
@@ -69,53 +69,56 @@ func init() {
...
+									}
+								}
+							}
+						}
+					}
+        ]
+			}
+		]
+	}
+	`
```
krancour: All of the above is just white-space fixes, _except_ the addition of `"additionalProperties": false` on line 110.

pkg/service/schema.go
```diff
@@ -55,6 +55,7 @@ func (i InputParametersSchema) MarshalJSON() ([]byte, error) {
 				Schema: jsonSchemaVersion,
 				Type:   "object",
 				inputParametersSchema: inputParametersSchema(i),
+				Additional:            false,
```
krancour: Always explicitly say that the top-level schema doesn't support additional properties. This is a critical in preventing attempts to update params that are in the provisioning params schema, but not in the updating params schema.

pkg/service/schema.go
```diff
@@ -70,14 +71,11 @@ func (i InputParametersSchema) Validate(valMap map[string]interface{}) error {
...
-		if ok {
-			if err := propertySchema.validate(k, v); err != nil {
-				return err
-			}
-		} else if i.Additional != nil {
-			if err := i.Additional.validate(k, v); err != nil {
-				return err
-			}
+		if !ok {
+			return NewValidationError(k, "unrecognized field")
```
krancour: Reject unrecognized fields.

pkg/service/schema.go
```diff
@@ -196,13 +194,41 @@ func (i IntPropertySchema) validate(context string, value interface{}) error {
...
+	} else if intVal, ok := value.(int32); ok {
+		val = int64(intVal)
+	} else if intVal, ok := value.(*int32); ok {
+		val = int64(*intVal)
+	} else if intVal, ok := value.(int); ok {
+		val = int64(intVal)
+	} else if intVal, ok := value.(*int); ok {
+		val = int64(*intVal)
+	} else {
+		return NewValidationError(context, "field value is not of type integer")
```
krancour: All of the above just makes validations more tolerant of the types in the map that _might_ represent an integer.

pkg/service/schema.go
```diff
@@ -278,9 +304,17 @@ func (f FloatPropertySchema) validate(context string, value interface{}) error {
...
+	if floatVal, ok := value.(float64); ok {
+		val = floatVal
+	} else if floatVal, ok := value.(*float64); ok {
+		val = *floatVal
+	} else if floatVal, ok := value.(float32); ok {
+		val = float64(floatVal)
+	} else if floatVal, ok := value.(*float32); ok {
+		val = float64(*floatVal)
+	} else {
+		return NewValidationError(context, "field value is not of type float")
```
krancour: All of the above just makes validations more tolerant of the types in the map that might represent a float.

pkg/service/schema.go
```diff
@@ -342,12 +376,16 @@ type ObjectPropertySchema struct {
 // MarshalJSON provides functionality to marshal an ObjectPropertySchema to JSON
 func (o ObjectPropertySchema) MarshalJSON() ([]byte, error) {
 	type objectPropertySchema ObjectPropertySchema
+	ops := objectPropertySchema(o)
+	if ops.Additional == nil {
+		ops.Additional = &falsePropertySchema{}
+	}
```
krancour: Oh... Go...

In JSON schema, `additionalProperties` either needs to be `false` or another schema object. Go is strongly typed and can't have something be a boolean or an object...

What we do here is if `Additional == nil`, we replace it with a dummy `falsePropertySchema` whose custom marshaling will just produce the JSON value `false`.

pkg/service/schema.go
```diff
@@ -365,19 +403,21 @@ func (o ObjectPropertySchema) validate(
 	for _, requiredProperty := range o.RequiredProperties {
 		_, ok := valMap[requiredProperty]
 		if !ok {
-			propetyContext := fmt.Sprintf("%s.%s", context, requiredProperty)
-			return NewValidationError(propetyContext, "field is required")
+			propertyContext := fmt.Sprintf("%s.%s", context, requiredProperty)
+			return NewValidationError(propertyContext, "field is required")
```
krancour: Just fixing a misspelling.

pkg/service/schema.go
```diff
@@ -365,19 +403,21 @@ func (o ObjectPropertySchema) validate(
...
 		if ok {
-			propetyContext := fmt.Sprintf("%s.%s", context, k)
-			if err := propertySchema.validate(propetyContext, v); err != nil {
+			if err := propertySchema.validate(propertyContext, v); err != nil {
 				return err
 			}
-		} else if o.Additional != nil {
-			if err := o.Additional.validate(k, v); err != nil {
+		} else if o.Additional == nil {
+			return NewValidationError(propertyContext, "unrecognized field")
```
krancour: If `Additional == nil` then unrecognized properties are _not_ allowed-- reject.

pkg/service/schema.go
```diff
@@ -497,3 +537,18 @@ func (p *PlanSchemas) addCommonSchema(sp *ServiceProperties) {
...
+
+// MarshalJSON provides functionality to marshal an ObjectPropertySchema to JSON
+func (falsePropertySchema) MarshalJSON() ([]byte, error) {
+	return []byte("false"), nil
+}
+
+// No-op
+func (falsePropertySchema) validate(string, interface{}) error {
+	return nil
+}
```
krancour: See [this comment](https://github.com/Azure/open-service-broker-azure/pull/445/files#r189351592)

pkg/service/schema_test.go
```diff
@@ -10,6 +10,97 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+func TestValidateInputParametersSchema(t *testing.T) {
```
krancour: There were no tests for `InputParametersSchema`. This adds them. They're mostly the same as the `ObjectPropertySchema` tests, except unconditionally don't allow additional properties.

pkg/service/schema_test.go
```diff
@@ -150,69 +254,84 @@ func TestValidateIntProperty(t *testing.T) {
...
+	// This should pass validation-- nil == no value == valid
 	err = ips.validate(fieldName, nil)
 	assert.Nil(t, err)
 
 	ips = IntPropertySchema{
 		MinValue: ptr.ToInt64(3),
 	}
-	err = ips.validate(fieldName, 2.0)
+	// This should fail validation because the value is too small
+	err = ips.validate(fieldName, 2)
```
krancour: You'll see a lot of this. Since validations got more tolerant of what's in the maps vis-a-vis types, we don't have to use floats in the tests where we would have preferred to use ints.

pkg/service/schema_test.go
```diff
@@ -48,6 +139,7 @@ func TestValidateStringProperty(t *testing.T) {
 	const fieldName = "xyz"
 
 	sps := StringPropertySchema{}
+	// This should fail validation because the value is of the wrong type
```
krancour: Most of the test changes are just additional comments.

pkg/services/fake/catalog.go
```diff
@@ -40,6 +40,15 @@ func (m *Module) GetCatalog() (service.Catalog, error) {
...
 				},
+				Schemas: service.PlanSchemas{
+					ServiceInstances: service.InstanceSchemas{
+						ProvisioningParametersSchema: service.InputParametersSchema{
+							PropertySchemas: map[string]service.PropertySchema{
+								"someParameter": &service.StringPropertySchema{},
+							},
+						},
+					},
+				},
```
krancour: Validation now rejects fields not in the schemas. Functional tests against the `api` package often use the parameter `someParameter`. Those were all failing because they weren't in the schema for this fake plan-- so I added schema for the fake plan.

PR #457, opened 2018-05-21
Refine mssql binding/unbinding to not involve unnecessary Login

zhongyi-zhang:
Currently mssql binding creates a Login in the master database, then creates a user in the new database for the Login. Then Login is unnecessary and brings into one more db connection and operation. Similar to unbinding. This PR cuts down the Login from binding and unbinding.
It's a breaking change for bound mssql service instances. Users should unbind all the mssql instances before upgrading. As the module is in PREVIEW stability, I think it is OK to make this change. Rather, we should make this change before STABLE.


krancour:
@zhongyi-zhang compared to Postgres and MySQL, I do not know MS SQL well. Has this _always_ been unnecessary? Or is it new that this is not necessary.

zhongyi-zhang:
@krancour I am not sure whether it was new when we added the mssql module. Maybe I didn't catch up with the new feature at that time...
Logins are sever-level principals. The users to create in bindings are database-level. So contained database users are enough. But your concern is right. I double checked the doc and tested it e2e: Azure Data Warehouse doesn't support contained database users and our mssql module has service plans to create warehouse.
FYI: https://docs.microsoft.com/en-us/sql/relational-databases/security/contained-database-users-making-your-database-portable

So we can keep creating logins and just close the PR. Or separate warehouse plans to a new service first and apply different approaches to the corresponding services. Both are OK to me. What do you think? Looks the benefit of the latter is not obvious?

zhongyi-zhang:
Oh, https://github.com/Azure/open-service-broker-azure/pull/458 is going to remove warehouse plans. So we can merge this after that PR merged.

jeremyrickard:
I'm in favor of making it a separate service. I've already removed it from the existing plans for sql-database, as it didn't really fit in terms of parameters/pricing tiers. 

I was going to make an issue to put it into a new module, but we could put it in this module w/ different bind behavior or put into another module. The params and stuff are different enough that I think maybe a stand alone thing is god enough (the ARM template is mostly the same, but it's using an older version of the API.

zhongyi-zhang:
@jeremyrickard agree!

krancour:
@jeremyrickard you are more in touch with details of SQL Server than me. Do you agree with @zhongyi-zhang's proposed change?

krancour:
@jeremyrickard I know you've done recent refactoring on this module. Are the changes from this PR already implemented by another recent PR? Or is this still relevant?

Issue #392, opened 2018-04-12
Update SQL Database Plans


jeremyrickard:
The Azure SQL Database Plans seem to be out of date with what's in https://azure.microsoft.com/en-us/pricing/details/sql-database/managed

Should we update this to match current state @krancour @arschles? It looks like the DTU pricing is still available but it's _not_ what shows up if you load the page by default.

zhongyi-zhang:
The problem is how to define plans for the core model. The flexibility of core-model pricing looks against these specified service plans, unless the broker can delegate the specification of plans to the cluster operator.

jeremyrickard:
@zhongyi-zhang I think we can leave them as well for now and then deprecate them. I chatted with a PM from MySQL and PostgreSQL and she indicated that we were generally moving away from the DTU based pricing. OSB is specifically looking at a mechanism for adding deprecation to plans, so once that's a thing we can look at adding it to the catalog properly. This was more about adding those new plans to mirror what's there.  Mostly likely defining the three sku types and accepting the other bits as parameters, with defaults that match the portal now.

zhongyi-zhang:
@jeremyrickard mirror plans sound good.

jeremyrickard:
@zhongyi-zhang we are in the process of adding the newer plans for SQL Database. In the case of MySQL and PostgreSQL, the PM for those products wanted us to remove the older plans. For SQL Database, they need to keep the older plans around as well because the new vCore plans won't have sub-core capability that you can get with DTU based plans.

I'd like to propose we refactor the standard and premium plans to follow the same model and expose DTUs as a parameter (we have the ability to express allowed values using JSON Schema now so we can constrain it in docs and via schema/param validation) in order to simplify things a little more, so we maybe end up with:

plans:
basic
standard
premium
general-purpose (new vCore plan)
business-critical  (new vCore plan)

zhongyi-zhang:
@jeremyrickard sounds good. Thanks for the info!

PR #458, opened 2018-05-21
Update SQL Database plans

jeremyrickard:
This PR updates the Azure SQL Database module to do the following:

* Removes the Data Warehouse Plans (we can make them a new service if we want them back, but they have different bind behavior and pricing/parameters so they don't make sense imho as part of sql database). 
* Updates the arm templates to use the new API that supports vCore plans
* Introduce the new "general purpose" and "business critical" vCore Plans. 
   - This is implemented in a manner similar to the MySQL and PostgreSQL modules, with single plans that expose cores and storage as provision parameters
* Refactors the existing standard and premium plans to be collapsed into a single plan for each, with DTU exposed as a provision parameter. Increased the number of "tiers" for each one up to 3000 DTUs for standard and 4000 for Premium

Closes: #392


pkg/services/mssql/all_in_one_arm_template.go
```diff
@@ -9,43 +9,19 @@ var allInOneARMTemplateBytes = []byte(`
...
 			"type": "Microsoft.Sql/servers",
-			"name": "[parameters('serverName')]",
-			"apiVersion": "[variables('SQLapiVersion')]",
+			"name": "{{ .serverName }}",
+			"apiVersion": "2015-05-01-preview",
 			"location": "[parameters('location')]",
 			"properties": {
-				"administratorLogin": "[parameters('administratorLogin')]",
-				"administratorLoginPassword": "[parameters('administratorLoginPassword')]",
+				"administratorLogin": "{{ .administratorLogin }}",
```
krancour: 💯 on these `{{}}` changes.

pkg/services/mssql/all_in_one_arm_template.go
```diff
@@ -54,33 +30,35 @@ var allInOneARMTemplateBytes = []byte(`
 				{
 					"type": "firewallrules",
 					"name": "{{.Name}}",
-					"apiVersion": "[variables('SQLapiVersion')]",
+					"apiVersion": "2014-04-01-preview",
```
krancour: I think you had mentioned this to me in passing-- just double-checking, however, that the `apiVersion` on the firewall rules _is_ supposed to be different from the `apiVersion` for the server (and that that `apiVersion` _is_ supposed to be different from the one for the database, for that matter).

tests/lifecycle/mssql_cases_test.go
```diff
@@ -67,6 +68,24 @@ var mssqlTestCases = []serviceLifecycleTestCase{
...
+			"storage": int64(25),
+			"firewallRules": []interface{}{
+				map[string]interface{}{
+					"name":           "AllowAll",
+					"startIPAddress": "0.0.0.0",
+					"endIPAddress":   "255.255.255.255",
+				},
+			},
+		},
+	},
```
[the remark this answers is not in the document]
jeremyrickard: Yeah. I was thinking of just adding it in parallel to the one above so we can verify that both work on the same DBMS. I tried that outside of code and it worked but want to verify

pkg/services/mssql/plan_details.go
```diff
@@ -0,0 +1,244 @@
...
+type vCorePlanDetails struct {
+	tier          string
+	tierShortName string
+	includeDBMS   bool
+}
+
+func (v vCorePlanDetails) getProvisionSchema() service.InputParametersSchema {
+	schema := service.InputParametersSchema{
+		PropertySchemas: map[string]service.PropertySchema{
+			"cores": service.IntPropertySchema{
```
krancour: I know this compiles, but is this working ok? (Schema is right and validations are as expected?) In the MySQL and PostgreSQL modules, I used references for all property schemas. _I can't remember why, but it might have had more to do with where I am headed with param access._ Looking at the functions on those types, the receivers all take non-references, so it would seem this _should_ be ok. (In the case of MySQL and PostgreSQL, Go is automatically dereferencing those references when the functions are invoked on them.) Bottom line-- no need to change this as long as you're certain it is working. If my upcoming PR breaks this, then that's on me to make the param access stuff more resilient so that this sensible bit of code here _doesn't_ break (and so module authors don't need to be hyper-aware of whether they should use a reference here or not).

krancour:
A few nits; few questions. But overall, LGTM. Just waiting on your responses before I make that official.

pkg/services/mssql/plan_details.go
jeremyrickard: Good question! I just assumed it was working because it was compiling. Let me validate that it is.

jeremyrickard: Validated.

pkg/services/mssql/all_in_one_arm_template.go
jeremyrickard: yep. I ran into issues using just the latest for everything and tried a few other combinations before I gave up and just fell back to what the portal produces if you use the `Automation Options` button

PR #463, opened 2018-05-23
Adding Cognitive Services Text Analytics API module

neilpeterson:
I understand that there is a hold on new modules, no worries. I have a need for this functionality in a toy project. I am contributing the module back in the event that it is useful / of interest.

I've configured it as experimental;  lilting and life cycle tests look good.

```
--- PASS: TestServices/lifecycle/TestServices/lifecycle/cognitive/text-analytics-s4 (95.15s)
--- PASS: TestServices/lifecycle/TestServices/lifecycle/cognitive/text-analytics-s3 (110.64s)
--- PASS: TestServices/lifecycle/TestServices/lifecycle/cognitive/text-analytics-s2 (110.00s)
--- PASS: TestServices/lifecycle/TestServices/lifecycle/cognitive/text-analytics-s1 (95.40s)
--- PASS: TestServices/lifecycle/TestServices/lifecycle/cognitive/text-analytics-s0 (104.62s)
--- PASS: TestServices/lifecycle/TestServices/lifecycle/cognitive/text-analytics-free (95.59s)
```

I've named the module `cognitive` with the expectation that additional Cognitive Service APIs such as face recognition, language, and emotion API plans could be added. I would be happy to refactor if a different class / plan schema should be used.

Simple Demo - https://github.com/neilpeterson/open-service-broker-azure-samples/tree/master/osba-text-analytics

Thanks


krancour:
@neilpeterson thanks for this awesome contribution. I can tell a lot of work went into this and at a glance, I think it looks pretty good. While we _are_ in a little bit of a holding pattern on new modules, I don't see any reason that we can't merge this _eventually_. I just don't know how soon.

One bit of early feedback I can give (if @jeremyrickard concurs) is on the structure of services and plans that are offered. I think all of the cognitive services (of which there are many) would share a lot of common management code (provision, bind, etc.), so expanding this one module over time to manage multiple cognitive services seems to make sense to me...

That said, I think each of the cognitive services that are offered are different enough from one another (in terms of the functionality they each provide) that from a catalog perspective, they should be separate services. i.e. Instead of having one service "cognitive service" with many plans representing different types of cognitive services, we should promote individual services (sorry for overloading of the word "service") to be services _in the catalog_. So the service here would be "azure text analytics" or something like that.

As for plans thereof, I poked around the portal a bit and it seems that the features user inputs required for provisioning don't vary by tier. With the only difference between tiers, from what I can tell, being quotas, I would suggest that the proposed "azure text analytics" service have just a single plan and that tier be specified by means of a provisioning parameter that can be defaulted to something sensible like S0.

@jeremyrickard would you concur with this analysis?

neilpeterson:
@krancour - thanks for the review and notes. This makes sense to me. 

`..that from a catalog perspective, they should be separate services.`

PR #464, opened 2018-05-24
Enable instance update for SQL DB

jeremyrickard:
Enable instance update for SQL DB.

Some notes:

* For vCore based plans, we only allow storage to be increased. The portal allows both directions, but to be safe we are only allowing an increase
* Firewall rules can of course be updated on DBMS / All In One
* Removed ssl enabled from the params because it's not valid for SQL DB


pkg/services/mssql/plan_details.go
```diff
@@ -22,6 +22,11 @@ type planDetails interface {
 	getTierProvisionParameters(
 		service.Instance,
 	) (map[string]interface{}, error)
+	getTierUpdateParameters(
+		service.Instance,
+	) (map[string]interface{}, error)
+	getUpgradeSchema() *service.InputParametersSchema
```
krancour: Nit: I think elsewhere, we use the term "update" rather than "upgrade." Can we change this for consistency? Or am I overlooking a reason for "upgrade" being used here?

pkg/services/mssql/plan_details.go
```diff
@@ -56,7 +64,12 @@ func (d dtuPlanDetails) getProvisionSchema() service.InputParametersSchema {
 	if d.includeDBMS {
 		addDBMSParameters(schema.PropertySchemas)
 	}
-	return schema
+	return &schema
+}
+
+func (d dtuPlanDetails) getProvisionSchema() service.InputParametersSchema {
+	schema := d.getUpgradeSchema()
+	return *schema
```
krancour: Just confirming-- for ftu-based plan, all provisioning params are updatable?

pkg/services/mssql/common_update.go
```diff
@@ -0,0 +1,40 @@
+package mssql
+
+import (
+	"github.com/Azure/open-service-broker-azure/pkg/service"
+)
+
+func buildDatabaseUpdateGoTemplateParameters(
```
krancour: iirc, in other modules we haven't differentiated between the Go template params that are fed into an initial ARM deployment vs those that are fed into an update to an existing ARM deployment. Here, the name seems to imply we're differentiating those two cases. I'm sure there is a reason for that. Can you explain for my own edification?

pkg/services/mssql/dbms_types.go
```diff
@@ -3,8 +3,7 @@ package mssql
 import "github.com/Azure/open-service-broker-azure/pkg/service"
 
 type dbmsProvisioningParams struct {
-	SSLEnforcement string         `json:"sslEnforcement"`
```
zhongyi-zhang: It should be also removed from the doc.

pkg/services/mssql/plan_details.go
jeremyrickard: Yeah, can do.

pkg/services/mssql/dbms_types.go
jeremyrickard: I don't think it was actually IN the docs so this made the code reflect the true state.

pkg/services/mssql/common_update.go
jeremyrickard: The nature of this model (with different structure of params based on the plans) made it feel cleaner to implement in terms of interfaces, so in this case I needed to have a method that dealt with the update parameters from the instance vs the provision params. This was an attempt to make the methods very clearly spell out what they were working with. It is using the same template, so perhaps better naming might be:

_buildDatabaseGoTemplateParametersForUpdate_ and _buildDatabaseGoTemplateParametersForProvision_ to make it clear? 

The main difference is really: 
_p, err = details.getTierUpdateParameters(instance)_ vs _	p, err = details.getTierProvisionParameters(instance)_

pkg/services/mssql/dbms_types.go
zhongyi-zhang: Oh yes, it is just a invalid parameter.

PR #468, opened 2018-06-05
Updating roadmap

seanmck:
Updating the roadmap to reflect the current state of services as well as near-term plans.


jeremyrickard:
This is docs only, CI not needed.

PR #470, opened 2018-06-06
Conditionally Drive Location as Required in Schema

jeremyrickard:
This PR conditionally makes `location` required based on the default location environment variable being set.

Also cleans up some dead code.


pkg/service/catalog.go
```diff
@@ -158,6 +162,7 @@ func NewService(
 	serviceManager ServiceManager,
 	plans ...Plan,
 ) Service {
+	azureConfig, _ := azure.GetConfigFromEnvironment()
```
jeremyrickard: @krancour this seemed like the cleanest way to grab this. I didn't want to bubble this through the rest of the framework and didn't think there was a better way to do this. open to suggestions.

krancour:
Closing this because it will be obsolesced by the plan to stop treating the `location` and `resourceGroup` params like they're special.

PR #471, opened 2018-06-06
Migrate stability to plan instead of module

jeremyrickard:
This PR pushes the stability definition from module to plans. Also changes the filtering to plan level. 

Todo: Update the module docs and README to indicate stability at plan level instead of modules...


pkg/boot/catalog.go
```diff
@@ -42,8 +42,18 @@ func GetCatalog(
...
+			filteredPlans := []service.Plan{}
+			for _, plan := range svc.GetPlans() {
+				if plan.GetStability() >= catalogConfig.MinStability {
+					filteredPlans = append(filteredPlans, plan)
+				}
+			}
+			if len(filteredPlans) > 0 {
+				svc.SetPlans(filteredPlans)
+				services = append(services, svc)
+				usedServiceIDs[serviceID] = moduleName
```
krancour: Prior to this, services and plans were immutable at runtime. I prefer to keep it that way because rationalizing about immutable things is just easier...

I think something like this would work:

```
services = append(
  services,
  NewService(
    svc.GetProperties(),
    svc.GetServiceManager(),
    filteredPlans...,
  ),
)
usedServiceIDs[serviceID] = moduleName
```

jeremyrickard: @krancour updated.

krancour:
I want to merge this _after_ #478

krancour:
Assigning to me because I will do the rebase if required.

krancour:
@jeremyrickard lint error.

jeremyrickard:
@krancour lint fixed.

PR #472, opened 2018-06-11
First pass of documentation changes for GA. Remove experimental

jeremyrickard:
services from README.md, annotate module docs.

### DIFF
--- a/docs/modules/aci.md
+++ b/docs/modules/aci.md
@@ -1,6 +1,6 @@
 # [Azure Container Instances](https://azure.microsoft.com/en-us/services/container-instances/)
 
-_Note: This module is EXPERIMENTAL and future releases may break the API._
+_Note: This module is EXPERIMENTAL and is not included in the General Availability release of Open Service Broker for Azure. It will be added in a future OSBA release._
 
 ## Services & Plans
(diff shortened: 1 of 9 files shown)


docs/modules/aci.md
```diff
@@ -1,6 +1,6 @@
 # [Azure Container Instances](https://azure.microsoft.com/en-us/services/container-instances/)
 
-_Note: This module is EXPERIMENTAL and future releases may break the API._
+_Note: This module is EXPERIMENTAL and is not included in the General Availability release of Open Service Broker for Azure. It will be added in a future OSBA release._
```
carolynvs: I think this should explain what someone should do if they want to install this as I expect we will get questions. Even if the answer is (use this older release until we add this back), that would help avoid getting lots of questions.

jeremyrickard: The answer is use an older version. Personally, I think we should just cull them and then add back as we add the modules. I think it's going to be confusing to have them around but have no way to enable them in the broker. I can add some wording that if you'd like to use it, use an older version of OSBA but I think that looks bad and is kind of confusing. Not sure there is a great outcome here?

PR #475, opened 2018-06-11
Adding use of Redis SET to store instance and binding ids

jeremyrickard:
This will enable us to bulk operate over the set of ids later for
transformation purposes.

Changed write binding and delete binding to use pipelines, then added a step to other method pipelines to add or remove from a set (SAdd, SRem).


krancour:
Nice and clean. Thanks for doing this @jeremyrickard.

Letting CI rip.

PR #478, opened 2018-06-13
Refactor so all params-related things are driven by schema

krancour:
This PR completes the unification of plan specific parameters definition, validation, definition of default values, encryption/decryption (for sensitive params) under one mechanism-- a wrapper of sorts around `map[string]interface{}` that uses schema to inform data access, marshaling/unmarshaling operations, etc.

This is, unfortunately, a necessarily large PR. Many apologies for that. The change had many "ripples," which prompted further refactoring to ensure things would continue working.

Some additional highlights from this PR:

- Comments a recently added assertion that was causing test failures if Redis wasn't reset between test runs. This was a practical measure to allow me to keep moving.
- To limit the size of the refactor to the extent possible, preview and experimental modules have been removed from the build using build tags (and comments wherever entire files couldn't be excluded). @jeremyrickard has already updated docs accordingly in another PR.
- One ripple would have required refactoring of how params are compared to one another when deciding whether an inbound provisioning or updating request is a duplicate of a request already in-progress. This logic was actually flawed to begin with and relates to an open/unresolved issue in the OSB spec. Rather than waste time refactoring a flawed heuristic, I elected to _stop_ doing the thing that made this comparison difficult in the first place-- we've _stopped_ defaulting / auto-generating values for `location` and `resourceGroup` params and, in fact, we have entirely stopped treating those params as if they are special in some way. Lumped in the same category are tags. The bottom line is this allows a straight comparison of inbound params to existing params to determine if a request is logically equivalent to current state or to an operation already in-progress.
- As an entirely necessary validation of this refactored code, I have merged in some e2e tests that were hanging out in another branch for some time. I have _not_ integrated these into our CI process because I don't believe that these tests, in their current form, do an adequate job of reporting what failed, where, and why. One needs to scour logs to make that determination. Until this is fixed, they're an inadequate substitute for our existing service lifecycle tests. At least, however, we can manually kick off e2e tests to receive some additional assurances of code health before a release.

@jeremyrickard let's walk these changes together.


krancour:
Kicking CI. Previous run failed due to what I think may have been a transient network error while verifying vendored code.

krancour:
@jeremyrickard still running into a problem with vendored code. I believe that what has happened is that with many modules (temporarily) disabled, the sum total of what is computer to be required during verification is less than what is actually in the `vendor/` dir currently.

I'm going to run `dep ensure` to update the `vendor/` dir and amend this PR.

PR #479, opened 2018-06-13
Create a new install guide

jeremyrickard:
This PR creates a new install guide and a helm chart that uses etcd-operator to create a new etcd cluster with backup capabilities.


docs/installation-aks.md
```diff
@@ -0,0 +1,407 @@
+# Install and Operate Open Service Broker for Azure on an Azure Container Service managed cluster
+
+Open Service Broker for Azure allows you to provision Azure services from your Kubernetes cluster. OSBA is integrated with Kubernetes using [Service Catalog](https://github.com/kubernetes-incubator/service-catalog). Both Service Catalog and OSBA have data persistence needs. When using Service Catalog and OSBA for development, it is sufficient to use the embedded storage options available with each application. For production use cases, however, we recommend more robust solutions. This guide provides details on setting up OSBA and associated software including Service Catalog, etcd and Redis for production scenarios. If you are new to OSBA, you may find the [AKS](quickstart-aks.md) or [Minikube](quickstart-minikube.md) Quickstart guides useful.
```
carolynvs: nit: May want to break this up into smaller paragraphs, e.g.

```
Open service broker for azure...

Both Service Catalog and OSBA...

If you are new to OSBA, ...
```

docs/installation-aks.md
```diff
@@ -0,0 +1,407 @@
...
+
+* [Prerequisites](#prerequisites)
+  * [Existing Clusters](#existing-cluster)
```
carolynvs: broken link

docs/installation-aks.md
```diff
@@ -0,0 +1,407 @@
...
+  * [Install Helm](#install-helm)
+  * [Install etcd](#install-etcd)
+    * [Create Storage Account](#create-storage-account)
+    * [Install etcd Operator](#install-etcd-operator)
+    * [Create etcd Cluster](#create-etcd-cluster)
+  * [Install Service Catalog](#install-service-catalog)
+* [Create an Azure Redis Cache](#create-azure-redis-cache)
+* [Create a service principal](#create-a-service-principal)
+* [Configure the cluster with Open Service Broker for Azure](#configure-the-cluster-with-open-service-broker-for-azure)
+* [Next Steps](#next-steps)
```
carolynvs: broken link

docs/installation-aks.md
```diff
@@ -0,0 +1,407 @@
...
+
+When you create an AKS cluster, you must provide a resource group. Create one with the az cli using the following command.
+
+```console
+az group create --name aks-group --location eastus
+```
+
+#### Create a Kubernetes cluster using AKS
+
+* As AKS is currently in preview, you will need to enable it in your subscription
```
carolynvs: We should remove this, right?

docs/installation-aks.md
```diff
@@ -0,0 +1,407 @@
...
+* AKS currently does _not_ support RBAC, so we will need to explicity disable that when we install service catalog.
```
carolynvs: This too

docs/installation-aks.md
```diff
@@ -0,0 +1,407 @@
...
+
+1. Enable AKS in your subscription, use the following command with the az cli:
```
carolynvs: And this

docs/installation-aks.md
```diff
@@ -0,0 +1,407 @@
...
+## Cluster Configuration
+
+Before you can install OSBA onto your cluster, you will first need to install Service Catalog. Service Catalog uses etcd for data persistence and it is *strongly* recommended that you create an etcd cluster with backup and recovery capabilities. We recommend using [etcd operator](https://github.com/coreos/etcd-operator) configured with backup and recovery via Azure Blob Storage for this purpose. The following section provides guidance on how to configure etcd operator on your cluster and then how to install Service Catalog.
+
+### Install Helm
+
+We will use Helm to install etcd Operator, Service Catalog and OSBA. In order to install Helm in your cluster, use the `helm` CLI:
+
+```console
+helm init
```
carolynvs: Is this the most safe way to run helm in prod? I'm wondering if since we are kind of saying this is prod safe way to install on aks that we may want to include any other security best practices in this command if necessary.

docs/installation-aks.md
```diff
@@ -0,0 +1,407 @@
...
+
+### Install etcd
+
+#### Create Storage Account
+
+In order to configure etcd operator to use Azure Blob Storage for backup and recovery purposes, you will need to first create a storage account:
+
+
+First, save your desired account name to an environment variable. It will be reused a few times.
+ **Bash**
```
carolynvs: needs a new line before this so that it gets its own paragraph

docs/installation-aks.md
```diff
@@ -0,0 +1,407 @@
...
+
+```console
+helm install --name etcd-operator stable/etcd-operator --values=contrib/k8s/etcd-operator/etcd-operator-values.yaml
+```
+
+This will create three deployments: etcd-operator, etcd-backup-operator, and restore-operator. This will also create three new Custom Resource Definitions: `etcdclusters.etcd.database.coreos.com`, `etcdbackups.etcd.database.coreos.com` and `etcdrestores.etcd.database.coreos.com`. These can be used to create an etcd cluster, along with backup and restore operations. 
+
+#### Create etcd Cluster
+
+Once etcd Operator has been installed, you can create a cluster. For production scenarios, we recommend a three node cluster. For convenience purposes, we have provided a Helm chart that will create a cluster and configure a Kubernetes CronJob to enable automatic backups of the cluster.
```
carolynvs: it would help to say why a 3 node cluste is good, like "blah blah HA blerp bleep"

carolynvs: ~For convenience purposes~

docs/installation-aks.md
```diff
@@ -0,0 +1,407 @@
...
+**PowerShell**
+```console
+az storage container create --name etcd-backups --account-name $env:AZURE_STORAGE_ACCOUNT --account-key $env:AZURE_STORAGE_KEY
+```
+
+#### Install etcd Operator
+
+Service Catalog requires etcd to persist state for everything it manages. We recommend using etcd Operator to install and manage an etcd cluster for Service Catalog. We have included sample values.yaml file to setup etcd operator with a version that supports Azure backup. 
+
+```console
```
carolynvs: super mega nit: For command blocks that just have a command and no `$` and output, don't specify `console` so that the command prints in black. This better matches how the command blocks look when you do use `$` along with `console` so that commands are always in black, and console output is always in blue on the doc page.

it just helps people visually scan the page.

docs/installation-aks.md
```diff
@@ -0,0 +1,407 @@
...
+svc-cat-etcd-client   ClusterIP   10.0.2.237     <none>        2379/TCP            3m
+```
+
+The `svc-cat-etcd-client` service is what you will use to configure Service Catalog.
+
+### Install Service Catalog
+
+Once you have created an etcd cluster, it is time to install Service Catalog. You will use Helm to install Service Catalog. There are a few values you will need to override for your Service Catalog installation:
+
+* RBAC must be disabled (pending AKS support)
```
carolynvs: oops you said that you are doing another pass for rbac, right? I'll stop blathering about rbac then 😇

docs/installation-aks.md
```diff
@@ -0,0 +1,407 @@
...
+* Embedded etcd must be disabled
+* You must point the installation at an external etcd.
```
carolynvs: "external etcd" maybe should be called "an etcd cluster" so that people aren't thinking we are talking about an off-cluster ... etcd cluster.

p.s. cluster cluster cluster

docs/installation-aks.md
```diff
@@ -0,0 +1,407 @@
...
+helm repo add svc-cat https://svc-catalog-charts.storage.googleapis.com
+helm install svc-cat/catalog --name catalog --namespace catalog \
+   --set rbacEnable=false \
+   --set apiserver.storage.etcd.useEmbedded=false \
+   --set apiserver.storage.etcd.servers=http://svc-cat-etcd-client.default.svc.cluster.local:2379
+```
+
+## Create Azure Redis Cache
+
+Open Service Broker for Azure uses Redis as a backing store for its state. We recommend using a managed Redis service, such as Azure Redis Cache. By default, Azure Redis Cache only keeps data in memory. For best results, you will want to use the Premium tier in order to configure backups of the Redis data. Please see [How to configure data persistence for a Premium Azure Redis Cache](https://docs.microsoft.com/en-us/azure/redis-cache/cache-how-to-premium-persistence) for instructions on how to create an Azure Redis Cache. Once created, you can obtain the hostname and keys from the Portal or via the CLI.
```
carolynvs: "We recommend using Azure Redis Cache, a managed Redis service". Let's come right out and say that we commend out own stuff. 😀

carolynvs: "By default, Azure Redis Cache ... For best results, "

As long as we are making recommendations let's be crystal clear

"You should use the Premium tier of Azure Redis Cache, so take advantage of persistent backups because by default Redis is an in-memory only database. See How to configure..."

docs/installation-aks.md
```diff
@@ -0,0 +1,407 @@
...
+spec:
+  etcdCluster:
+    name: svc-cat-etcd
+  backupStorageType: ABS
+  abs:
+    path: <abs continer>/<BACKUP-FILE> 
+    absSecret: <abs-credential-secret-name> 
+```
+
+You will need to replace the spec.abs.path value with the backup you'd like to restore from. You will need the storage container as well as the file name. For example, if your absSecret was called `dandy-clownfish-svc-cat-etcd` and you used etcd-backups as the storage container as directed above and the file name was etcd.backup.2018-06-12_19:31:05, your restore yaml would look like:
```
carolynvs: let's add some backticks around `spec.abs.path` because ... codeblocks, all the beautiful codeblocks.

krancour:
@jeremyrickard do you want this in v0.12.0?

jeremyrickard:
We can do this after the v0.12.0 release, but ideally shortly there after. I’d like to go through this and update for RBAC since AKS is GA now. Shouldn’t be any major changes and will not have any code changes.

docs/installation-aks.md
jeremyrickard: still required :(

docs/installation-aks.md
jeremyrickard: this one is removed.

docs/installation-aks.md
jeremyrickard: still required :(

docs/installation-aks.md
jeremyrickard: I believe the more proper way is to create a service account and create the RBAC policies. We do this in the quickstarts now, so i have updated this. I'll ping Matt F and see if he has any more feedback.

Makefile
```diff
@@ -349,6 +352,7 @@ PUBLISH_RC_CHART_CMD := bash -c ' \
 	&& mkdir repo \
 	&& cd repo \
 	&& sed -i "s/^version:.*/version: 0.0.1-$(GIT_VERSION)/g" ../open-service-broker-azure/Chart.yaml \
+	&& sed -i "s/^version:.*/version: 0.0.1-$(GIT_VERSION)/g" ../svc-cat-etcd/Chart.yaml \
```
krancour: Do we want the version of the etcd operator chart tied explicitly to the version of OSBA? I'm thinking not, but am willing to be convinced otherwise.

Is it possible that `Azure/helm-charts` is a better home for this new chart?

The new install docs, however, do clearly belong in this repo. No question there.

jeremyrickard: >Do we want the version of the etcd operator chart tied explicitly to the version of OSBA? I'm thinking not, but am willing to be convinced otherwise.

It's not really the etcd operator chart, it's the chart that creates the cluster _for_ service ctalog...but that doesn't change the fact that the answer is _probably not_. 

> Is it possible that Azure/helm-charts is a better home for this new chart?

Possible. I thought we had kind of turned that repo into something more akin to just examples. I could go either way.

krancour: > I thought we had kind of turned that repo into something more akin to just examples.

Ya... I've been saying that, haven't I? Maybe I wasn't saying what I meant...

All I wanted to avoid was indiscriminately filling that repo with lots of forks of upstream charts that create additional burden without adding a lot of value. The highest value things sort of ended up being demonstrations of services that we didn't already have demos for-- i.e. examples, as you say.

idk if examples need to be the _only_ use for that repo, though.

In my mind, the best place to keep chart source is close to the app source. (That's why the OSBA chart is here in this repo.) In case, there is no app source (so to speak). The chart just contains resources managed by an operator. That makes me feel a little bit like it's out of place here.

There is, perhaps, an even stronger argument to be made, however, by acknowledging that the etcd this creates is only _indirectly_ a prerequisite of OSBA. The thing most proximately dependent on it is svc cat...

That being the case, I wonder if we should be trying to upstream this to svc cat instead. It may even be that _their_ documentation should be going over "here's how to run a decent etcd in production and here's a chart for it" instead of ours doing that. Perhaps ours should just say "follow the directions for setting up production-grade svc cat."

Thoughts?

jeremyrickard: The only issue I have with trying to upstream this is that it's _very_ Azure specific. AFAIK, the operator only supports AWS and Azure for backup/recovery, so it would be of limited utility I think to the community in general. Happy to open it and see what happens.

PR #480, opened 2018-06-15
Fixing MSSQL update

jeremyrickard:
* Neglected to split common "dbms" params between update and provision
   - Location and RG were required for update
* Bad parameter use on all in one update


krancour:
@jeremyrickard do you want this in before or after #481?

pkg/services/mssql/plan_details.go
```diff
@@ -53,7 +53,23 @@ func (d dtuPlanDetails) getUpdateSchema() service.InputParametersSchema {
...
+	// option
+	if len(d.allowedDTUs) > 0 {
+		ips.PropertySchemas["dtus"] = &service.IntPropertySchema{
+			AllowedValues: d.allowedDTUs,
+			DefaultValue:  ptr.ToInt64(d.defaultDTUs),
+			Description: "DTUs are a bundled measure of compute, " +
+				"storage, and IO resources.",
+		}
+	}
+	return ips
```
jeremyrickard: Could probably do some DRY refactor here, but I feel like explicitly repeating this is a little more readable.

krancour:
Deferring to @arschles's LGTM. Merging.

PR #481, opened 2018-06-15
refactor codecs and details

krancour:
This is the last major bit of refactoring that I want to accomplish.

This finally consolidates all unsecured instance details with their secure complements by introducing a new type `service.SecureString` that is seamlessly encrypted during marshaling and decrypted during unmarshaling.

Making this work required that the `service.SecureString` could pluck a codec "from thin air" at the time of marshaling/unmarshaling. To facilitate that, there is now a "global" codec configured in the `crypto` package. All other uses of codecs have been modified to use that as well.

This work also exposed a bug in the `Params` type that is used for managing user input, so it is also fixed in this PR.


pkg/api/init_test.go
```diff
@@ -0,0 +1,17 @@
...
+
+	"github.com/Azure/open-service-broker-azure/pkg/crypto"
+	"github.com/Azure/open-service-broker-azure/pkg/crypto/noop"
+)
+
+func TestMain(m *testing.M) {
+	if err := crypto.InitializeGlobalCodec(noop.NewCodec()); err != nil {
+		log.Fatal(err)
+	}
+	os.Exit(m.Run())
```
arschles: nit: maybe worth using testify suite sometime later on

PR #482, opened 2018-06-15
make the project layout more natual. 

andyliuliming:
so we can use direnv setup the GOPATH automatically.


krancour:
@andyliuliming I can't see any good reason to adopt these changes, to be honest.

I'm going to cite a few reasons.

We've taken great pains to make sure that our entire workflow is containerized. You actually do _not_ need Go setup on your machine _at all_ (nevermind `GOPATH` being configured correctly) in order to build or test OSBA. When you type something like `make test-unit` or `make build`, everything that follows happens in a container that is already configured correctly. fwiw, you do this because it guarantees a consistent development environment for all contributors and also guarantees consistency between contributors and CI. We _never_ run into "it worked on my machine" in this project because of these choices.

Containerized tooling is great for discreet tasks like building or testing, but I can see where it falls short for something like Go integration with your IDE. We ourselves use VS Code and the fact that all dependencies are vendored really minimizes the role of the `GOPATH` in your IDE's ability to build the code. In the end, any dependencies that are globally installed on your GOPATH should be of no consequence to OSBA.

In the end, I'm not sure what this PR ends up accomplishing other than shortening some import paths (which we never really have to type anyway... because tooling). On top of that, wouldn't this essentially _require_ `direnv` (i.e. another tool) to contribute to OSBA?

I'm going to close this, but I am happy to hear any rebuttal if you feel that I'm not seeing this clearly and that our way of doing things poses some serious problems that I am not recognizing yet.

andyliuliming:
the benifit of the pr is it does not require the $GOPATH/src/github.com/open-service-broker-azure path.
the project can be cloned to any path.

PR #485, opened 2018-06-15
Update AKS quickstart to use RBAC on the new cluster.

jeremyrickard:
AKS supports RBAC now, so I've updated the AKS Quickstart to reflect this.


docs/quickstart-aks.md
```diff
@@ -167,14 +164,14 @@ You should also ensure that the `Microsoft.Compute` and `Microsoft.Network` prov
...
+    az aks create --resource-group aks-group --name osba-quickstart-cluster --generate-ssh-keys --kubernetes-version 1.9.6 --enable-rbac
     ```
 
     Note: Service Catalog may not work with Kubernetes versions less than 1.9.0. If you are attempting to use an older AKS cluster, you will need to upgrade. The earliest 1.9.x release available from AKS is 1.9.1, so you will need to upgrade to at least that version.
 
-1. Configure kubectl to use the new cluster
+1. Configure kubectl to use the new cluster. You'll initially want to use the cluster admin user to configure Helm and Service Catalog.
     ```console
-    az aks get-credentials --resource-group aks-group --name osba-quickstart-cluster
+    az aks get-credentials --resource-group aks-group --name osba-quickstart-cluster --admin
```
carolynvs: 👍

PR #492, opened 2018-06-22
Update services and plans to be immutable

### LINKED ISSUE #491: Duplicate fields in some json schema
There is some duplication in some of the JSON schema required fields entries. This is invalid JSON schema and prevents the broker from working with CF.

jeremyrickard:
This moves Services and Plans to immutability. 

Fixes #491

### DIFF
--- a/pkg/service/catalog.go
+++ b/pkg/service/catalog.go
@@ -155,8 +155,7 @@ func NewService(
 		plans:             plans,
 		indexedPlans:      make(map[string]Plan),
 	}
-	for _, planIfc := range s.plans {
-		p := planIfc.(plan)
+	for _, p := range s.plans {
 		s.indexedPlans[p.GetID()] = p
 	}
 	return s
--- a/pkg/service/catalog.go
+++ b/pkg/service/catalog.go
@@ -61,12 +61,12 @@ type Service interface {
 	SetPlans([]Plan)
 	GetParentServiceID() string
 	GetChildServiceID() string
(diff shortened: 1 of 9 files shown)


pkg/service/catalog.go
```diff
@@ -17,7 +17,7 @@ type catalog struct {
 	Services        []json.RawMessage `json:"services"`
 	services        []Service
 	indexedServices map[string]Service
-	jsonMutex       sync.Mutex
+	jsonMutex       *sync.Mutex
```
jeremyrickard: I made this change because when I refactored the method receivers to not be pointers i got lots of complaints about passing around the Mutex copy

pkg/service/catalog.go
```diff
@@ -157,160 +145,113 @@ func (c *catalog) GetService(serviceID string) (Service, bool) {
...
+	s := service{
 		ServiceProperties: serviceProperties,
 		serviceManager:    serviceManager,
 		plans:             plans,
 		indexedPlans:      make(map[string]Plan),
 	}
 	for _, planIfc := range s.plans {
-		p := planIfc.(*plan)
-		p.Schemas.addCommonSchema(serviceProperties)
+		p := planIfc.(plan)
```
krancour: Shouldn't have to make this type assertion here. Just add `planIfc` to the indexed plans instead. Better yet, just change `for _, planIfc := range s.plans {` to `for _, p := range s.plans {`.

Issue #428, opened 2018-05-04
Clean up OpenShift template


krancour:
@jimzim see post-merge comments on #425

PR #494, opened 2018-06-23
Rename CRYPTO_AES256_KEY to AES256_KEY in OpenShift template

jaredhocutt:
The environment variable in the OpenShift template was named `CRYPTO_AES256_KEY`, which was causing the deployment of `osba-open-service-broker-azure` to fail.

Per the code this variable should be named `AES256_KEY`:
https://github.com/Azure/open-service-broker-azure/blob/be79faa48d04794538100f3b7b6fc9a40ee04c91/pkg/crypto/aes256/config.go#L12

### DIFF
--- a/contrib/openshift/osba-os-template.yaml
+++ b/contrib/openshift/osba-os-template.yaml
@@ -171,7 +171,7 @@ objects:
                 secretKeyRef:
                   name: osba-open-service-broker-azure
                   key: redis-password
-            - name: CRYPTO_AES256_KEY
+            - name: AES256_KEY
               valueFrom:
                 secretKeyRef:
                   name: osba-open-service-broker-azure


krancour:
Either of those two environment variables should work. Note the const `envconfigPrefix`, its value of `"CRYPTO"` and how it is used when the environment variables are processed.

jaredhocutt:
When deploying it yesterday (2018-06-23) it would not deploy and the logs were showing the following error:

`AES256 key was not specified`

By changing the environment variable from `CRYPTO_AES256_KEY` to `AES256_KEY` the error went away and the deployment was successful.

krancour:
@jaredhocutt can you confirm the version of the image that you were using?

krancour:
@jaredhocutt also note that the Helm chart for v1.0.0 uses the `CRYPTO_AES256_KEY` env var without any issue.

https://github.com/Azure/open-service-broker-azure/blob/master/contrib/k8s/charts/open-service-broker-azure/templates/deployment.yaml#L58-L62

jaredhocutt:
@krancour I'm using `microsoft/azure-service-broker:v0.10.0` per the OpenShift template. I also confirmed in OpenShift that this is in fact the version being used as well.

https://github.com/Azure/open-service-broker-azure/blob/f0ba3bd6f702110e93e22780611feba9f3eb5477/contrib/openshift/osba-os-template.yaml#L151

krancour:
So I am wondering if you were actually, unknowingly, using image  version `v0.10.0`, which seems to be hard-coded into the Open Shift template. (Clearly that's wrong.)

There's already an open issue to fix the aspects of that template that were hardcoded and probably shouldn't have been. See #428 and post-merge comments on PR #425.

krancour:
> But as it is now, the template does not work and this PR fixes that.

That's not quite true. This would "fix" that in master (where there actually isn't even a problem), but that template for deploying on Open Shift would still reference v0.10.0-- which wouldn't have the fix.

imo, short of addressing #425 / #428 completely, a more productive PR might be to simply update the existing template to reference OSBA v1.0.0. (Although, fyi, we will likely release a v1.0.1 on Monday.)

krancour:
Sorry... I stand corrected on the above. I mistakenly thought this PR changed Go code, but see now that it changes the template. Still... if the template were to change, the change I'd rather see is an update to reference a newer broker image. Hope that makes sense.

jaredhocutt:
@krancour That works for me.

I'll wait until after `v1.0.1` is released on Monday and will test to make sure that solves the issue. Then I'll update the PR with that change instead.

krancour:
@jaredhocutt go ahead and amend this commit as we discussed and I will simply merge it just _before_ the release.

krancour:
@jaredhocutt actually... we're ready to go, so if you don't mind I'm just going to close this PR and PR what we discussed.

jaredhocutt:
All good @krancour.

Thanks for getting this fixed. Apologies for the delay, I was travelling today for work and am just now getting caught up.

Issue #493, opened 2018-06-22
Developer Documentation for using broker-cli is out of date


ageekymonk:
Hi, 

  The documentation for how to locally develop  and interact with the broker using **broker-cli** is out of date.

krancour:
That seems as if it's probably true. Can @ageekymonk, can you please help us out a bit by pinpointing where in the documentation things started not working for you?

krancour:
Believe this was fixed by #506. Please re-open if I've got that wrong.

PR #506, opened 2018-06-27
Fixing the build for broker-cli and updating the documentation.

ageekymonk:
Hi, 
The broker-cli to use to test the local development osba build is broken. Fixing it. 
Let me know if you need to make any changes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -218,7 +218,7 @@ else
 	docker-compose kill test-api-compliance-broker
 	docker-compose rm -f test-api-compliance-broker
 endif
-	
+
 LINT_CMD := gometalinter ./... \
 	--concurrency=1 \
 	--disable-all \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -422,7 +422,7 @@ endif
 CONTRIB_BINARY_DIR := contrib/bin
 CLI_BINARY_NAME := broker-cli
 
-BUILD_CMD := go build -o $(CONTRIB_BINARY_DIR)/$(CLI_BINARY_NAME) ./cmd/cli
+BUILD_CMD := go build -o $(CONTRIB_BINARY_DIR)/$(CLI_BINARY_NAME) ./contrib/cmd/cli
```
krancour: Ah... I see what happened here. iirc, there used to be a separate `Makefile` under `contrib`. When they were consolidated, it would appear this path (and those below) weren't adjusted accordingly.

It's gone unnoticed because we rely more often on an e2e test suite these days than the CLI and when I use the CLI (I think I am one of the few contributors who does), I tend to ad-hoc compile with `go run contrib/cmd/cli/*.go ...`

This is a good catch. Thank you for fixing!

krancour:
@ageekymonk does this fix #493?

PR #508, opened 2018-06-28
Enabling Redis

ageekymonk:
Hi, 
I enabled redis. I also tested provision, deprovision and bind. They seem to be working. Let me know if i need to change anything.

### DIFF
--- a/pkg/services/rediscache/arm_template.go
+++ b/pkg/services/rediscache/arm_template.go
@@ -1,112 +1,40 @@
-// +build experimental
-
 package rediscache
 
 // nolint: lll
 var armTemplateBytes = []byte(`
 {
-	"$schema": "http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json",
+	"$schema": "http://schema.management.azure.com/schemas/2014-04-01-preview/deploymentTemplate.json",
 	"contentVersion": "1.0.0.0",
 	"parameters": {
-		"location": {
-			"type": "string"
-		},
-		"serverName": {
-			"type": "string",
-			"metadata": {
-				"description": "The name of the Azure Redis Cache to create."
-			}
-		},
-		"redisCacheSKU": {
-			"type": "string",
-			"allowedValues": [
-				"Basic",
-				"Standard",
-				"Premium"
-			],
-			"metadata": {
-				"description": "The pricing tier of the new Azure Redis Cache."
-			}
-		},
-		"redisCacheFamily": {
-			"type": "string",
-			"metadata": {
-				"description": "The family for the sku."
-			},
-			"allowedValues": [
-				"C",
-				"P"
-			]
-		},
-		"redisCacheCapacity": {
-			"type": "int",
-			"allowedValues": [
-				0,
-				1,
-				2,
-				3,
-				4,
-				5,
-				6
-			],
-			"metadata": {
-				"description": "The size of the new Azure Redis Cache instance. "
-			}
-		},
-		"diagnosticsStatus": {
-			"type": "string",
-			"defaultValue": "OFF",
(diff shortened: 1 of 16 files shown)


pkg/services/rediscache/arm_template.go
```diff
@@ -1,112 +1,40 @@
-// +build experimental
-
 package rediscache
 
 // nolint: lll
 var armTemplateBytes = []byte(`
 {
-	"$schema": "http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json",
+	"$schema": "http://schema.management.azure.com/schemas/2014-04-01-preview/deploymentTemplate.json",
```
krancour: Not clear on why this PR reverts this ARM template to use an older version of the management API for Redis...

pkg/services/rediscache/arm_template.go
```diff
@@ -1,112 +1,40 @@
...
 		"tags": {
 			"type": "object"
 		}
 	},
 	"resources": [
 		{
-			"apiVersion": "2015-08-01",
-			"name": "[parameters('serverName')]",
+			"apiVersion": "2018-03-01",
+			"name": "{{.serverName}}",
```
krancour: +1 on using Go template params instead of ARM params wherever possible.

pkg/services/rediscache/provision.go
```diff
@@ -22,67 +20,64 @@ func (s *serviceManager) GetProvisioner(
...
-	dt := instanceDetails{
+) (service.InstanceDetails, error) {
+	dt := &instanceDetails{
 		ARMDeploymentName: uuid.NewV4().String(),
 		ServerName:        uuid.NewV4().String(),
 	}
-	dtMap, err := service.GetMapFromStruct(dt)
-	return dtMap, nil, err
+
+	return dt, nil
```
krancour: Nit: Let's return a struct literal here instead of assigning to `dt`.

tests/e2e/test_cases_test.go
```diff
@@ -10,7 +10,7 @@ import (
 
 func getTestCases() ([]e2eTestCase, error) {
 	testCases := getPostgreSQLTestCases()
-	// testCases = append(testCases, rediscacheTestCases...)
+	testCases = append(testCases, rediscacheTestCases...)
```
krancour: Please make this the _first_ group of test cases added to `testCases`. Redis takes longer than anything else we have to provision and deprovision, so I want to assurance that regardless of how serialized or parallelized the tests may be, the Redis cases are always kicked off first.

krancour:
@ageekymonk thanks for a solid PR. I made a few minor comments, but this mostly looks really good.

My biggest open question has to do with the change in API version in the ARM template.

Also, note that there are a few lint errors.

pkg/services/rediscache/arm_template.go
ageekymonk: I used portal to generate the arm template and it had only this version and i checked few other modules and they had the same. So i thought it is the correct version.

pkg/services/rediscache/provision.go
ageekymonk: Updated it.

tests/e2e/test_cases_test.go
ageekymonk: I have moved redis to the top.

jeremyrickard:
Closes #513

krancour:
LGTM. Note that this seems to remove some diagnostic stuff from the ARM template, but I'm ok with that because we never willfully added that feature in the first place. I'm about 99% sure its existence was mere copy/pasta from some Redis ARM template example.

krancour:
@ageekymonk thanks for the help!

PR #524, opened 2018-07-06
Add TLS support to the broker

### LINKED ISSUE #225: Use TLS
Because this is likely soon a requirement that auth be done over TLS and because it's a good idea, in general, the broker needs to support using TLS to secure inbound API calls.

Our Helm chart in Azure/helm-charts should generate a certificate (possible since Helm 2.7) and enable this option.

krancour:
Fixes #225

Highlights:

- Improves DI / bootstrapping to make API server easier to configure
- Introduces new configuration object for API server
- Automatically uses TLS cert and key if found
- Helm chart automatically generates CA and cert signed by that CA, unless you opt out
- Client gets SSL support-- including option to skip cert validation

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,4 +13,3 @@ ARG BASE_PACKAGE_NAME
 COPY --from=0 /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
 COPY --from=0 /go/src/$BASE_PACKAGE_NAME/bin/broker /app/broker
 CMD ["/app/broker"]
-EXPOSE 8080
--- a/cmd/broker/broker.go
+++ b/cmd/broker/broker.go
@@ -178,7 +178,7 @@ func main() {
 	}()
 
 	// Run broker
-	if err := broker.Start(ctx); err != nil {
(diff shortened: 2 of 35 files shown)


Dockerfile
```diff
@@ -13,4 +13,3 @@ ARG BASE_PACKAGE_NAME
 COPY --from=0 /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
 COPY --from=0 /go/src/$BASE_PACKAGE_NAME/bin/broker /app/broker
 CMD ["/app/broker"]
-EXPOSE 8080
```
krancour: May wonder why this line is going away...

This PR adds the flexibility for the broker to run on _any_ port, dependent on the value of the `API_SERVER_PORT` env var. (The chart will pick 8080 or 8443 depending on if TLS is being used or not-- and really those are just for vanity.) But since the software on the Docker image can use any port, there's no point in keeping the directive `EXPOSE 8080`, because according to Docker's own documentation, that directive doesn't have any real effect at runtime and only serves to document expectations between image authors and people who run containers.

jeremyrickard: 👍

PR #528, opened 2018-07-12
Remove Search Module

### LINKED ISSUE #527: Remove Search Module
We will no longer offer the Search module, so we need to remove it from the code base.

jeremyrickard:
This removes the search module

Closes #527


jeremyrickard:
@seanmck removing search module per Slack.

PR #532, opened 2018-07-15
fix var name

haroldwongms:
Container is looking for var name of AES256_KEY.


krancour:
See #494 to understand why this shouldn't be necessary and to find a probable explanation of how you might have ended up in circumstances that may have led you to believe it's necessary.

Long story short, use the latest official chart and it should already be compatible with the latest official image.

haroldwongms:
@krancour Thank you.  Weird.  I tried deploying last week and got that error.  I went back and looked at my cluster and it was using version 1.0.1 so not sure how that error cropped up.  Must be the gremlins 'cause it is working just fine now.

Closing.

PR #544, opened 2018-08-09
Enable Configuring read regions and auto failover in CosmosDB

norshtein:
Hi, this pr enables configuring read regions and auto failover in CosmosDB in following services:
- azure-cosmosdb-sql
- azure-cosmosdb-sql-account
- azure-cosmosdb-mongo-account
- azure-cosmosdb-table-account
- azure-cosmosdb-graph-account

In the provision and update step in these services, I add two parameter: `readRegions` and `autoFailoverEnabled`, customers can use these two parameters to manage geo-replication of the cosmos account. 
Document has been added. Life cycle test case has been added. Feel free to contact me if you have any question.


pkg/services/cosmosdb/cosmos-utils.go
```diff
@@ -178,3 +182,218 @@ func deleteDatabase(
...
+				fmt.Sprintf("given region %s is not allowed", region),
+			)
+		}
+	}
+	return nil
+}
+
+// Allowed CosmosDB read locations, it is different from Azure regions.
+// We use a map here to record all allowed regions.
+var allowedReadLocations = map[string]bool{
```
jeremyrickard: Is there an API we can fetch this from instead of hard coding the list?

Also, do customers typically have access to euap reagions?

pkg/services/cosmosdb/cosmos-utils.go
```diff
@@ -178,3 +182,218 @@ func deleteDatabase(
...
+func (c *cosmosAccountManager) waitForReadLocationsReady(
+	ctx context.Context,
+	instance service.Instance,
+) (service.InstanceDetails, error) {
+	dt := instance.Details.(*cosmosdbInstanceDetails)
+	resourceGroupName := instance.ProvisioningParameters.GetString("resourceGroup")
+	accountName := dt.DatabaseAccountName
+	databaseAccountClient := c.databaseAccountsClient
+
+	err := pollingUntilReadLocationsReady(
```
jeremyrickard: Instead of polling this, could we instead make use of the async engine like we do waiting for parent/child to be created?

pkg/services/cosmosdb/cosmos-utils.go
```diff
@@ -178,3 +182,218 @@ func deleteDatabase(
...
+				resourceGroupName,
+				accountName,
+			)
+			if err != nil {
+				return err
+			}
+
+			//Check whether every read location's state is "Succeeded"
+			allSucceed := true
+			readLocations := *(result.DatabaseAccountProperties.ReadLocations)
```
jeremyrickard: we do something similar when we check for parent/child relationships with the async engine. I'd prefer if we used that here as well, vs polling and blocking in this.

pkg/services/cosmosdb/cosmos-utils.go
norshtein: I have gone over the [spec](https://github.com/Azure/azure-rest-api-specs/tree/master/specification/cosmos-db/resource-manager/Microsoft.DocumentDB/stable/2015-04-08) provided by CosmosDB team and didn't find any API which can provide the read regions list. So we have to hard code it temporarily. And most customers don't have access to euap regions, we may delete all euap regions from the list.

pkg/services/cosmosdb/cosmos-utils.go
norshtein: Seems this case is different from checking for parent/child relationships. Checking for parent/child relationships is a general operation and it is done before the first provision step is executed. It's reasonable to make checking a `delayedTask` and trigger the real executing task after the checking is completed. But here, this function itself will be executed inside an async task. Will it be a little strange if we make it a `delayedTask` and manipulate async engine directly here?

norshtein: Hi @jeremyrickard , maybe you are worry about `polling`? Actually this function works like [WaitForCompletion](https://github.com/Azure/open-service-broker-azure/blob/master/pkg/services/cosmosdb/common_deprovision.go#L44) in deprovision, it is executed in the go routine and won't block any other operation.

norshtein: Hi @jeremyrickard , any follow-up on this topic?

jeremyrickard: @norshtein I'm occupied with other things today so I probably won't get a chance to circle back to this until Monday, my apologies! I'll revisit this first thing in the morning on Monday!

krancour: @norshtein @jeremyrickard I think the async stuff is being approached reasonably here. `waitForReadLocationsReady` is registered as a provisioning/update step and the general requirements for those is that they be idempotent and interruptable, and those conditions seem to be met here. `waitForReadLocationsReady` calls `pollingUntilReadLocationsReady` and that is interruptable and it seems that if the the application process died while that was executing, it would seamlessly be resumed by the async engine (like any other provisioning step) and would eventually succeed.

I do have other issues with the implementation here, but this isn't one of them. I wouldn't worry about this particular detail. I'll comment on the other issues closer to where they appear in the code...

pkg/services/cosmosdb/cosmos-utils.go
```diff
@@ -178,3 +182,216 @@ func deleteDatabase(
...
+
+const succeeded = "succeeded"
+
+// For now, this method will return on either context is cancelled or
+// every region's state is "succeeded" in seven consecutive check.
+// The reason why we need seven consecutive check is that
+// the read region is created one by one, there is a small gap between
+// the finishment of previous creation and the start of the next creation.
+// By this check, we can detect gaps shorter than 1 mintue,
+// and report success within 70 seconds after completion.
```
krancour: This heuristic will not work.

It seems implied here that read locations are added and become ready one-by-one (i.e. in serial).

1. Is that even true? That seems _really_ odd. My gut would expect that `result.DatabaseAccountProperties.ReadLocations` consistently reports the same number of read locations, regardless of their state. Shouldn't we just expect that if n locations were requested that each time we poll, some m =< n will be ready and some n - m will _not_ be ready? If that's the case, shouldn't we only have to poll until all n are reported as ready _just once_ ? (Maybe this isn't how this API works?? Correct me if I am wrong.)

2. Let's assume I am totally wrong about no.1 above. So let's say that `result.DatabaseAccountProperties.ReadLocations` does _not_ consistently report the same number of read locations, regardless of their state. In such a case, don't we only need to wait until read regions m == desired read regions n _and_ all n are ready?

Which of these is correct depends a lot on the mechanics of the underlying API, and I'm not an expert on that... but whichever is the case, one of the two above will prove to be a logically correct algorithm that is better than the heuristic currently used here... and either will return faster in the successful case than this heuristic would.

Additionally-- some concession probably needs to be made for a timeout here. There should be a finite amount of time we will wait for all read locations to be ready before giving up.

norshtein: Hi @krancour , `result.DatabaseAccountProperties.ReadLocations` works in following way: 
Say I provision an instance with one write location `A` and three read locations `B,C,D`. At first, when `A` is creating, the function will return only one read location, `A` with state "creating'; when the creation of `A` is finished, creation of `B` will start, now the function will return two read locations, `A` with state "succeeded" and `B` with state "creating"; when the creation of `B` is finished, creation of `C` will start, now it will returns three locations, `A` and `B` with state "succeeded" and `C` with state `creating`, and so on.
You are right, it's better to check the the number of read locations and their state. I'm trapped in conventional thinking and it's actually not a good design.

As for a timeout, it has a very wide range based on how many read regions the user creates. How about estimate an upper bound based on the number of read regions? Say the user creates 3 regions, then we set the time out to `3 * 15 * time.Minutes`?

pkg/services/cosmosdb/cosmos-utils.go
norshtein: Hi @krancour , if we change the implementation of `waitForReadLocationsReady` as discussed below, the function itself will be idempotent and interruptable. If the process died for some reason, when it is restarted, it is still able to report success when all regions are created.

Besides, I'm a little confused how does the async engine deals with the process termination. I'm guessing that when the application is restarted, the async engine will find tasks in the task queue and rerun them. If I guessed it right, then the previous implementation should also be idempotent and interruptable. `waitForReadLocationsReady` will be rerun, invoking `pollingUntilReadLocationsReady` function, waiting until all read locations ready. Or my guess is totally wrong and the async engine doesn't work in this way...

jeremyrickard: @norshtein You're correct the current implementation would be. My apologies for the confusion.

pkg/services/cosmosdb/cosmos-utils.go
krancour: That seems like a huge amount of time to wait for something like this to complete. Does it really take that long? Is there no way to bring up the read regions in parallel?

norshtein: Hi @krancour , for now, we can't make the creation in parallel. The underlying creation is managed by Cosmos team, currently the creation is serial on both Azure portal and go-sdk.

norshtein: And 15 mintues for one read location is a pessimistic estimation, I will investigate it and find a more precise bound.

norshtein:
A summary of today's commit:
- In `pollingUntilReadLocationsReady`, change the timeout for each read region from 15 mintues to 7 minutes. It is observed that the creation of an empty read region takes about 5 minutes.
- In `pollingUntilReadLocationsReady`, disable timeout in updating step, it is because updating may cause copying data from existing region to new region and it's unpredictable how long the copy process will take.
- Truncate the id of read regions to 50 characters.

norshtein:
Hi @krancour and @jeremyrickard , could you please have a look?

krancour:
Something I'm not quite understanding about this PR is why there's so much complexity in the update process.

It seems that with provisioning, the steps are deploy the ARM template, then wait for read regions to be ready, but with updating, it seems you're not initially using ARM (using a Cosmos management client instead), waiting for read regions to be ready, and _then_ updating the ARM template.

This seems quite odd to me because typically (OSBA or otherwise) you shouldn't mutate Azure resources that were created using ARM through other means. You should just update the deployment. That's how all OSBA updates work, up until now. They just update the existing ARM deployment.

So what I'm trying to understand is the rationale behind this more complex update process.

norshtein:
Hi @krancour , the process of update is: update read regions, waiting for read regions to be ready and then update other properties. Both of the first step and the third step are done using ARM. The reason is mentioned [here](https://github.com/Azure/open-service-broker-azure/blob/abbfa097e02ed76e90dd7e4b258e9fa53a48d0c0/pkg/services/cosmosdb/graph-api-update.go#L23). The cosmosDB has a really strange contraint: it cannot update properties and
add/remove regions at the same time. So we have to update read regions firstly, keeping other properties the same as in provision, waiting for read regions ready, and then update other properties.

norshtein:
Any update on this PR?

norshtein:
Hi, I have finished this feature(ideally), could you please have a look on it?

And I think there are two items worth considering:
- Due to the reason claimed [here](https://github.com/norshtein/open-service-broker-azure/blob/cosmos-pr-clean-commit/pkg/services/cosmosdb/cosmos-utils.go#L255-L257), there is no time limit for polling in update step. Can we set a very long time limit for it?
- It is designed that the user is not allowed to fill provision parameter `location` in the `readRegions` array as claimed [here](https://github.com/norshtein/open-service-broker-azure/blob/cosmos-pr-clean-commit/docs/modules/cosmosdb.md#about-read-regions). But the validation requires both `location` and `readRegions` so that it can't be done using property schema. In updating step, it can be done using `ValidateUpdatingParameters`. But currently OSBA don't have `ValidateProvisioningParameters` method, so validating it in provisioning step requires an extra provisioning step. As far as I'm concerned, it's neither well-designed nor user-friendly to add an extra provisioning step. So my current implementation is ignoring `location` in `readRegions` array. That is, if the user fill provision parameter `location` in the `readRegions` array by mistake, the result is the same as removing `location` from the `readRegions` array.

How do you think about these two items?

campbelldgunn:
What is the latest on this? I am very interested in COSMOSDB hitting preview. But want to make sure it is performant and interops well with OSBA.

norshtein:
Hi @campbelldgunn , this feature is almost finished and my concern is the two items mentioned above. And it should work well if users follow the document and no unexpected exception happens in updating.

Hey @jeremyrickard and @krancour , do you have any suggestion on these two items? Also, please let me know if you have any concern blocking the merge.

pkg/services/cosmosdb/plan_schemas.go
```diff
@@ -16,6 +16,19 @@ func generateUpdatingParamsSchema() service.InputParametersSchema {
...
 	return service.InputParametersSchema{
 		PropertySchemas: map[string]service.PropertySchema{
+			"readRegions": &service.ArrayPropertySchema{
+				Description: "Read regions to be created, your data" +
+					" will be synchronized across these regions.",
+				DefaultValue:            []interface{}{},
+				CustomPropertyValidator: readLocationsValidator,
+			},
+			"autoFailoverEnabled": &service.StringPropertySchema{
+				Description: "Specifies if you want Cosmos DB to perform" +
```
zhongyi-zhang: Property titles are missing.

pkg/services/cosmosdb/graph-api-update.go
```diff
@@ -7,26 +7,52 @@ import (
...
+		instance.ProvisioningParameters,
+		instance.UpdatingParameters,
+		instance.Details.(*cosmosdbInstanceDetails),
+		"GlobalDocumentDB",
+		"EnableGremlin",
+		map[string]string{
+			"defaultExperience": "Graph",
+		},
+	); err != nil {
+		return nil, fmt.Errorf("error deploying ARM template: %s", err)
```
zhongyi-zhang: Is "error deploying ARM template" already appended in cosmosAccountManager.updateReadLocations? Same problem in other service managers.

pkg/services/cosmosdb/cosmos-utils.go
```diff
@@ -178,3 +182,308 @@ func deleteDatabase(
...
+// may be provided, in which case will cause bug in updating.
+// To avoid this case, we truncate the id of read locations to 50 characters.
+// It is tested that truncating the read region id won't affect user's usage.
+func generateIDForReadLocation(
+	databaseAccountName string,
+	location string,
+) string {
+	locationID := fmt.Sprintf("%s-%s", databaseAccountName, location)
+	if len(locationID) > 50 {
+		locationID = locationID[0:50]
```
zhongyi-zhang: There is a simliar function `generateAccountName` in common-provision.go. Does it make more sense to follow a similar implementation, or just extend it and reuse?

norshtein: The reason is this function is invoked in both provisioning and updating, and if the region is not changed, we must guarantee the id is not changed. `generateAccountName ` will generate a new name every time it is invoked.

norshtein:
There is a constraint in CosmosDB: it can't update `readRegions` and other properties at the same time. And my previous implementation is using complex logic to enable updating `readRegions` and other properties at the same time. After consideration, I decide to remove this logic. I think it's a good practice to leave as it is rather than using complex logic to enable it in the broker

pkg/services/cosmosdb/common-provision.go
```diff
@@ -49,53 +49,41 @@ func (c *cosmosAccountManager) buildGoTemplateParams(
...
-			"139.217.8.252",
-			"52.244.48.71",
-		)
-	} else {
-		filters = append(filters, "0.0.0.0")
-	}
-	filters = append(filters, ipFilters.GetStringArray("allowedIPRanges")...)
-	if len(filters) > 0 {
-		p["ipFilters"] = strings.Join(filters, ",")
+	readLocations := pp.GetStringArray("readRegions")
```
zhongyi-zhang: I think we should add a comment to explain that it is a difference between REST API naming and Portal naming.

norshtein: Will add, thanks.

pkg/services/cosmosdb/common_update.go
```diff
@@ -1,11 +1,47 @@
...
+	ppData := make(map[string]interface{})
+	upData := make(map[string]interface{})
+	for k, v := range pp.Data {
+		ppData[k] = v
+	}
+	for k, v := range up.Data {
+		upData[k] = v
+	}
+	ppData["readRegions"] = nil
+	upData["readRegions"] = nil
```
zhongyi-zhang: Using built-in function delete(upData, "readRegions") is better.

norshtein: Thanks, will change it.

jeremyrickard:
This LGTM with all the clarification. Will merge it pending lifecycle tests.

PR #552, opened 2018-08-23
Use user provided update parameters without judging empty

### LINKED ISSUE #550: Why judge empty in update?
Hi, it is noticed in [pkg/api/update.go#L367](https://github.com/Azure/open-service-broker-azure/blob/master/pkg/api/update.go#L367), the code judges whether the value is empty. The value is raw content provided by user, and the user might update the instance using empty value. For example, user might [update a cosmos instance](https://github.com/Azure/open-service-broker-azure/pull/544) using `{"readRegions": []}` to delete all existing read regions. Is there any special concern doing the judgement here?

norshtein:
Close #550 .


krancour:
@norshtein?

norshtein:
Sorry @krancour , seems I mis-closed this PR...

PR #553, opened 2018-08-23
Make all plans in cosmosdb module to StabilityPreview

### LINKED ISSUE #551: Is there any special concern keeping the cosmos module as `Experimental`?
Hi, we are using the cosmos module recently, but seems every plan under this module has  `StabilityExperimental`. From our point of view, this module works pretty well. Is there any special concern keeping the cosmos module as `Experimental`?

norshtein:
Close #551


jeremyrickard:
@norshtein should we merge in #544 before we do this. Any thoughts?

norshtein:
I still have some little optimization on that pr, please keep that pr open and it is expected to complete tomorrow.

krancour:
tbh #544 makes me hesitate to mark this module preview because such a major feature seems an indicator that we're still figuring out the features, experience, and implementation for this module.

jeremyrickard:
@krancour 

I *think* that the addition of #544 gets us to pretty much parity with cosmos in terms of capabilities according to https://docs.microsoft.com/en-us/azure/templates/microsoft.documentdb/databaseaccounts

Doesn't seem like there are really any other outstanding capabilities to get added, unless we're talking some sort of "DR" built into the module. 

Do you have any other gaps in mind?

jeremyrickard:
FWIW I do think that we should ptobably enable the capability for  #544 in before marking this stable. Whether that means we don't handle update because it's overly complex due to Cosmos DB or we do handle it, because the service *is* complex and the broker should hide that.

norshtein:
I think it's worth trying to handle update in the broker. Then we can listen to user's feedback and improve it.

krancour:
@jeremyrickard my concerns aren't over feature parity so much as wanting to be sure we've modeled the many services and plans appropriately for real world use. For instance, are all the parent/child relationships well-aligned with how developers and their applications tend to consume the various APIs that Cosmos supports?

jeremyrickard:
@krancour I think it's a valid consideration to think about what needs to happen, and I'dd suggest that we probably should open up some discrete issues to maybe track the things you're concerned about.

At the same time, if we refer back to our `stability` documentation: 

>preview - We have a better understanding how a service should look, but we don't yet guarantee backward compatibility. We do guarantee that this service won't go back to experimental, so we are committing to making it exist in some shape or form

>experimental - We have an idea for a new service in the catalog that we want to support, but don't have a good idea how it should look. We are essentially “throwing something against the wall” to see if it'll stick. Experimental services may be radically changed or removed at any time

I think that in the case of CosmosDB, we're past the "experimental" stage. If we introduce parent-child relationships, I think we can do that in non-destructive ways. We aren't providing a guarantee of backward compatibility, but we *are* saying that since it's in preview, we are committing to making sure "it exists in some shape or form". I think we can both agree CosmosDB will continue to exist in *some* shape or form. The addition of #544 adds the last significant piece of functionality that gives us broad coverage of the service. 

The fact that we don't guarantee that an experimental service will continue to exist is probably a turn off to users of the service and I think we would both agree we'd like to get more feedback on the current implementation, so I really don't see any downside to making CosmosDB a `preview` service.

lachie83:
I agree with @jeremyrickard using the definitions provided in the stability documented cited. I think that the risk we run by leaving this in experimental is in-deference in-terms of commitment level which may deter the community from using it. We still have flexibility for change in preview with a commitment level of a goal of eventually having a stable release. You create the opportunity for real feedback to shape the features/functionality from users.

campbelldgunn:
I think that based on the definition @jeremyrickard mentions on the `stability` documentation that it appears it meets `preview`. Thought I am also interested in parent/child relationships and happy to validate it, if that assists in moving this forward. Thanks.

krancour:
I agree that based on our definitions, we can move this into `preview`.

zhongyi-zhang:
Please also update the stability in the module doc. Adding an overview like MySQL doc would be better!

norshtein:
Hi @zhongyi-zhang ,I'd like to make changes on both code and doc after #544 is merged...

jeremyrickard:
@norshtein I've merged #544. If you want to update anything here, go for it. Otherwise I'll merge tomorrow.

norshtein:
Hi @jeremyrickard , I have updated the doc, please take a review on it, thanks! And I don't have any other thing to change.

Issue #545, opened 2018-08-09
Support multiple OSBA instances with single Redis


jeremyrickard:
Currently, OSBA requires Redis for Storage and Async task execution. Currently, it's not really possible to safely operate like this. 

We should provide a mechanism for allowing this mode of operation. It would dovetail nicely with namespaced service brokers in Service Catalog!

PR #558, opened 2018-08-24
Adding optional prefix to Redis (async and storage)

jeremyrickard:
* This adds a prefix to the Storage Redis code
* Prefix is optional and is pre-pended
* This can be used to enable multiple OSBA instances to share a Redis
  without having data collisions.
* The default behavior when installing with Helm will be to not provide an empty value, which should result in not preprending a unique identifier. 
* Since this will be defined in the environment variables for the deployment, restarts should continue working with that identifier and pick up anything queued. 

Closes: #545


norshtein:
Hi @jeremyrickard , seems config files for Cloud Foundry are not changed. This feature is also useful for Cloud Foundry. Could you please add `REDIS_PREFIX` in [this file](https://github.com/Azure/open-service-broker-azure/blob/master/contrib/cf/manifest.yml) and [this file](https://github.com/Azure/open-service-broker-azure/blob/master/contrib/cf/pcf-tile/tile.yml) ?

jeremyrickard:
Yep, I haven’t quite finished testing this out, so I haven’t fully updated all the configs and docs. When I remove the WIP tag I’ll give you a shout.

contrib/k8s/charts/open-service-broker-azure/README.md
```diff
@@ -121,6 +121,7 @@ Broker chart and their default values.
 | `redis.port` | _If and only if_ `redis.embedded` is `false`, this option specifies the port to connect to on the remote Redis host. | `6380` |
 | `redis.redisPassword` | Specifies the Redis password. If `redis.embedded` is `true`, this option sets the password on the Redis cache itself _and_ provides it to OSBA. If `redis.embedded` is `false`, this option only provides the password to OSBA. In that case, the value of this option must be the correct password for the remote Redis cache. | `"password"`; __Do not use this default value in production!__ |
 | `redis.enableTls` | _If and only if_ `redis.embedded` is `false`, this option specifies whether to use a secure connection to the remote Redis host. | `true` |
+| `redis.prefix` | Set this value to enable multiple OSBA instances to share one Redis. | `` |
```
krancour: Just to cover our bases, I would maybe note that this prefix must not be changed for an existing deployment of OSBA.

jeremyrickard: Yeah, that's a good call.

contrib/k8s/charts/open-service-broker-azure/values.yaml
```diff
@@ -87,3 +87,8 @@ redis:
   ## If not embedded, specifies whether to use a secure connection to the
   ## remote Redis host
   enableTls: true
+
+  ## Used to introduce a unique identifier to enable multiple OSBA instances 
+  ## to share one redis. If nothing is provided here, no additional identifier
+  ## is added.
```
krancour: Same comment as [here](https://github.com/Azure/open-service-broker-azure/pull/558/files#r215982116).

pkg/async/redis/cleaner.go
```diff
@@ -61,15 +61,15 @@ func (e *engine) defaultClean(
 				if err := e.cleanActiveTaskQueue(
 					ctx,
 					workerID,
-					getActiveTaskQueueName(workerID),
+					e.getActiveTaskQueueName(workerID),
```
krancour: All changes to the async package should probably be PR'ed to deis/async or else #560 will undo much of this PR.

jeremyrickard: Yeah, I think this was in flight before that move happened. I'll go ahead and send a PR there. Would it be better to revert the async here?

krancour: I'm thinking make the PR to deis/async first. We'll get that merged and do a minor release. I'll update #560 to use async 1.1, get that merged, and then you can rebase this PR and all the async changes will just go away when the merge conflict is resolved.

contrib/cf/manifest.yml
```diff
@@ -15,12 +15,14 @@ applications:
...
     STORAGE_REDIS_DB: 0
     STORAGE_REDIS_ENABLE_TLS: true
+    STORAGE_REDIS_PREFIX:
     CRYPTO_ENCRYPTION_SCHEME: AES256
     CRYPTO_AES256_KEY: AES256Key-32Characters1234567890
     ASYNC_REDIS_HOST: <HOSTNAME FROM AZURE REDIS CACHE>
     ASYNC_REDIS_PASSWORD: <PRIMARYKEY FROM AZURE REDIS CACHE>
     ASYNC_REDIS_PORT: 6380
     ASYNC_REDIS_DB: 1
+    ASYNC_REDIS_PREFIX:
```
krancour: Looks like in our Helm chart, we only make `REDIS_PREFIX` settable. (In the code, both storage and async will fall back on `REDIS_PREFIX` if they can't find `STORAGE_REDIS_PREFIX` and `ASYNC_REDIS_PREFIX`, respectively.)

I think we should _also_ only expose `REDIS_PREFIX` here. The reason is that this prevents those two settings from being tuned independently and forces both to have the same value-- which imo is enormously useful for helping to maintain a correlation / traceability by grouping all keys for a given OSBA deployment together.

krancour:
Is there a missing change for the OpenShift template?

contrib/cf/manifest.yml
jeremyrickard: Yeah, great point. I'll make that change,.

jeremyrickard:
@krancour yeah, forgot about the OpenShift template, thanks!

PR #562, opened 2018-08-30
[WIP] DO NOT MERGE: Use go 1.11.0 w/ modules (instead of dep)

krancour:
This builds on other, unmerged work, so let's not even think about merging this yet.


krancour:
Besides the fact that this needs a significant rebase, I want a total do-over on this based on some things I learned from using Go modules on another project. Closing this. Will PR something similar at a later date.

Issue #561, opened 2018-08-29
Changes on Redis module


norshtein:
Hi @jeremyrickard and @krancour , we have noticed that our customer has potential demand to use OSBA to manage their Redis cache. But firstly, we should make Redis module to `StabilityPreview` to convince the customer. Based on the [template](https://docs.microsoft.com/en-us/azure/templates/microsoft.cache/redis), I have made a draft on what will be changed to promote it to `StabilityPreview`. Please bring up your suggestions so that we can pay attention to them while working.

- Provision:
  - Make `skuCapacity` as an optional parameter. For now, there is only one default capacity for each plan, we can provide users more options after parameterizing capacity. It has several valid values for different plan.
  - For premium plan, following provisioning parameters will be added:
    - `shardCount`. Optional, integer, default 1.
    - `subnetId`. Optional, string, default null.
    - `staticIP`. Optional, string, default null, is valid only when `subnetId` is provided.
    - `redisConfiguration`. Optional, object, default null.

- Update:
Provide the ability to update an existing Redis instance. Allowed update parameters:
  - `skuCapacity`. The capacity can only be updated from a smaller size to a larger size.
  - `enableNonSslPort`.
  - For premium plan, following updating parameters is available:
    - `shardCount`. Optional.
    -  `redisConfiguration`. Optional.

One concern is the parameter `redisConfiguration` is too complicated and there are so many possible keys in it. It's hard for service broker to validate user provided keys, but one feature for premium cache is configuring data persistence, which needs `redisConfiguration` parameter. I'm really in a dilemma.

You may have noticed that `zones` and `tenantSettings` is in the template but not in the provisioning parameters. Reasons:
- `zones`: This feature is still in preview state in Redis, so I haven't added it to OSBA.
- `tenantSettings`: Actually I haven't found any doc describing what the object should be... So I haven't made it as a parameter. Please tell me if you have any information about it...

CC: @bingosummer @zhongyi-zhang

krancour:
Can this be broken up into smaller issues so that individual improvements can be considered independently. I don't want all these new features to be an all or nothing.

zhongyi-zhang:
#577 LGTM.
Both #575 and #576 almost LGTM. They are good features in Redis service. The only thing I am not sure is that, the subnet is an extra ARM resource -- it is too complicated for a broker service. Similar problem on the parameter for storage connection string in #576.
But anyway they are optional, we don't need to stop someone from using them. What do you think?

[...]

PR #564, opened 2018-09-11
Redis dev

norshtein:
Make changes based on the proposal in  #561 .


pkg/services/rediscache/bind.go
```diff
@@ -20,13 +20,23 @@ func (s *serviceManager) GetCredentials(
...
 
-	redisPort := 6379
+	var redisPort int
+	var scheme string
+	if dt.NonSSLEnabled {
+		redisPort = 6379
+		scheme = "redis"
+	} else {
+		redisPort = 6380
+		scheme = "rediss"
```
zhongyi-zhang: Cool, as Spring Cloud connector fixed https://github.com/spring-cloud/spring-cloud-connectors/issues/235.

tests/lifecycle/rediscache_cases_test.go
```diff
@@ -2,14 +2,197 @@
...
+	virtualNetworksClient := networkSDK.NewVirtualNetworksClientWithBaseURI(
+		azureConfig.Environment.ResourceManagerEndpoint,
+		azureConfig.SubscriptionID,
+	)
+	virtualNetworksClient.Authorizer = authorizer
+
+	virtualNetworkName := uuid.NewV4().String()
+	eastus := "eastus"
+	subnetName := "default"
+	subnetCIDR := "172.19.0.0/24"
```
zhongyi-zhang: Try this to get a string pointer: https://github.com/Azure/go-autorest/blob/master/autorest/to/convert.go#L30.
It is the usage in sdk samples: https://github.com/Azure-Samples/azure-sdk-for-go-samples/blob/master/network/network.go#L38

norshtein: Thanks, will fix it.

norshtein:
Besides new features in Redis cache, this pr also makes [changes](https://github.com/Azure/open-service-broker-azure/pull/564/files#diff-719732f43be191becfae60b2137b2fcf) on the lifecycle test framework based on @zhongyi-zhang 's code. That's, add `preProvisionFns` in `serviceLifecycleTestCase` struct to provision necessary resources before executing the provisioning steps.

norshtein:
Hey @krancour @jeremyrickard  , could you please have a review on it?

docs/modules/rediscache.md
```diff
@@ -25,10 +24,24 @@ Provisions a new Redis cache.
 |----------------|------|-------------|----------|---------------|
 | `location` | `string` | The Azure region in which to provision applicable resources. | Required _unless_ an administrator has configured the broker itself with a default location. | The broker's default location, if configured. |
 | `resourceGroup` | `string` | The (new or existing) resource group with which to associate new resources. | N | If an administrator has configured the broker itself with a default resource group and nonde is specified, that default will be applied, otherwise, a new resource group will be created with a UUID as its name. |
+| `skuCapacity` | `integer` | The size of the Redis cache to deploy.  Valid values: for C (Basic/Standard) family (0, 1, 2, 3, 4, 5, 6), for P (Premium) family (1, 2, 3, 4). | N | If not provided, `0` is used for C (Basic/Standard) family; `1` is used for P (Premium) family. |
```
carolynvs: Is there a unit of measurement that we can mention in this description? GB?

docs/modules/rediscache.md
```diff
@@ -25,10 +24,24 @@ Provisions a new Redis cache.
...
+| `enableNonSslPort ` | `string` | Specifies whether the non-ssl Redis server port (6379) is enabled. Valid values: (`enabled`, `disabled`) | N | If not provided, `enabled` is used. **Note**:  this behavior is different from Azure portal. In OSBA, non-SSL port is enabled by default. That's because we want to make sure your application can work normally even your application doesn't support Redis with SSL. |
```
carolynvs: I lean towards not enabling the unsecured port by default. This kind of goes against the Microsoft value of "secure by default". If someone wants to use the unsecured port, then they have this flag here to do so, no?

docs/modules/rediscache.md
```diff
@@ -25,10 +24,24 @@ Provisions a new Redis cache.
...
+| -------------------------------------------------------- | --------- | ------------------------------------------------------------ | ------------------------------------------------------------ | ------------------------------------------------------------ |
+| `shardCount`                                             | `integer` | The number of shards to be created on a Premium Cluster Cache. This action is irreversible. The number of shards can be changed later. | N                                                            | If not specified, no additional shard will be created.       |
+| ` subnetId `                                             | `string`  | The full resource ID of a subnet in a virtual network to deploy the Redis cache in. The subnet should be in the same region with Redis cache. Example format: /subscriptions/{sub}/resourceGroups/{rg}/providers/Microsoft.Network/virtualNetworks/{vn}/subnets/{sn} | N                                                            | If not specified, the Redis cache will be deployed as common. |
```
carolynvs: > If not specified, the Redis cache will be deployed as common

This isn't clear to me what "as common" means?

docs/modules/rediscache.md
```diff
@@ -1,22 +1,21 @@
 # [Azure Redis Cache](https://azure.microsoft.com/en-us/services/cache/)
 
-_Note: This module is EXPERIMENTAL. To enable this module, you must run Open Service Broker for Azure with modules.minStability set to `experimental`_
+_Note: This module is EXPERIMENTAL and is not included in the General Availability release of Open Service Broker for Azure. It will be added in a future OSBA release._
```
carolynvs: Why does this remove the instructions on how to enable the module?

docs/modules/rediscache.md
```diff
@@ -44,11 +57,35 @@ Binding returns the following connection details and shared credentials:
...
+
+##### Update
+
+Updates existing Redis cache.
+
+###### Updating parameters
+
+| Parameter Name     | Type      | Description                                                  | Required |
+| ------------------ | --------- | ------------------------------------------------------------ | -------- |
+| `skuCapacity`      | `integer` | The size of the Redis cache to deploy.  Valid values: for C (Basic/Standard) family (0, 1, 2, 3, 4, 5, 6), for P (Premium) family (1, 2, 3, 4).  **Note**: you can only update from a smaller capacity to a larger capacity, the reverse is not allowed. | N        |
```
carolynvs: Again, it would really help to specify the unit here.

pkg/services/rediscache/plan_schemas.go
```diff
@@ -20,6 +24,119 @@ func generateProvisioningParamsSchema() service.InputParametersSchema {
...
 					" to associate new resources.",
 			},
+			"enableNonSslPort": &service.StringPropertySchema{
+				Title:         "Enable non-SSL port",
+				Description:   "Specifies whether the non-ssl Redis server port (6379) is enabled.",
+				AllowedValues: []string{"enabled", "disabled"},
+				DefaultValue:  "enabled",
+			},
+			"skuCapacity": &service.IntPropertySchema{
+				Title:         "Sku capacity",
```
carolynvs: nit: I believe that it should be SKU (throughout this PR) because it is an acronym.

pkg/services/rediscache/provision.go
```diff
@@ -77,5 +73,58 @@ func (s *serviceManager) deployARMTemplate(
 	}
 	dt.PrimaryKey = service.SecureString(primaryKey)
 
+	if instance.ProvisioningParameters.GetString("enableNonSslPort") == enabled {
```
carolynvs: This can be simplified to a single assignment

```go
dt.NonSSLEnabled = (instance.ProvisioningParameters.GetString("enableNonSslPort") == enabled)
```

pkg/services/rediscache/provision.go
```diff
@@ -77,5 +73,58 @@ func (s *serviceManager) deployARMTemplate(
...
+	var enableNonSslPort string
+	if pp.GetString("enableNonSslPort") == enabled {
+		enableNonSslPort = "true"
+	} else {
+		enableNonSslPort = "false"
+	}
+
+	redisConfiguration := pp.GetObject("redisConfiguration").Data
+	if value, ok := redisConfiguration["rdb-backup-enabled"]; ok {
+		if value == enabled {
```
carolynvs: same as above, this can be simplified into a single assignment

```go
redisConfiguration["rdb-backup-enabled"] = (value == enabled)
```

pkg/services/rediscache/update.go
```diff
@@ -1,13 +1,78 @@
...
+		map[string]interface{}{},
+		tags,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("error updating redis instance %s", err)
+	}
+
+	nonSSLPortEnabled := up.GetString("enableNonSslPort")
+	if nonSSLPortEnabled != "" {
+		if nonSSLPortEnabled == enabled {
```
carolynvs: Same as before this can be simplified:

```go
dt.NonSSLEnabled = (nonSSLPortEnabled == enabled)
```

pkg/services/rediscache/catalog.go
```diff
@@ -3,6 +3,10 @@ package rediscache
 import "github.com/Azure/open-service-broker-azure/pkg/service"
 
 func (m *module) GetCatalog() (service.Catalog, error) {
+	bpd := newBasicPlanDetail()
```
jeremyrickard: Can we just declare the planDetail structs inline here? This function isn't used anywhere else and I think it would be clearer if we just detailed it here so it was clear what was happening, without needing to jump to another file.

krancour:
#561 proposes a lot of different changes and per [this comment](https://github.com/Azure/open-service-broker-azure/issues/561#issuecomment-421116545), it would be nice if we could break this PR up into smaller, more discrete PRs. I'd like to consider each new feature individually instead of having one all-or-nothing PR with many features.

docs/modules/rediscache.md
norshtein: That's configured to be enabled to keep consistent with [current implementation](https://github.com/Azure/open-service-broker-azure/blob/master/pkg/services/rediscache/arm_template.go#L20). I agree it's reasonable to make insecure connection disabled by default. I will change the code later.

docs/modules/rediscache.md
norshtein: Actually I haven't changed this section... I think something went wrong with my version control tool or text editor...

pkg/services/rediscache/plan_schemas.go
norshtein: Thanks, will fix it.

pkg/services/rediscache/provision.go
norshtein: That's neat. I will change it later.

pkg/services/rediscache/catalog.go
norshtein: The `planDetail` struct is also used when [generating plan schemas](https://github.com/Azure/open-service-broker-azure/pull/564/files#diff-b9a96ba77a9f65df70ee14de705dae70). Necessary information is extracted from the struct to generate plan schemas. If we make the struct inline, the function for generating plan schemas should also be inlined, which will makes this file very big.

norshtein:
Hi @krancour , thanks for the advise. I'm going to break this pr into several small PRs.

pkg/services/rediscache/catalog.go
jeremyrickard: Is the `newBasicPlanDetail` function (or `newStandardPlanDetail()` or `newPremiumPlanDetail()` invoked anywhere but here? That was my point. Currently it looks like no, so you could just do this no?

```
bpd := planDetail{
   planName:        basic,
    allowedCapacity: []int64{0, 1, 2, 3, 4, 5, 6},
}
```

My comment was that the method is invoked only here, not in other places so it seemed unneeded.

docs/modules/rediscache.md
norshtein: Detailed cache size can be found [here](https://azure.microsoft.com/en-us/pricing/details/cache/). I'll add it later.

pkg/services/rediscache/plan_detail.go
```diff
@@ -0,0 +1,33 @@
+package rediscache
+
+const basic = "basic"
+const standard = "standard"
+const premium = "premium"
+
+type planDetail struct {
```
jeremyrickard: i wasn't questioning the definition of the struct, just those methods below that a.) aren't exported and b.) are only called from catalog.go..unless I'm missing a usage.

pkg/services/rediscache/catalog.go
norshtein: Aha, I get your point. Yes, this function is only used here. I'll change it later, thanks!

norshtein:
Going to close this PR, smaller PRs will be opened later.

Issue #563, opened 2018-09-10
Error provisioning azure-mysql-5-7 basic plan


arapulido:
Hello,

I am running Service Catalog 0.1.31 and OSBA 1.1.0 on minikube. I had never had issues with provisioning a mysql instance everytime I tried. Today I created a new minikube cluster and tried those versions and the provisioning failed.

Logs don't show anything useful. Automatically, because of the failure of the provisioning, the auto-deprovision kicks off and the osba pods crashes:

```
panic: interface conversion: service.InstanceDetails is nil, not *mysql.allInOneInstanceDetails

goroutine 137 [running]:
github.com/Azure/open-service-broker-azure/pkg/services/mysql.(*allInOneManager).deleteARMDeployment(0xc4200800b0, 0xc348c0, 0xc420063dc0, 0xc420028900, 0x24, 0x0, 0x0, 0xc420028930, 0x24, 0xc3a380, ...)
	/go/src/github.com/Azure/open-service-broker-azure/pkg/services/mysql/all_in_one_deprovision.go:23 +0x178
github.com/Azure/open-service-broker-azure/pkg/services/mysql.(*allInOneManager).(github.com/Azure/open-service-broker-azure/pkg/services/mysql.deleteARMDeployment)-fm(0xc348c0, 0xc420063dc0, 0xc420028900, 0x24, 0x0, 0x0, 0xc420028930, 0x24, 0xc3a380, 0xc420362480, ...)
	/go/src/github.com/Azure/open-service-broker-azure/pkg/services/mysql/all_in_one_deprovision.go:14 +0x79
github.com/Azure/open-service-broker-azure/pkg/service.(*deprovisioningStep).Execute(0xc4201018e0, 0xc348c0, 0xc420063dc0, 0xc420028900, 0x24, 0x0, 0x0, 0xc420028930, 0x24, 0xc3a380, ...)
	/go/src/github.com/Azure/open-service-broker-azure/pkg/service/deprovisioner.go:67 +0xe5
github.com/Azure/open-service-broker-azure/pkg/broker.(*broker).executeDeprovisioningStep(0xc42008e100, 0xc348c0, 0xc420063d00, 0xc37ee0, 0xc420063cc0, 0x0, 0x0, 0x0, 0x0, 0x0)
	/go/src/github.com/Azure/open-service-broker-azure/pkg/broker/deprovision.go:91 +0x619
[...]

PR #565, opened 2018-09-11
Handle `nil` instance detail in deprovisioning

norshtein:
According to #563, there exists the possibility that the instance detail is `nil` when error happens in early provisioning, so we do a special judgement when creating deprovisioning step. If the instance detail is `nil`, we simply return without executing real deprovisioning function.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -4,6 +4,7 @@
 [[projects]]
   name = "github.com/Azure/azure-sdk-for-go"
   packages = [
+    "services/cognitiveservices/mgmt/2017-04-18/cognitiveservices",
     "services/cosmos-db/mgmt/2015-04-08/documentdb",
     "services/eventhub/mgmt/2017-04-01/eventhub",
     "services/keyvault/mgmt/2016-10-01/keyvault",
@@ -212,6 +213,6 @@
 [solve-meta]
   analyzer-name = "dep"
   analyzer-version = 1
-  inputs-digest = "add149e2845794e0833cf506bf80d3255d7c7ed34e39ad07fa6062cfa35ceed0"
+  inputs-digest = "57d57986dbe9b1ac6030ddabd52602b141327ff9560373efbff73b3dfe213244"
   solver-name = "gps-cdcl"
   solver-version = 1
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -8,6 +8,7 @@
     "services/eventhub/mgmt/2017-04-01/eventhub",
     "services/keyvault/mgmt/2016-10-01/keyvault",
     "services/mysql/mgmt/2017-04-30-preview/mysql",
+    "services/network/mgmt/2018-01-01/network",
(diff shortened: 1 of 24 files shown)


norshtein:
So we have two possible implementations:
1. Do the judgement in http deprovision handler, an Error code should be returned when the service instance detail is `nil`, which causes the deletion of the service instance fails. 
2. Wrapping special judgement when creating deprovisioning step. In this case, the service instance can be successfully deleted(since we return `nil` as error, the broker will think the deprovisioning step succeed). 

I prefer using the second implementation. When the service instance detail is `nil`, we can guarantee no service instance is created and no resource is allocated, and it's reasonable and user-friendly to just delete this record in the PaaS platform.

zhongyi-zhang:
@norshtein for 1# we can also deprovision successfully by just deleting the invalid instance from store and returning OK.

norshtein:
Yes, you're right. The advantage of 1# is it's more "standard" and it might save function invoking time, and the disadvantage of 1# is that more code need to be injected to existing code. I'm OK with both implementations and I can change current implementation to 1#. Which one do you prefer @zhongyi-zhang  @jeremyrickard @krancour ?

krancour:
I prefer option 1.

I should also say that we hadn't put a lot of thought into what it looks like to deprovision something that previously failed provisioning. (Is that even valid?) I don't have time to dig in right now, but I would strongly urge looking to the OSB spec for some clarification in how this is expected to work.

zhongyi-zhang:
See lines above this section: https://github.com/openservicebrokerapi/servicebroker/blob/v2.14/spec.md#polling-last-operation-for-service-bindings
It specifies a "MUST" to the platform and only a "MAY" to the broker. That delegates how to handle it to the broker. With `nil` instance detail -- it must be an orphan, we can eliminate the instance in deprovision API handler level. With valid instance detail, the deprovision request is delegated to the service module. And the service module must ensure deleting all the Azure resources it created, whatever the instance state is, successful or failed.

norshtein:
Here comes another question: what if the service instance is not `nil` but the service instance is not successfully created? Say it fails in ARM deployment, in this case, the deprovision step of this service will be executed, which will return an error for most services (since the SDK can't delete an non-existing instance). Then the service instance will go to state `delete failed`, and can never be deleted by normal operation.

zhongyi-zhang:
@norshtein that's what the service modules should handle correctly by giving consideration to both success and failure state. Generally they can just ignore no-content deletions. Think about that: even the deprovisioning of a successful provisioning might fail, and only part of Azure resources are deleted. Then the platform retry deprovisioning, the situation would be the same as a failed provisioning.

norshtein:
But the deprovision will always fail in my example. Currently almost no service's deprovisioning step judges whether the service instance is existing.

zhongyi-zhang:
So they are separate issues for the service modules. Fortunately they seldom occur as Azure REST APIs are not so unstable. We can fix them later.

PR #567, opened 2018-09-14
[Redis] Make capacity and ssl configurable and make instance updatable

norshtein:
As @krancour said in #564 , this PR is the first one of separated PRs. Later PRs will be based on this PR. This PR enables several basic and fundamental features mentioned in #561 :
- Provision
  - Make `skuCapacity` as an optional parameter. For now, there is only one default capacity for each plan, we can provide users more options after parameterizing capacity. It has several valid values for different plan.
  - Make `enableNonSslPort` as an optional parameter.

Provide the ability to update an existing Redis instance:
- Update 
  - `skuCapacity`. The capacity can only be updated from a smaller size to a larger size.
  - `enableNonSslPort`

From my point of view, these features is necessary and should be added anyway. So I put them together so that we can avoid annoying rebase.

### DIFF
--- a/docs/modules/rediscache.md
+++ b/docs/modules/rediscache.md
@@ -6,29 +6,30 @@ _Note: This module is EXPERIMENTAL. To enable this module, you must run Open Ser
 
 ### Service: azure-rediscache
 
-| Plan Name | Description |
-|-----------|-------------|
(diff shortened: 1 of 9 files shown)


norshtein:
I prefer to send other PRs after this PR merged, in case code in this PR need to be changed.

PR #569, opened 2018-09-14
Adding Tongyao Si and Zhongyi Zhang to the CODEOWNERS file

jeremyrickard:
@zhongyi-zhang this was my initial guess at who from your team would make sense as a first round of people to onboard as OSBA maintainers. Comment here and let me know if you disagree!


jeremyrickard:
Docs only, no CI.

🎊

Issue #566, opened 2018-09-14
Additional SQL functionalities


zhongyi-zhang:
It is some effort in https://github.com/CloudFoundryOnAzure/open-service-broker-azure. The functionalities are to cover some more SQL scenarios. The design and implementation would be in OSBA style. And most things are enough generic.
The codes will be refined and separated into small PRs for easier review. Let me schedule the road map first. 

| Short description | Long description | PR link | Review done and merged |
|-----------|-----------|-----------|-----------|
| Add a service to use existing Azure SQL server | The same idea with #559. It aligns with original dbms-only service as a parent service in MSSQL module. | | |
| Add a service to use existing Azure SQL server | It aligns with original database-only service as a child service in MSSQL module. Except provisioning behaviors, be par with original database-only service. | | |
| Add additional tags and a tag filter | To avoid confusing users who don't want these additional functionalities. From user view, it would be `ENABLE_MIGRATION_SERVICES` and `ENABLE_DISASTER_RECOVERY_SERVICES` in environment variables. All default to `false`. | | |
| Add a map of V2 GUIDs for services and plans | To achieve co-existence with MASB. Replace the IDs in catalog with the map when booting. Disabled by default. | | |
| Add a separate module for MSSQL Failover Group | This is complicated. Let's follow up later after solving above items. Just a reminder here. | | |

Each of them is an independent change. The PR order can also be arbitrary. Just list them in a natural order from my understanding. I'll create one more PR after a previous one merged.

Here is the blueprint of MSSQL module. The items in yellow are proposed above.
![image](https://user-images.githubusercontent.com/13326941/45525414-66922080-b805-11e8-9fe0-7f907cd22fe6.png)

Issue #559, opened 2018-08-27
Implement a "Bring Your Own" capability to register an existing SQL Server instance


jeremyrickard:
We should implement a new "byob" style service for SQL Server that will enable customers to bring an existing SQL Server instance into the OSBA fold.

OSBA should be able to:
* Create bindings on this instance
* Remove bindings from this instance

Out of scope:
* Deprovision
* creating new databases on the SQL Server

norshtein:
Hey, I think @zhongyi-zhang has implemented this feature.

jeremyrickard:
@zhongyi-zhang can you open a PR for this if you've already started? would be cool to start looking at it and figure out anything that needs to change in the broker.

zhongyi-zhang:
@jeremyrickard it is done but hard to create a pure PR. Let me explain. Several months ago, in the discussion with Aaron, he thought some CF scenarios are out of OSBA scope. So, we temporarily decided to develop an "advanced OSBA" for CF users only. We planed to regularly merge upstream OSBA codes, and develop no-conflict features carefully. As a result, we have this repo: https://github.com/CloudFoundryOnAzure/open-service-broker-azure. It is still under development and doesn't have any release. Registering an existing SQL server as a service instance is a part of this PR: https://github.com/CloudFoundryOnAzure/open-service-broker-azure/pull/3. It bases on some other changes.
I saw the mail thread about the meeting with you and Ning. Glad to hear that we will keep one branch. I think we can discuss how to deal with it later after you understand the overview of CF advanced requirements and what we tried to implement. We are sorting the related docs and will share them soon. Any of your thoughts are super appreciated.  I really love your thoughts and works on OSBA.

krancour:
I used to like this idea, however, my thoughts are changing a bit.

If the driver for this is to facilitate migrations from MASB to OSBA, there's a simpler plan that's now been proposed. I think that plan needs further consideration and discussion before committing to the BYO feature.

norshtein:
Hi @krancour , if you are worried about "byob" service may affect normal users usage, @zhongyi-zhang has set up a mechanism that "byob" service is default off and invisible to users, it becomes visible after the user turning it on in the manifest. 

I think "byob" service is necessary since many customers are managing Azure SQL server by themselves and they want to take advantage of OSBA to make their access safer and more convenient.

zhongyi-zhang:
Closed via #570.

PR #570, opened 2018-09-17
Add registered mssql dbms service

zhongyi-zhang:
Solve #559 and the first step in #566. Service doc will be updated after other related changes in #566 is done.

The changes in lifecycle test conflict PR https://github.com/Azure/open-service-broker-azure/pull/564 a bit. Will solve it if that one merged first.

### DIFF
--- a/pkg/services/mssql/dbms_registered_provision.go
+++ b/pkg/services/mssql/dbms_registered_provision.go
@@ -1,0 +1,123 @@
+package mssql
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/Azure/open-service-broker-azure/pkg/service"
+)
+
+func (d *dbmsRegisteredManager) GetProvisioner(
+	service.Plan,
+) (service.Provisioner, error) {
+	return service.NewProvisioner(
+		service.NewProvisioningStep("preProvision", d.preProvision),
+		service.NewProvisioningStep("getServer", d.getServer),
+		service.NewProvisioningStep("testConnection", d.testConnection),
+	)
+}
+
+func (d *dbmsRegisteredManager) preProvision(
+	_ context.Context,
+	instance service.Instance,
+) (service.InstanceDetails, error) {
+	pp := instance.ProvisioningParameters
+	return &dbmsInstanceDetails{
+		ARMDeploymentName:          "",
+		ServerName:                 pp.GetString("server"),
+		AdministratorLogin:         pp.GetString("administratorLogin"),
+		AdministratorLoginPassword: service.SecureString(pp.GetString("administratorLoginPassword")), // nolint: lll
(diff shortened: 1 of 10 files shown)


pkg/services/mssql/dbms_registered_provision.go
```diff
@@ -0,0 +1,123 @@
...
+	if err != nil {
+		return nil, err
+	}
+	defer masterDb.Close() // nolint: errcheck
+
+	// Is there a better approach to verify if it is a sys admin?
+	rows, err := masterDb.Query("SELECT 1 FROM fn_my_permissions (NULL, 'DATABASE') WHERE permission_name='ALTER ANY USER'") // nolint: lll
+	if err != nil {
+		return nil, fmt.Errorf(
+			`error querying SELECT from table fn_my_permissions: %s`,
```
norshtein: Why the string is wrapped with `` ` `` instead of `"`, the same below

pkg/services/mssql/dbms_registered_provision.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+	defer rows.Close() // nolint: errcheck
+	if !rows.Next() {
+		return nil, fmt.Errorf(
+			`error user doesn't have permission 'ALTER ANY USER'`,
+		)
+	}
+	if err := rows.Err(); err != nil {
+		return nil, fmt.Errorf(
+			`error iterating rows`,
```
norshtein: Here `err` is not `nil`, we should add it to the returned error message.

pkg/services/mssql/dbms_registered_provision.go
zhongyi-zhang: Oh, there is no point to use `\`` here. Will change them to `"`. Thanks.

pkg/services/mssql/dbms_registered_update.go
```diff
@@ -0,0 +1,35 @@
...
+) error {
+	return nil
+}
+
+func (d *dbmsRegisteredManager) GetUpdater(
+	service.Plan,
+) (service.Updater, error) {
+	return service.NewUpdater(
+		service.NewUpdatingStep("updateAdministrator", d.updateAdministrator),
+		service.NewUpdatingStep("testConnection", d.testConnection),
```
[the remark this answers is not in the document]
zhongyi-zhang: Good point and thanks for the suggestion. Will follow it. More further, I think `testConnection` should also be a function call in provisioning step so that it wouldn't look weird.

pkg/services/mssql/dbms_registered_provision.go
zhongyi-zhang: Will fix.

norshtein:
LGTM

PR #572, opened 2018-09-19
Add preProvisionFns in serviceLifecycleTestCase

norshtein:
This PR add `preProvisionFns` in `serviceLifecycleTestCase`, this slice can contain any number of `preProvisionFn`. `preProvisionFn` is used to create any needed resource and add information to the provisioning parameters before the provisioning step is executed.


norshtein:
Going to close it since #570  has already contained this PR.

PR #574, opened 2018-09-19
Add method to get token and config in test framework

norshtein:
This PR adds two methods in test framework. They are used to get `azure.Config` and `autorest.BearerAuthorizer`, which may be required to create resources in [preProvisionFns](https://github.com/Azure/open-service-broker-azure/pull/572)


norshtein:
Going to close this PR since #570 has already contained this PR.

PR #577, opened 2018-09-19
[Redis] Make Redis premium able to set shard count

norshtein:
This PR cherry-picks #572 and #574 , and make Redis premium able to set shard count. Following features are enabled:

- Provision:
For premium plan, following provisioning parameters are added:
  - `shardCount`. Optional, integer, default not set, that is, no additional shard will be created.

- Update:
For premium plan, following updating parameters are allowed:
  - `shardCount`. Optional, integer, and can't be updated with `skuCapacity` at the same time.


pkg/services/rediscache/update.go
```diff
@@ -20,6 +21,17 @@ func (s *serviceManager) ValidateUpdatingParameters(
...
+
+	// Can't update `shardCount` and `skuCapacity` at the same time
+	if strings.ToLower(instance.Plan.GetName()) == premium {
+		ppSku := pp.GetInt64("skuCapacity")
+		upSku := up.GetInt64("skuCapacity")
+		ppShardCount := pp.GetInt64("shardCount")
+		upShardCount := up.GetInt64("shardCount")
+		if ppSku != upSku && ppShardCount != upShardCount {
+			return fmt.Errorf("can not update `shardCount` and `skuCapacity` at the same time") // nolint: lll
+		}
```
zhongyi-zhang: Checking this in OSBA vs leaving it to Azure, which is better?

norshtein: Leaving it to Azure leads to `update failed`, I prefer to validate it in the broker. Additionally, if we choose to leave this to Azure, one question we need to answer is why don't we leave all validations to Azure? For example, the constraint "Can't update the instance from a larger capacity to a smaller capacity" can also be left to Azure but for now we validate it in the broker.

zhongyi-zhang: I think the one in example can be also left to Azure. The validation in a broker should ensure that the parameters input reflect what users want to do (with the right keys). As for whether it is valid or not, that's what Azure should determine.

norshtein: I mentioned this example because [code](https://github.com/Azure/open-service-broker-azure/blob/master/pkg/services/mysql/all_in_one_update.go#L13) in mysql module does the same validation. I use the same logic to keep the same with current implementation, but I think it's reasonable to leave non-trivial validation to Azure, because it's hard to find all possible validations and deal with all of them in the broker.

zhongyi-zhang: OK, they might still have a difference. Decrease storage size would never be supported. What about updating `shardCount` and `skuCapacity` at the same time? Seems it is not something that can't be supported technically, but something not supported yet (the resource provider can just do one after another). Anyway, I agree to keep the validation here. We can remove it if someday this is doable.

PR #581, opened 2018-09-22
Buildpack is deprecated in favour of buildpacks

Samze:


norshtein:
Thanks for the contribution.

PR #582, opened 2018-09-24
Add nodeSelector beta.kubernetes.io/os=linux

PatrickLang:
OSBA won't pull or start on a Windows node. This constrains it to a Linux node.

Test steps:

```
helm dep up .\open-service-broker-azure\
cd .\open-service-broker-azure\charts
tar xvzf redis-0.1.0.tgz 
# Add nodeselector to redis/templates/deployment.yaml
helm install .\open-service-broker-azure\ --name osba --namespace osba --set azure.subscriptionId=$ENV:AZURE_SUBSCRIPTION_ID --set azure.tenantId=$ENV:AZURE_TENANT_ID --set azure.clientId=$ENV:AZURE_CLIENT_ID --set azure.clientSecret=$ENV:AZURE_CLIENT_SECRET
```

### DIFF
--- a/contrib/k8s/charts/open-service-broker-azure/values.yaml
+++ b/contrib/k8s/charts/open-service-broker-azure/values.yaml
@@ -5,7 +5,7 @@ image:
   ## Image location, NOT including the tag
   repository: microsoft/azure-service-broker
   ## Image tag
-  tag: v0.0.1 # This gets automatically overridden by our release process
+  tag: v1.1.0 # This gets automatically overridden by our release process
   ## "IfNotPresent", "Always", or "Never"
   pullPolicy: IfNotPresent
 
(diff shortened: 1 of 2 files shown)


contrib/k8s/charts/open-service-broker-azure/values.yaml
```diff
@@ -5,7 +5,7 @@ image:
   ## Image location, NOT including the tag
   repository: microsoft/azure-service-broker
   ## Image tag
-  tag: v0.0.1 # This gets automatically overridden by our release process
+  tag: v1.1.0 # This gets automatically overridden by our release process
```
jeremyrickard: @PatrickLang can you revert this? You don't need to update this because the file gets overridden when the chart is released.

PR #591, opened 2018-09-28
Validate parent-child relationship before provisioning

### LINKED ISSUE #589: Validate the parent-child relationship before provisioning child service instance
Currently, there is no method validating the parent-child relationship, so users may misuse it. For example, the user creates a `azure-mysql-5-7-dbms` instance and gives it an alias `dbms`, and then tries to creates a `
azure-postgresql-9-6-database` instance with `parentAlias` set to `dbms`, which is definitely a incorrect behavior, but the broker doesn't prevent this. Should we validate the parent-child relationship before provisioning child service instance?

norshtein:
Close #589


jeremyrickard:
This change will likely break how things work with Kubernetes. Please do not merge it until we can discuss more.

pkg/api/provision.go
```diff
@@ -324,9 +324,19 @@ func (s *server) isParentProvisioning(instance service.Instance) (bool, error) {
 		return false, err
 	}
 
-	//Parent has was not found, so wait for that that to occur
+	// Parent is not found, we treat this as a bad provision request
 	if !parentFound {
-		return true, nil
+		return false, fmt.Errorf(
```
jeremyrickard: This was done to support the way things can be done with Kubernetes. Take a Helm chart that may declare both a parent and child instance. When these are created, Kubernetes via Service Catalog will make the request for both resources, but there isn't a way to constrain the order. Therefore, you could get the child before the parent. In this scenario, the change you've introduced would cause things to end up in a failure state.

norshtein:
Got that. Thanks for the explanation!

zhongyi-zhang:
Extending _ParentServiceID_ to a set of IDs should be done before this change. Or, it would break the sql-dbms-registered service which was recently checked in. See catalog:
https://github.com/Azure/open-service-broker-azure/blob/master/pkg/services/mssql/catalog.go#L374
https://github.com/Azure/open-service-broker-azure/blob/master/pkg/services/mssql/catalog.go#L323
BTW, I think _ChildServiceID_ can be minimized to a simple bool to identity whether it is a parent service.

zhongyi-zhang:
@norshtein here: https://github.com/Azure/open-service-broker-azure/blob/master/pkg/service/schema.go#L544. The `alias` is appended to provisioning parameters if it is a parent service.

norshtein:
Thanks for pointing it out!

PR #598, opened 2018-10-08
Add a service to use existing Azure SQL database

zhongyi-zhang:
The third step in https://github.com/Azure/open-service-broker-azure/issues/566. It is also a child service like the original database-only service.
Several changes might look tricky. Point it out if it doesn't look good to you, I'd love to do refactor.

### DIFF
--- a/docs/modules/mssql.md
+++ b/docs/modules/mssql.md
@@ -1,14 +1,16 @@
 # [Azure SQL Database](https://azure.microsoft.com/en-us/services/sql-database/)
 
-Open Service Broker for Azure (OSBA) contains three Azure SQL Database services. These services enable you to select the most appropriate provisioning scenario for your needs. These services are:
+Open Service Broker for Azure (OSBA) contains five Azure SQL Database services. These services enable you to select the most appropriate provisioning scenario for your needs. These services are:
 
 | Service Name | Description |
 |--------------|-------------|
-| `azure-sql-12-0` | Provision both a SQL Server DBMS and a database. |
-| `azure-sql-12-0-dbms` | Provision only a SQL Server Database Management System (DBMS). This can be used to provision multiple databases at a later time. |
-| `azure-sql-12-0-database` | Provision a new database only upon a previously provisioned DBMS. |
+| [`azure-sql-12-0`](#service-azure-sql-12-0) | Provision both a SQL Server DBMS and a database. |
+| [`azure-sql-12-0-dbms`](#service-azure-sql-12-0-dbms) | Provision only a SQL Server Database Management System (DBMS). This can be used to provision multiple databases at a later time. |
+| [`azure-sql-12-0-database`](#service-azure-sql-12-0-database) | Provision a new database only upon a previously provisioned DBMS. |
+| [`azure-sql-12-0-dbms-registered`](#service-azure-sql-12-0-dbms-registered) | Register an existing server as a DBMS service instance. |
+| [`azure-sql-12-0-database-from-existing`](#service-azure-sql-12-0-database-from-existing) | Taking over an existing database upon a previous DBMS as a database service instance. The service requires `ENABLE_MIGRATION_SERVICES` to be `true` in OSBA environment variables. |
 
-The `azure-sql` service allows you to provision both a DBMS and a database. When the provision operation is successful, the database will be ready to use. You can not provision additional databases onto an instance provisioned through this service. The `azure-sql-dbms` and `azure-sql-database` services, on the other hand, can be combined to provision multiple databases on a single DBMS.  For more information on each service, refer to the descriptions below.
+The `azure-sql-12-0` service allows you to provision both a DBMS and a database. When the provision operation is successful, the database will be ready to use. You can not provision additional databases onto an instance provisioned through this service. The `azure-sql-12-0-dbms` and `azure-sql-12-0-database` services, on the other hand, can be combined to provision multiple databases on a single DBMS. The `azure-sql-12-0-dbms-registered` is in the same position with `azure-sql-12-0-dbms`. And `azure-sql-12-0-database-from-existing` is in the same position with `azure-sql-12-0-database`. For more information on each service, refer to the descriptions below.
 
 ## Services & Plans
 
@@ -478,3 +480,54 @@ curl -X PUT \
 }
 '
 ```
+
+### Service: azure-sql-12-0-dbms-registered
+
+It is to *register* an existing Azure SQL Server as a SQL DBMS service instance. Both **azure-sql-12-0-database** service and **azure-sql-12-0-database-from-existing** service can be its child service.
+
+#### Behaviors
+
+##### Provision
+
+Please refer to [**azure-sql-12-0-dbms** service](#service-azure-sql-12-0-dbms) with extra required provisioning parameters below.
+
+###### Provisioning Parameters
+
+| Parameter Name | Type | Description | Required | Default Value |
+|----------------|------|-------------|----------|---------------|
+| `server` | `string` | The SQL server name. | Y | |
+| `administratorLogin` | `string` | The administratorLogin input when creating the SQL server. | Y | |
+| `administratorLoginPassword` | `string` | The administratorLoginPassword input when creating the SQL server. | Y | |
+
+##### Update
+
+Update the `administratorLogin` and/or `administratorLoginPassword` as they may change and the server is assumed to be managed by yourself.
+
+###### Updating Parameters
+
+| Parameter Name | Type | Description | Required | Default Value |
+|----------------|------|-------------|----------|---------------|
+| `administratorLogin` | `string` | New administratorLogin. | N | |
+| `administratorLoginPassword` | `string` | New administratorLoginPassword. | N | |
+
+##### Bind, Unbind, Deprovision
+
+Please refer to [**azure-sql-12-0-dbms** service](#service-azure-sql-12-0-dbms).
+
+### Service: azure-sql-12-0-database-from-existing
+
+It is to create SQL database service instance from existing Azure SQL Database *for taking over the database*. Both **azure-sql-12-0-dbms** service and **azure-sql-12-0-dbms-registered** service can be its parent service.
+
+##### Provision
+
+Please refer to [**azure-sql-12-0-database** service](#service-azure-sql-12-0-database) with extra required provisioning parameters below.
+
+###### Provisioning Parameters
+
+| Parameter Name | Type | Description | Required | Default Value |
+|----------------|------|-------------|----------|---------------|
+| `database` | `string` | The SQL database name. | Y | |
+
+##### Update, Bind, Unbind, Deprovision
+
+Please refer to [**azure-sql-12-0-database** service](#service-azure-sql-12-0-database).
--- a/pkg/services/mssql/catalog.go
+++ b/pkg/services/mssql/catalog.go
@@ -5,6 +5,7 @@ import "github.com/Azure/open-service-broker-azure/pkg/service"
 func buildBasicPlan(
 	id string,
 	includesDBMS bool,
+	fe bool,
 ) service.PlanProperties {
 
 	planDetails := dtuPlanDetails{
@@ -17,7 +18,7 @@ func buildBasicPlan(
 		includeDBMS: includesDBMS,
 	}
 
-	return service.PlanProperties{
+	planProperties := service.PlanProperties{
 		ID:          id,
 		Name:        "basic",
 		Description: "Basic Tier, 5 DTUs, 2GB Storage, 7 days point-in-time restore",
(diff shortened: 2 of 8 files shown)


docs/modules/mssql.md
```diff
@@ -477,4 +479,55 @@ curl -X PUT \
...
+Update the `administratorLogin` and/or `administratorLoginPassword` as they may change and the server is assumed to be managed by yourself.
+
+###### Updating Parameters
+
+| Parameter Name | Type | Description | Required | Default Value |
+|----------------|------|-------------|----------|---------------|
+| `administratorLogin` | `string` | New administratorLogin. | N | |
+| `administratorLoginPassword` | `string` | New administratorLoginPassword. | N | |
+
+##### Bind, Unbind, Deprovision
```
norshtein: It's better to add the behavior of deprovisioning here. The same below with database from existing

pkg/services/mssql/mssql.go
```diff
@@ -50,21 +55,25 @@ func New(
 		armDeployer:          armDeployer,
 		serversClient:        serversClient,
 	}
+	databaseMan := databaseManager{
```
norshtein: Mgr is the abbr for manager.

zhongyi-zhang: Thanks, will change to this one!

docs/modules/mssql.md
zhongyi-zhang: Yeah, Will polish it. I forgot to clarify this in the doc.

pkg/services/mssql/plan_details.go
```diff
@@ -74,6 +75,19 @@ func (d dtuPlanDetails) getProvisionSchema() service.InputParametersSchema {
...
 
+func (d dtuPlanDetails) getFeProvisionSchema() service.InputParametersSchema {
+	ips := service.InputParametersSchema{
+		PropertySchemas: map[string]service.PropertySchema{},
+	}
+	if d.includeDBMS {
+		ips = getDBMSCommonProvisionParamSchema()
+	}
+	ips.PropertySchemas["database"] = &service.StringPropertySchema{
+		Description: "The name of the existing database",
```
norshtein: Missed `Title`.

norshtein: The same below.

tests/lifecycle/driver_test.go
```diff
@@ -50,6 +53,9 @@ func TestServices(t *testing.T) {
 }
 
 func TestMain(m *testing.M) {
+	if err := crypto.InitializeGlobalCodec(noop.NewCodec()); err != nil {
```
norshtein: What's the purpose of this statement?

zhongyi-zhang: It is to use `GetMapFromStruct` in mssql_cases_test.go. When `json.Marshal` hits our `SecureString` type, initialization for `crypto` is needed: https://github.com/Azure/open-service-broker-azure/blob/master/pkg/service/secure_string.go#L15.

pkg/services/mssql/plan_details.go
zhongyi-zhang: Thanks for the good catches. Fixed.

tests/lifecycle/driver_test.go
norshtein: Thanks for the explanation.

docs/modules/mssql.md
```diff
@@ -1,14 +1,16 @@
 # [Azure SQL Database](https://azure.microsoft.com/en-us/services/sql-database/)
 
-Open Service Broker for Azure (OSBA) contains three Azure SQL Database services. These services enable you to select the most appropriate provisioning scenario for your needs. These services are:
+Open Service Broker for Azure (OSBA) contains five Azure SQL Database services. These services enable you to select the most appropriate provisioning scenario for your needs. These services are:
```
jeremyrickard: maybe we should change this wording so that we don't need to update it every time we make changes here? Maybe "a number of" would be better.

docs/modules/mssql.md
```diff
@@ -477,4 +479,61 @@ curl -X PUT \
...
+
+Please refer to [**azure-sql-12-0-dbms** service](#service-azure-sql-12-0-dbms).
+
+##### Deprovision
+
+Do nothing. The SQL server would not be deleted in Azure.
+
+### Service: azure-sql-12-0-database-from-existing
+
+It is to create SQL database service instance from existing Azure SQL Database *for taking over the database*. Both **azure-sql-12-0-dbms** service and **azure-sql-12-0-dbms-registered** service can be its parent service.
```
jeremyrickard: Do we need to have both parent services? 

The `azure-sql-12-0-dbms` is managed by OSBA. If you're going to register some pre-existing database, is that really a scenario that you'll be encountering?

pkg/services/mssql/catalog.go
```diff
@@ -5,6 +5,7 @@ import "github.com/Azure/open-service-broker-azure/pkg/service"
 func buildBasicPlan(
 	id string,
 	includesDBMS bool,
+	fe bool,
```
jeremyrickard: what is the "fe" variable representing?

pkg/services/mssql/plan_details.go
```diff
@@ -74,6 +75,20 @@ func (d dtuPlanDetails) getProvisionSchema() service.InputParametersSchema {
 	return ips
 }
 
+func (d dtuPlanDetails) getFeProvisionSchema() service.InputParametersSchema {
```
jeremyrickard: Same question as above. If we onboard new developers here, it might be better to use a more descriptive name for this path? I'm not sure what Fe means.

jeremyrickard:
Hey @zhongyi-zhang, I have a few questions regarding this, could you let me know on the questions? Thanks!

pkg/services/mssql/plan_details.go
zhongyi-zhang: @jeremyrickard yeah, actually I am worried about the same thing. But sadly I don't have an idea to make it elegant. I use "Fe" as a key word to indicate that "this is a 'from-existing' services related item". Replacing with "FromExisting" also reads weird to me. And in this PR, we have to use it everywhere... Appreciate your thoughts on this...

docs/modules/mssql.md
zhongyi-zhang: Oh, I mean "either" here. I'll polish the words...

docs/modules/mssql.md
zhongyi-zhang: Thanks for the suggestion.

docs/modules/mssql.md
norshtein: The question is: will there have a `azure-sql-12-0-database-from-existing` instance whose parent is `azure-sql-12-0-dbms`? If so, it indicates the user provisions a `azure-sql-12-0-dbms` instance using OSBA and then creates a database on it outside OSBA. Will we meet this scenario?

zhongyi-zhang: @norshtein this is very rare but valid.

jeremyrickard: Yeah, that is my question. I feel like the more common scenario is existing DBMS and we want to provision new databases on that existing DBMS.  

An existing database probably exists on an existing DBMS. In that case, do the CF users really want to manage the DBMS and the DB via OSBA? Or do they want to _use_ the database via OSBA through normal binding operations?

We should maybe draw out exactly what scenarios are going to be supported? There are lots of different combinations, I'm not sure we should support all of them due to complexity.

Maybe these?

* OSBA provisions DBMS and DB (separately or together)
* User provisions DBMS, registers it with OSBA
* User uses OSBA to provision new databases on previously registered DBMS
* User registers existing database (can that simply be a stand alone service?)

jeremyrickard: > @norshtein this is very rare but valid.

Have you actually had users _ask_ for this use case? Or is this a case where you are just trying to cover the possibilities? I think it's a pretty odd use case tbqh.

pkg/services/mssql/plan_details.go
jeremyrickard: Maybe just make it lengthy so it's clear

getProvisionSchemaForExistingIntance()

it's long, but it has clarity.

docs/modules/mssql.md
zhongyi-zhang: I prefer just following the statement of `azure-sql-12-0-database`. The detail is confusing.

norshtein: If I understand this correctly, after user registers an dbms to OSBA, it's **nearly the same** between `azure-sql-12-0-dbms` and `azure-sql-12-0-dbms-registered`, except the behavior in deprovisioning is different. So we may not need to clarify every use case in the document.

So from user's perspective, there is no difference between creating a database on  `azure-sql-12-0-dbms` and creating a database on `azure-sql-12-0-dbms-registered`, the only thing they need to do is to specify `parentAlias`, regardless of the real type of parent instance. 

And for `azure-sql-12-0-database-from-existing`, similar logic comes: though it's very odd to make `azure-sql-12-0-dbms` to be the parent of `azure-sql-12-0-database-from-existing`, but it's valid, just as you can make `azure-sql-12-0-dbms-registered` to be the parent of `azure-sql-12-0-database-from-existing`.

zhongyi-zhang:
The last commit replaced all "Fe" with "ForExistingInstance" as a suffix. Does it look good to you? I am going to continue polishing the doc.

docs/modules/mssql.md
zhongyi-zhang: @norshtein that's correct! Now we consider how to describe it brief... as someone didn't understand PostgreSQL doc, the simpler one correctly and opened an issue.

docs/modules/mssql.md
```diff
@@ -477,4 +479,61 @@ curl -X PUT \
     }
 }
 '
-```
\ No newline at end of file
+```
+
+### Service: azure-sql-12-0-dbms-registered
+
+It is to *register* an existing Azure SQL Server as a SQL DBMS service instance. Both **azure-sql-12-0-database** service and **azure-sql-12-0-database-from-existing** service can be its child service.
```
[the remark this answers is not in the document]
zhongyi-zhang: I think we should add a separated page to describe the parent and child concept, and just refer the concept in module docs. That's the most straightforward statement to our design. Avoiding the extra clarification and chattering make it more confusing.

zhongyi-zhang:
Added a single page to clarify the parent-child model. What do you think of the commit? Your comments would be valuable as I am not a native English speaker...

docs/parent-child-model-for-multiple-layers-services.md
```diff
@@ -0,0 +1,17 @@
...
+
+For a parent service, `alias` is always one of the required provisioning parameters. It is the correlative key for a parent service instance to engage with child service instances. For easy memo, it is recommended to give the same value to the service instance name.
+
+## Child Service
+
+For a child service, `parentAlias` is always one of the required provisioning parameters. It should be specified an existing `alias` of a parent service instance. A child service instance can only be created successfully in this way.
+
+## Notes
+
+* A parent service instance can have multiple children. That is the way to achieve that, databases share a DBMS.
```
norshtein: How about change the second sentence to "For example, a dbms instance can have several database instance as its children".

docs/parent-child-model-for-multiple-layers-services.md
```diff
@@ -0,0 +1,17 @@
...
+
+* A service module can contain several parent services. For example MS SQL service module, parent DBMS service instances can be created by OSBA, and you can bring your own Azure SQL servers to register them as parent service instances.
```
norshtein: Why mention "A service module can contain several parent services"? Nothing prevents this from happening, and users are not likely to think that only one parent service can appear in a module.

docs/parent-child-model-for-multiple-layers-services.md
```diff
@@ -0,0 +1,17 @@
...
+
+## Parent Service
+
+For a parent service, `alias` is always one of the required provisioning parameters. It is the correlative key for a parent service instance to engage with child service instances. For easy memo, it is recommended to give the same value to the service instance name.
+
+## Child Service
+
+For a child service, `parentAlias` is always one of the required provisioning parameters. It should be specified an existing `alias` of a parent service instance. A child service instance can only be created successfully in this way.
+
+## Notes
```
norshtein: One thing may be confusing to users, and it can be added to somewhere in the doc:
- On instance level, parent-child is a "one-to-many" relationship. That is, for a parent instance, it can have several child instance, but for a child instance, it has and only has one parent instance.
- On service level, parent-child is a "many-to-many" relationship. That is, a parent service has several valid child service, and a child service has several valid parent service.
For example, service A can be parent of service C,D; service B can be parent of service D,E. Note here both service A and B can be parent service of service D. However, for an instance of service D, it can only have one parent instance, either an instance of service A or an instance of service B is OK, but it can't have two parent instance.

norshtein: If it is abstract to users, we may add this statement to MSSQL module and use specific service name. After all, only MSSQL module has "many-to-many" parent-child relationship on service level currently.

docs/parent-child-model-for-multiple-layers-services.md
zhongyi-zhang: It is a natural description to hint the "many-to-many" relationship to me. Do you agree?

norshtein: Yeah, it's actually a hint, but if we decide to describe the "many-to-many" in this general parent-child-model doc, a more detailed description should be provided on "many-to-many" mechanism, or we shouldn't hint it and just explain it in MSSQL doc.

zhongyi-zhang:
@jeremyrickard does it look good to you now?

Issue #607, opened 2018-10-09
Changes on storage module


norshtein:
Several enhancements can be added to storage module:
- Add the ability to provision storage general purpose v2
- Make several properties configurable:
  - `accountType`. This is the name of the sku. Allowed values: [Standard_LRS, Standard_GRS, Standard_RAGRS, Standard_ZRS, Premium_LRS, Premium_ZRS]
  - `accessTier`. Allowed values: [Hot, Cool]. Hot storage is optimized for storing data that is accessed frequently ,and cool storage is optimized for storing data that is infrequently accessed and stored for at least 30 days.
  - `supportsHttpsTrafficOnly`. Allows https traffic only to storage service if sets to true.
- Make service instance updatable.
- Add several fields to credential:
  - primaryBlobServiceEndPoint
  - primaryFileServiceEndPoint
  - primaryQueueServiceEndPoint
  - primaryTableServiceEndPoint

I'll separate these features into several PRs as in Redis module.

jeremyrickard:
I think maybe as part of this work, it might be good to investigate if more info is needed in the credential?

See: https://github.com/Azure/open-service-broker-azure/issues/578

In the portal, if you go to Properties for a storage account, you can get things like "Primary Table Service Endpoint" and "PRIMARY BLOB SERVICE ENDPOINT". I think you can infer these things from the _name_ of the storage account, but it might be good to provide these to make life easier for the OSBA user? I believe that is the info the user was looking for in #578 .

norshtein:
Oh, thanks for the tips! Agree that it should be added to the credential.

PR #608, opened 2018-10-10
Remove outdated property in arm template and add fields in credential

norshtein:
This is one of PRs for implementing #607


norshtein:
Explanation for removing `encryption` in `properties`: currently, in Azure Portal, it's not configured and encryption is on by default.

Issue #609, opened 2018-10-10
Document for setting up Azure Redis is incorrect


norshtein:
See: https://github.com/Azure/open-service-broker-azure/tree/master/contrib/cf#create-an-azure-redis-cache

This guidance sets up an Azure Redis cache with `Basic` tier, which is insufficient in production, see [here](https://azure.microsoft.com/en-us/pricing/details/cache/). We have suffered a data loss using `Basic` tier cache.

norshtein:
Had a quick look on k8s guidance and didn't find any section for creating Azure Redis cache. It should be a CF-only doc error.

zhongyi-zhang:
This is a definite lethal. We should add a conspicuous note. @krancour @jeremyrickard is the helm chart aware of it?

zhongyi-zhang:
I notice that the embedded Redis can't be used in production: https://github.com/Azure/open-service-broker-azure/tree/master/contrib/k8s/charts/open-service-broker-azure#configuration. Is it still recommended to create an Azure Redis for production?

PR #610, opened 2018-10-10
Fix doc for setting up Azure Redis cache

norshtein:


norshtein:
Fix #609

zhongyi-zhang:
I think a conspicuous note should be added to have users take care of this.

norshtein:
Thanks! Warning message is added.

PR #614, opened 2018-10-12
Add `tags` field in several modules' schema

### LINKED ISSUE #613: Some modules' ProvisioningParametersSchema miss `tags` property
Except for MSSQL, MySQL, PostgreSQL, all of other modules miss `tags` property in ProvisioningParametersSchema(and UpdatingParametersSchema, if updatable). For example, [redis](https://github.com/Azure/open-service-broker-azure/blob/master/pkg/services/rediscache/plan_schemas.go#L18), [cosmos](https://github.com/Azure/open-service-broker-azure/blob/master/pkg/services/cosmosdb/plan_schemas.go#L125).

norshtein:
Fix #613

### DIFF
--- a/pkg/services/cosmosdb/plan_schemas.go
+++ b/pkg/services/cosmosdb/plan_schemas.go
@@ -16,6 +16,12 @@ func generateUpdatingParamsSchema() service.InputParametersSchema {
 	const maxIntervalInSecondsMax = 86400
 	return service.InputParametersSchema{
 		PropertySchemas: map[string]service.PropertySchema{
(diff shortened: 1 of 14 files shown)


zhongyi-zhang:
Can we move all the tag properties to https://github.com/Azure/open-service-broker-azure/blob/master/pkg/service/schema.go#L538?

norshtein:
This is not a parameter that every service can have. For example, when provisioning a database, `tags` cannot be specified, so it's not proper to add `tags` to `AddCommonSchema`

norshtein:
Please hold on PRs until #617 is solved.

PR #619, opened 2018-10-15
Tune service catalog install command for AKS

carolynvs:
* Disable health checks by default to workaround https://github.com/Azure/AKS/issues/417
* Lower logging threshold (from 10 to 2) so that the logs don't include every successful HTTP request/response (which is what you get at 4 which unfortunately is where Service Catalog logs its info messages).


carolynvs:
I've created a follow-on issue in service catalog, https://github.com/kubernetes-incubator/service-catalog/issues/2415, to address the difficulty of selecting a useful log level.

jeremyrickard:
Docs only, thanks for the PR @carolynvs

PR #622, opened 2018-10-16
[WIP] Add V2 GUID to make OSBA co-exist with MASB

zhongyi-zhang:
The 4th item in https://github.com/Azure/open-service-broker-azure/issues/566.

It is what we discussed in the mail thread and the meeting. The implementation might be disputable. To me, this is the minimal change in OSBA to achieve co-existence with MASB but not so elegant. @jeremyrickard @norshtein @krancour appreciate reviews from you all!


zhongyi-zhang:
Oh, it doesn't solve the conflictive service names issue. Need one more commit to add V2 service names. Anyway, it also follows the idea. I will work on it tomorrow.

krancour:
> Since azure-rediscache is still preview, the others are experimental, can we make a little change to azure-redis-cache, azure-serivce-bus, azure-event-hubs, like the change did on MS SQL -- azure-sqldb -> azure-sql -> azure-sql-12-0?

@zhongyi-zhang you can change names on anything that isn't stable.

I think including versions in service names is a good idea. (I won't bore anyone with an explanation for why, but I'll elaborate if anyone is curious.) So this get's a thumbs up from me.

pkg/service/catalog_config.go
```diff
@@ -12,12 +12,14 @@ import (
...
 type CatalogConfig struct {
 	MinStability            Stability
+	UseV2Guid               bool
 	EnableMigrationServices bool
 }
 
 type tempCatalogConfig struct {
 	CatalogConfig
 	MinStabilityStr            string `envconfig:"MIN_STABILITY" default:"PREVIEW"`
+	UseV2GuidStr               string `envconfig:"USE_V2_GUID" default:"false"`
```
krancour: _Naming things is hard._ I don't want to sound nitpicky, but the attribute name `UseV2GuidStr` and environment variable name `USE_V2_GUID` don't feel right to me.

I have two reasons:

1. I think the meaning of this flag isn't obvious. "v2" makes this sound in some way like it's newer or superior to what came before. That can be a problem because it could lead to people who haven't ready the documentation thoroughly opting into this "just because." And we don't want that.

2. The guids used throughout OSBA are based on version 4 of the UUID spec. Using "v2" here might falsely suggest the use of UUIDv2.

I would suggest we use attribute / environment variable named in a way that is something more like "enable masb coexistence mode." This is more clear as to its purpose.

zhongyi-zhang:
@krancour thanks for the review! Actually @norshtein brings a good news from PCF team just today. CF CC will allow registering even the same catalog twice by the Multi-service-registration epic: https://www.pivotaltracker.com/n/projects/2105761/stories/161050836.
So, we can leave the PR here. I will test the co-existence scenario once the feature is available. If we verify it does work, this PR can be closed. That would be ideal!

PR #624, opened 2018-10-17
Create a new module for Azure SQL Failover Group and add the first parent service "azure-sql-12-0-dr-dbms-pair-registered"

zhongyi-zhang:
Kick off the failover group part in https://github.com/Azure/open-service-broker-azure/issues/566!

### DIFF
--- a/pkg/services/mssqldr/catalog.go
+++ b/pkg/services/mssqldr/catalog.go
@@ -22,10 +22,7 @@ func (m *module) GetCatalog() (service.Catalog, error) {
 				},
 				Bindable: false,
 				Tags: []string{
-					"AzurePublicCloud",
-					"AzureUSGovernmentCloud",
-					"AzureChinaCloud",
-					"AzureGermanCloud",
+					"Azure",
 					"SQL",
 					"DBMS",
 					"Server",
--- a/pkg/services/mssqldr/catalog.go
+++ b/pkg/services/mssqldr/catalog.go
@@ -1,0 +1,58 @@
+package mssqldr
+
+import "github.com/Azure/open-service-broker-azure/pkg/service"
+
+// nolint: lll
+func (m *module) GetCatalog() (service.Catalog, error) {
+
+	return service.NewCatalog([]service.Service{
+		// dbms pair registered service
+		service.NewService(
+			service.ServiceProperties{
+				ID:             "00ce53a3-d6c3-4c24-8cb2-3f48d3b161d8",
+				Name:           "azure-sql-12-0-dr-dbms-pair-registered",
+				Description:    "Azure SQL 12.0-- disaster recovery DBMS pair registered",
+				ChildServiceID: "2eb94a7e-5a7c-46f9-b9d2-ff769f215845", // More children in fact
+				Metadata: service.ServiceMetadata{
+					DisplayName:      "Azure SQL 12.0-- disaster recovery DBMS Pair registered",
+					ImageURL:         "https://azure.microsoft.com/svghandler/sql-database/?width=200",
+					LongDescription:  "Azure SQL 12.0-- disaster recovery DBMS pair registered, as the primary server and the secondary server of failover groups",
+					DocumentationURL: "https://docs.microsoft.com/en-us/azure/sql-database/",
+					SupportURL:       "https://azure.microsoft.com/en-us/support/",
+				},
+				Bindable: false,
+				Tags: []string{
+					"AzurePublicCloud",
+					"AzureUSGovernmentCloud",
+					"AzureChinaCloud",
+					"AzureGermanCloud",
+					"SQL",
+					"DBMS",
+					"Server",
+					service.DRTag,
+					"Failover Group",
+				},
+				Extended: map[string]interface{}{
+					"version": "12.0",
+				},
+			},
+			m.dbmsPairRegisteredManager,
+			service.NewPlan(service.PlanProperties{
+				ID:          "5683ca92-372b-49a6-b7cd-96a14645ec15",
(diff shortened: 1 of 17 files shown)


pkg/services/mssqldr/catalog.go
```diff
@@ -0,0 +1,58 @@
...
+				Metadata: service.ServiceMetadata{
+					DisplayName:      "Azure SQL 12.0-- disaster recovery DBMS Pair registered",
+					ImageURL:         "https://azure.microsoft.com/svghandler/sql-database/?width=200",
+					LongDescription:  "Azure SQL 12.0-- disaster recovery DBMS pair registered, as the primary server and the secondary server of failover groups",
+					DocumentationURL: "https://docs.microsoft.com/en-us/azure/sql-database/",
+					SupportURL:       "https://azure.microsoft.com/en-us/support/",
+				},
+				Bindable: false,
+				Tags: []string{
+					"AzurePublicCloud",
```
norshtein: I think tags `Azure*Cloud` should be removed.

norshtein:
Better to add a doc in the PR. If you want to add a summarized doc after several PRs are merged, please add explanation about what's the functionality of `dbmsPairRegisteredManager`.

pkg/services/mssqldr/catalog.go
zhongyi-zhang: Yeah, thanks for the good catch.

zhongyi-zhang:
Sure. Will add a comment for `dbmsPairRegisteredManager`.

norshtein:
It's better to add the behavior of provision/bind/unbind/deprovision/update

zhongyi-zhang:
OK, I added a draft doc instead for the overview. Will continue polishing it according to the reviews during addressing the services in this module.

pkg/services/mssqldr/common.go
```diff
@@ -0,0 +1,44 @@
...
+	"net/url"
+
+	_ "github.com/denisenkom/go-mssqldb" // MS SQL Driver
+)
+
+func getDBConnection(
+	administratorLogin string,
+	administratorLoginPassword string,
+	fullyQualifiedDomainName string,
+	databaseName string, // nolint: unparam
```
norshtein: Why use `nolint: unparam` here? Parameter `databaseName` is used in the function and shouldn't cause a lint issue.

zhongyi-zhang: The linter reports a warning when a parameter always receives the same value in a function. Now the child services are not included in the module, the value of `databaseName` is always `master`.

pkg/services/mssqldr/dbms_pair_registered_update.go
```diff
@@ -0,0 +1,40 @@
...
+) error {
+	return nil
+}
+
+func (d *dbmsPairRegisteredManager) GetUpdater(
+	service.Plan,
+) (service.Updater, error) {
+	return service.NewUpdater(
+		service.NewUpdatingStep("updateAdministrators", d.updateAdministrators),
+		service.NewUpdatingStep("testPriConnection", d.testPriConnection),
```
norshtein: Here exists the same potential bug as I raised in #570 . The user may provide wrong password by mistake, which will cause step testConnection fails, but the wrong parameter will be written to store in step updateAdministrator and previous correct username and password will be overwritten.

pkg/services/mssqldr/mssqldr.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+// New returns a new instance of a type that fulfills the service.Module
+// interface and is capable of provisioning MS SQL servers and databases
+// using "Azure SQL Database"
+func New(
+	azureEnvironment azure.Environment,
+	armDeployer arm.Deployer,
+	serversClient sqlSDK.ServersClient,
+	databasesClient sqlSDK.DatabasesClient, // nolint: unparam
```
norshtein: I guess parameter `databasesClient` and `failoverGroupsClient` be used in later submitted PRs? Better to add a comment here.

pkg/services/mssqldr/plan_details.go
```diff
@@ -0,0 +1,96 @@
...
+			"secondaryServer",
+			"secondaryAdministratorLogin",
+			"secondaryAdministratorLoginPassword",
+		},
+		SecureProperties: []string{
+			"primaryAdministratorLoginPassword",
+			"secondaryAdministratorLoginPassword",
+		},
+		PropertySchemas: map[string]service.PropertySchema{
+			"primaryResourceGroup": &service.StringPropertySchema{
```
norshtein: `Title` is missed, the same in below schemas.

pkg/services/mssqldr/common.go
norshtein: Thanks for the explanation!

pkg/services/mssqldr/dbms_pair_registered_update.go
zhongyi-zhang: Yes, I almost forgot it. Thanks for raising it again.

zhongyi-zhang:
Got error "Microsoft Internal subscriptions are restricted from provisioning in this region. Please choose a different region.".  Tried another two and rerun now.

zhongyi-zhang:
@krancour @jeremyrickard could you help get another available location for the CI subscription to provision SQL Servers? Southcentralus is available. The cases in this module need one more. I can only guess and try...

zhongyi-zhang:
It is only Microsoft Internal subscriptions constraint. External users won't run into it.

norshtein:
Got it. Thanks for the explanation.

PR #626, opened 2018-10-18
Add temporary fix for CosmosDB lifecycle test

norshtein:
See IcM ticket in #617 . CosmosDB team is fixing the error. They suggest us "While we are working to fix this, please add a delay of 10s in between region changes succeeding and sending delete request." I make the go routine sleep for 20 seconds after all regions are created. I'll follow up on the issue after the bug is fixed by CosmosDB team.


norshtein:
Hi @zhongyi-zhang , could you please have a review on this? I'd like to merge this PR ASAP for not blocking our development.

norshtein:
Explanation of sleeping for 20s instead of 10s: tested locally, sleeping for 10s is not enough sometimes.

PR #628, opened 2018-10-24
Make account type in storage module configurable and updatable

### LINKED ISSUE #607: Changes on storage module

norshtein:
Close #607 . This is the late piece of code for implementing #607 . This PR makes account type in storage module configurable and updatable. One thing to note: if `accountType` is set to `Premium_LRS`, only "Hot" can be used in `accessTier`. But currently there is no convenient way to validate this, so I only add this constraint to doc. Shoule we reconsider adding `ValidateProvisioningParameters`?


docs/modules/storage.md
```diff
@@ -28,7 +28,8 @@ only, or an account with a container.
...
+| `accountType` | `string` | A combination of account kind and   replication strategy. Allowed values: for plan `blob-storage-account` and `blob-container`: [ "Standard_LRS", "Standard_GRS", "Standard_RAGRS"]; for plan `general-purpose-storage-account`: ["Standard_LRS", "Standard_GRS", "Standard_RAGRS", "Premium_LRS"]; for plan `general-purpose-v2-storage-account`: ["Standard_LRS", "Standard_GRS", "Standard_RAGRS", "Standard_ZRS", "Premium_LRS"]. Check [here](https://docs.microsoft.com/en-us/azure/storage/common/storage-redundancy#choosing-a-replication-option) for detailed explanation of replication strategy. **Note**: ZRS is only available in several regions, check [here](https://docs.microsoft.com/en-us/azure/storage/common/storage-redundancy-zrs#support-coverage-and-regional-availability) for allowed regions to use ZRS. | N | If not provided, "Standard_LRS" will be used as the default value for all plans. |
```
zhongyi-zhang: Too many words... What about just giving the super set of the allowed values instead of listing them all? The linked doc tells the details.

docs/modules/storage.md
```diff
@@ -66,7 +67,8 @@ Updates an existing storage account.
...
 | ` enableNonHttpsTraffic ` | `string`            | Specify whether non-https traffic is enabled. Allowed values:["enabled", "disabled"]. | N        |
 | `tags`                    | `map[string]string` | Tags to be applied to new resources, specified as key/value pairs. | N        |
-| `accessTier`              | `string`            | The access tier used for billing.    Allowed values: ["Hot", "Cool"]. Hot storage is optimized for storing data that is accessed frequently ,and cool storage is optimized for storing data that is infrequently accessed and stored for at least 30 days. **Note** : this field doesn't exist for plan `general-purpose-storage-account`. | N        |
+| `accessTier`              | `string`            | The access tier used for billing.    Allowed values: ["Hot", "Cool"]. Hot storage is optimized for storing data that is accessed frequently ,and cool storage is optimized for storing data that is infrequently accessed and stored for at least 30 days. **Note1** : this field doesn't exist for plan `general-purpose-storage-account`. **Note2** : this field can only set to "Hot" if you use "Premium_LRS" `accountType` | N        |
```
zhongyi-zhang: What about "doesn't exist" -> "is invalid"? And note2 -> "Premium_LRS" account only supports "Hot" in this field?

docs/modules/storage.md
```diff
@@ -66,7 +67,8 @@ Updates an existing storage account.
 | ------------------------- | ------------------- | ------------------------------------------------------------ | -------- |
 | ` enableNonHttpsTraffic ` | `string`            | Specify whether non-https traffic is enabled. Allowed values:["enabled", "disabled"]. | N        |
 | `tags`                    | `map[string]string` | Tags to be applied to new resources, specified as key/value pairs. | N        |
```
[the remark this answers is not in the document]
norshtein: However, if so, we will need to copy and paste this row every time we add a new property. How about move this line to the next line to `resourceGroup` in provisioning, and move this line to the top in updating?

docs/modules/storage.md
norshtein: Good suggestion. I will change it.

pkg/services/storage/plan_schemas.go
```diff
@@ -54,6 +54,22 @@ func (pd planDetail) generateProvisioningParamsSchema() service.InputParametersS
...
+		DefaultValue:  "Standard_LRS",
+		AllowedValues: []string{"Standard_LRS", "Standard_GRS", "Standard_RAGRS"},
+	}
+	if pd.planName == generalPurposeV1 {
+		sps := ips.PropertySchemas["accountType"].(*service.StringPropertySchema)
+		sps.AllowedValues = append(sps.AllowedValues, "Premium_LRS")
+	} else if pd.planName == generalPurposeV2 {
+		sps := ips.PropertySchemas["accountType"].(*service.StringPropertySchema)
+		sps.AllowedValues = append(sps.AllowedValues, "Premium_LRS", "Standard_ZRS")
+	}
```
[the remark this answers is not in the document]
norshtein: Yes, agree it's a better way. I'll change the code.

docs/modules/storage.md
norshtein: What do you mean by "giving the super set of the allowed values"?

docs/modules/storage.md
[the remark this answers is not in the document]
norshtein: ... OK, I'm using a markdown editor, so I'll need to copy and paste this row every time if it is at the end of the doc. If you're using plain text editor, there is no such problem. But please consider the suggestion...

docs/modules/storage.md
zhongyi-zhang: Give all the allowed values (a.k.a. the set of gp-v2). But the user should jump to the linked doc to confirm if the account type supports it or not. After all we have a default value for it, no need to attract readers' attention...

norshtein: It makes sense. I'll change it.

docs/modules/storage.md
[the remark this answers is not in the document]
norshtein: Now I think it's reasonable to make `tags` at the end of the table... I'll change it.

jeremyrickard:
LGTM. Running CI.

PR #629, opened 2018-11-01
Fix missing location validator in mssql-dbms-registered

zhongyi-zhang:


zhongyi-zhang:
Of course yes. Actually the non-registered dbms service applies this validator. It is a missing in registered dbms service.
Even though there is exception, I think this validator is mainly for spelling checking.

PR #631, opened 2018-11-06
Add database pair service as the child service in mssqldr module

zhongyi-zhang:
Some codes are quite similar to the single mssql module. This module doubles them, adds some validations, and the failover group.


pkg/services/mssqldr/common_provision.go
```diff
@@ -68,3 +71,185 @@ func validateServer(
...
+	p := map[string]interface{}{}
+	p["databaseName"] = databaseName
+	for key, value := range td {
+		p[key] = value
+	}
+	return p, nil
+}
+
+func deployDatabaseARMTemplate(
+	armDeployer *arm.Deployer,
```
norshtein: Why `armDeployer` is a pointer? Seems there is no change to the object. Call by value is enough...?

zhongyi-zhang: Yeah, no change. When I wrote these, I thought `arm.Deployer` is big enough to pass by pointer instead of value. Does it make sense?

PR #637, opened 2018-11-13
Add database pair registered service

zhongyi-zhang:
The third step in https://github.com/Azure/open-service-broker-azure/issues/566 for MSSQL DR.

### DIFF
--- a/pkg/services/mssqldr/catalog.go
+++ b/pkg/services/mssqldr/catalog.go
@@ -327,5 +327,63 @@ func (m *module) GetCatalog() (service.Catalog, error) {
 				),
 			),
 		),
+		// database pair registered service
+		service.NewService(
+			service.ServiceProperties{
+				ID:              "8480271a-f4c7-4232-b2b7-7f33391728f7",
+				Name:            "azure-sql-12-0-dr-database-pair-registered",
+				Description:     "Azure SQL 12.0-- disaster recovery database pair registered",
+				Bindable:        true,
+				ParentServiceID: "00ce53a3-d6c3-4c24-8cb2-3f48d3b161d8",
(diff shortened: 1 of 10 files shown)


tests/lifecycle/driver_test.go
```diff
@@ -50,6 +52,9 @@ func TestServices(t *testing.T) {
 }
 
 func TestMain(m *testing.M) {
+	if err := crypto.InitializeGlobalCodec(noop.NewCodec()); err != nil {
+		os.Exit(-1)
+	}
```
zhongyi-zhang: It is something we ever mentioned in https://github.com/Azure/open-service-broker-azure/pull/598. But the PR is still pending.

pkg/services/mssqldr/catalog.go
```diff
@@ -327,5 +327,63 @@ func (m *module) GetCatalog() (service.Catalog, error) {
...
+				Name:        "database",
+				Description: "Azure SQL Server-- database",
+				Free:        false,
+				Stability:   service.StabilityExperimental,
+				Metadata: service.ServicePlanMetadata{
+					DisplayName: "Azure SQL Server-- database",
+				},
+				Schemas: service.PlanSchemas{
+					ServiceInstances: service.InstanceSchemas{
+						ProvisioningParametersSchema: service.InputParametersSchema{
```
norshtein: Using a function will make it easier to maintain, just as `m.dbmsPairRegisteredManager.getProvisionParametersSchema()`

zhongyi-zhang: Thanks. Fixed.

zhongyi-zhang:
Thanks for the review.

Issue #42, opened 2017-11-02
Add Application Insights Module


arschles:

kyschouv:
Any ETA on this? We'd like to use OSBA in a project, and this is the one thing missing.

zhongyi-zhang:
Hi @kyschouv, we haven't scheduled for the module yet. Could you share the expectation of the module?

kyschouv:
It would create an application insights resource (preferably with a specified resource name, so it could be easily found), and put the instrumentation key in a secret (through a binding perhaps? Similar to how Redis works).

zhongyi-zhang:
Do we assume that the app is ready to enable App Insights, the only thing it required is the _instrumentation key_ to setup the client?

zhongyi-zhang:
One more question, do you want to specify any more App Insights properties besides the kind (web, java, etc.)? Ref: https://docs.microsoft.com/en-us/rest/api/application-insights/components/createorupdate#request-body.

kyschouv:
Yah we only ever need the instrumentation key (at least in our case). This is the ARM template we usually use when we're using app services (the `applicationId` probably isn't necessary):

```
    {
      "name": "[parameters('appInsightsName')]",
      "type": "Microsoft.Insights/components",
      "location": ["[variables('appInsightsLocation')]",](url)
      "apiVersion": "2014-04-01",
      "dependsOn": [],
      "properties": {
        "applicationId": "[variables('serviceId')]"
      }
    }
```

zhongyi-zhang:
Cool. Thanks for the info. The requirement looks not complicated. We can add an experimental module for it. What is the acceptable ETA for you? Loose ETA is the end of next month for me. (Sorry I have a tight timeline recently.) If this is urgent for you, I am pleased making my time to deliver it in two weeks (if the code review also goes smooth).

kyschouv:
It's not particularly urgent - end of next month would give us time to implement and test, as well as set up alerts against the resources. It'll be important to be able to specify the resource name for the app insights instances so we can find them and interact with them within the portal.

zhongyi-zhang:
Nice. I'll be back to the thread when it is done.

zhongyi-zhang:
@kyschouv the service is supported as an experimental module now. Have a try! Your feedback are welcome.
I am going to close the issue. Please create new issue if you have any enhancement request.

PR #640, opened 2018-11-16
Add experimental app insights module

zhongyi-zhang:
Solve https://github.com/Azure/open-service-broker-azure/issues/42


docs/modules/appinsights.md
```diff
@@ -0,0 +1,50 @@
...
+
+_Note: This module is EXPERIMENTAL. To enable this module, you must run Open Service Broker for Azure with modules.minStability set to `experimental`_
+
+## Services & Plans
+
+### Service: azure-appinsights
+
+| Plan Name | Description |
+|-----------|-------------|
+| `default` | For general application. |
```
norshtein: Noticed there are several types of application, why only `general` is added?

pkg/services/appinsights/types.go
```diff
@@ -0,0 +1,21 @@
+package appinsights
+
+import "github.com/Azure/open-service-broker-azure/pkg/service"
+
+type instanceDetails struct {
+	ARMDeploymentName  string               `json:"armDeployment"`
+	AppInsightsName    string               `json:"appInsightsName"`
+	InstrumentationKey service.SecureString `json:"instrumentationKey"`
```
norshtein: The type of `InstrumentationKey` cannot be `SecureString`. When marshaling to JSON, type `SecureString` will be encrypted, that is, the user will get a encrypted key.

zhongyi-zhang: In binding, it will be converted to a string. What I not sure is that, is it secret? We don't need use SecureString if not.

norshtein: 1. I don't think it is a secret, the value of `InstrumentationKey` is easily found at the main page of an app insight instance in Azure Portal. If it is a secret, it shouldn't be put at the main page.
2. The only difference between `string` and `SecureString` is when converting to JSON, `SecureString` will be encrypted and `string` will not. In memory, they're both bare non-encrypted string. So if `SecureString` is converted to `string` when binding, maybe use `string` at beginning is a better practice.

zhongyi-zhang: Consulted @HDYA, `InstrumentationKey` is like ID+KEY to the app insights. That makes sense -- anyone can use the app insights instance if he has this key. So, we should use `SecureString` for this field, right?

docs/modules/appinsights.md
zhongyi-zhang: Ah, I will add others as plans.

pkg/services/appinsights/types.go
norshtein: It makes sense. Thanks for the explanation.

PR #642, opened 2018-11-19
Add blob container service in storage module

norshtein:


pkg/services/storage/catalog.go
```diff
@@ -180,5 +180,41 @@ func (m *module) GetCatalog() (service.Catalog, error) {
...
 			}),
 		),
+		service.NewService(
+			service.ServiceProperties{
+				ID:              "fb6ce656-c16d-4b48-aff9-286714298af8",
+				Name:            serviceBlobContainer,
+				Description:     "A blob container inside an existing blob storage account",
+				ParentServiceID: "1a5b4582-29a3-48c5-9cac-511fd8c52756",
+				Metadata: service.ServiceMetadata{
+					DisplayName: "Azure Storage",
```
zhongyi-zhang: If I remember correctly, PCF Apps Manager (maybe more tools) shows this name in the UI. Users would get several different services with the same name.

pkg/services/storage/catalog.go
```diff
@@ -180,5 +180,41 @@ func (m *module) GetCatalog() (service.Catalog, error) {
...
+					ImageURL: "https://azure.microsoft.com/svghandler/storage/" +
+						"?width=200",
+					LongDescription: "Offload the heavy lifting of datacenter management" +
```
zhongyi-zhang: Same here. Referring some child service specific description would be better.

pkg/services/storage/catalog.go
zhongyi-zhang: Please also take a look at the field in other services in this module. That might be ignored in the last PR :).

pkg/services/storage/utils.go
```diff
@@ -30,3 +33,23 @@ func buildGoTemplate(
...
 }
+
+func createBlobContainer(storageAccountName, accessKey, containerName string) error { // nolint: lll
+	client, _ := storage.NewBasicClient(storageAccountName, accessKey)
+	blobCli := client.GetBlobService()
+	container := blobCli.GetContainerReference(containerName)
+	options := storage.CreateContainerOptions{
+		Access: storage.ContainerAccessTypePrivate,
+	}
+	_, err := container.CreateIfNotExists(&options)
```
zhongyi-zhang: As it is a child service instead of "allInOne", the name could be conflictive among child instances. I think it should be not use `Create"IfNotExists"`.

zhongyi-zhang: And I think fixing the method using here is not enough -- if the provision failed here, triggering deprovision, the existing container would be deleted.

norshtein: Agree with you. It is tricky, thanks for the reminding.

norshtein:
Review applied. Please have a review. @zhongyi-zhang

PR #644, opened 2018-11-20
Add MSSQL DR service: database-pair-from-existing-primary

zhongyi-zhang:
The last service piece of DR section in https://github.com/Azure/open-service-broker-azure/issues/566.
After this PR getting merged, I'll focus on the documentation for the module.


pkg/services/mssqldr/database_pair_provision_for_existing_primary.go
```diff
@@ -0,0 +1,85 @@
...
+			"deployPriARMTemplateForExistingInstance",
+			d.deployPriARMTemplateForExistingInstance,
+		),
+		service.NewProvisioningStep(
+			"deployFailoverGroupARMTemplateForExistingInstance",
+			d.deployFailoverGroupARMTemplateForExistingInstance,
+		),
+		service.NewProvisioningStep(
+			"deploySecARMTemplateForExistingInstance",
+			d.deploySecARMTemplateForExistingInstance,
```
norshtein: According to the doc, when using `azure-sql-12-0-dr-database-pair-from-existing-primary`, the user only has one primary database, and the secondary does not exist yet. Why use `deploySecARMTemplateForExistingInstance` here?

zhongyi-zhang: The `azure-sql-12-0-dr-database-pair` uses it, too. Becauase https://github.com/Azure/open-service-broker-azure/blob/master/pkg/services/mssqldr/database_pair_provision.go#L29-L30. I'll also add the comment here.

norshtein: It makes sense. Thanks for the explanation.

PR #645, opened 2018-11-20
Add multipleWriteRegionsEnabled attribute

norshtein:


docs/modules/cosmosdb.md
```diff
@@ -29,7 +29,8 @@ provisions an empty Database. Ready to use with existing Azure CosmosDB librarie
...
-| `autoFailoverEnabled` | `string ` | Specifies if you want Cosmos DB to perform automatic failover of the write region to one of the read regions in the rare event of a data center outage. Valid values are ""(unspecified), enabled, or disabled. | N | If not specified, default "disabled". |
+| `multipleWriteRegionsEnabled` | `string` | Specifies if you want  the account to write in multiple regions. Valid values are [ "enabled", "disabled"]. If set to "enabled", regions in `readRegions`  will also be writable. | N | If not specified, "disabled" will be used as the default value. |
+| `autoFailoverEnabled` | `string ` | Specifies if you want Cosmos DB to perform automatic failover of the write region to one of the read regions in the rare event of a data center outage. Valid values are [ "enabled", "disabled"]. **Note**: If `multipleWriteRegionsEnabled` is set to `enabled`, all regions will be writable, and this attribute will not work. | N | If not specified, default "disabled". |
```
zhongyi-zhang: The "default value" column can be just the default value without "If not specified ...", right?

zhongyi-zhang: Will the REST API return error with both `autoFailoverEnabled` and `multipleWriteRegionsEnabled` enabled?

norshtein: The second question: tested, won't produce an error.

norshtein: And the first question: yes, "if not specified" can be dropped sometimes... But consider the following sentence: `If not specified, no replication region will be created.` "If not specified" makes the sentence reads more clear. It is not a fatal error, how about just keep all "if not specified"?

Issue #646, opened 2018-11-21
MS SQL Server should support setting connection policy


zhongyi-zhang:

zhongyi-zhang:
@HDYA done!

HDYA:
wooooooooooow @zhongyi-zhang Brilliant!

PR #648, opened 2018-12-03
MSSQL DBMS supports setting connection policy

zhongyi-zhang:
Solve https://github.com/Azure/open-service-broker-azure/issues/646 and do some small refactors by the way.

### DIFF
--- a/docs/modules/mssql.md
+++ b/docs/modules/mssql.md
@@ -49,7 +49,7 @@ Provisions a new SQL Server and a new database upon that server. The new dbms an
 | `firewallRules[n].name` | `string` | Specifies the name of the generated firewall rule |Y | |
 | `firewallRules[n].startIPAddress` | `string` | Specifies the start of the IP range allowed by this firewall rule | Y | |
 | `firewallRules[n].endIPAddress` | `string` | Specifies the end of the IP range allowed by this firewall rule | Y | |
-| `connectionPolicy` | `string` | Refer [here](https://docs.microsoft.com/en-us/azure/sql-database/sql-database-connectivity-architecture#connection-policy) to specify a connection policy. Valid values are "Redirect", "Proxy", "Default" | N | |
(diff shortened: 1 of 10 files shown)


docs/modules/mssql.md
```diff
@@ -49,6 +49,7 @@ Provisions a new SQL Server and a new database upon that server. The new dbms an
 | `firewallRules[n].name` | `string` | Specifies the name of the generated firewall rule |Y | |
 | `firewallRules[n].startIPAddress` | `string` | Specifies the start of the IP range allowed by this firewall rule | Y | |
 | `firewallRules[n].endIPAddress` | `string` | Specifies the end of the IP range allowed by this firewall rule | Y | |
+| `connectionPolicy` | `string` | Refer [here](https://docs.microsoft.com/en-us/azure/sql-database/sql-database-connectivity-architecture#connection-policy) to specify a connection policy. Valid values are "Redirect", "Proxy", "Default" | N | |
```
norshtein: For provisioning step, a default value should be specified. The same below.

zhongyi-zhang: It is an optional parameter. The provision step can be skip if the field is empty like an update step. Does it make sense?

pkg/services/mssql/plan_details.go
```diff
@@ -281,71 +281,41 @@ func getDBMSCommonUpdateParamSchema() service.InputParametersSchema {
 					},
 				},
 			},
+			"connectionPolicy": &service.StringPropertySchema{
+				Title:       "Connection Policy",
+				Description: "The type of server connection policy.",
+				AllowedValues: []string{
+					"", // Allow empty. It doesn't call the setting method.
+					ConnectionPolicyDefault,
```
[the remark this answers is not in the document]
zhongyi-zhang: Yeah, you're right.

docs/modules/mssql.md
norshtein: It's developer's knowledge, however, if I'm a new user, I'll be confused about what will happen if I don't provide this non-required value, a brief clarification will be better.

zhongyi-zhang: Good point thanks. Updated now.

Issue #651, opened 2018-12-14
Missing titles for properties and enums in Service Plans' JSON schema


pkosiec:
**Description**

In some JSON schemas for Service Plans, there are properties without `title` property. Also, in enums there are no labels for values, which can be achieved in JSON schema with `oneOf`.

**Reasons**

The `title` property is widely used by JSON schema form rendering libraries as a field label. The suggested change, as well as introducing friendly, human-readable labels for enums, will greatly improve usability on UI. 

It's very similar to issue #519. 

**Examples** 

- Missing title in `alias` property for `Azure Database for MySQL 5.7-- DBMS Only` plans
- Missing title for `parentAlias` property:

![screen shot 2018-12-14 at 11 14 02](https://user-images.githubusercontent.com/7155799/49998224-0477c180-ff94-11e8-9773-1b937a0b4c7d.png)

- Missing enum value titles for "Backup redundancy" enum:

![screen shot 2018-12-14 at 11 18 48](https://user-images.githubusercontent.com/7155799/49998216-004ba400-ff94-11e8-8d1e-7452540f4d45.png)

---

**Expected result**
This is how the form based on JSON schema renders when all titles have `title` specified (as an example: Azure SQL Database 12.0 plans):

![screen shot 2018-12-14 at 11 13 41](https://user-images.githubusercontent.com/7155799/49998247-15283780-ff94-11e8-9ece-ffb15d5ab3d0.png)

pkosiec:
@jldeen please take a look at this. Thank you!

norshtein:
Hi @pkosiec , thanks for raising that. The first and the second item will be fixed soon. For the third item 'Missing enum value titles for "Backup redundancy" enum', `enum` is a json array in OSBA's [implementation](https://github.com/Azure/open-service-broker-azure/blob/master/pkg/service/schema.go#L111), each element in this array must be a string, and can't be an object, so a title can't be added to the enum value.

pkosiec:
Hi @norshtein,
Thank you for very fast reaction and your PR! I hope that all titles for all fields are covered now 🙂 

When it comes to the labels for enum values: yes, it would result in slightly changing the implementation. But not much - user would still provide the same `instanceCreateParameterSchema`. The definition of the JSON schema for Service Plans would change in the following way:

So, instead of having this specification:
```
{
    "title": "Before",
    "type": "object",
    "required": [
      "firstName",
      "lastName"
    ],
    "properties": {
[...]

PR #665, opened 2019-01-16
Enhance plan schemas

piotrmiskiewicz:
This PR enhance plan schemas by adding titles to enum values.
Related to #651


piotrmiskiewicz:
The cosmos DB (Mongo) provides a schema with 'readRegions':
```
"readRegions": {
  "type": "array",
  "title": "Read Regions",
  "description": "Read regions to be created, your data will be synchronized across these regions."
                                      },
```

The `items` field is missing. It should looks like:
```
"allowedIPRanges": {
   "type": "array",
   "title": "OneOf IP ranges",
   "description": "Values to include in IP Filter. Can be an IP Address or CIDR range.",
   "items": {
   "type": "string",
   "description": "Must be a valid IP address or CIDR"
    }
 }
```

I'd like to add missing items, but I don't know what should be added there.

pkg/services/servicebus/plan_schemas.go
```diff
@@ -82,7 +82,11 @@ func generateTopicBindingParamsSchema() service.InputParametersSchema {
 				Description: "Specifies whether to create a subscription in the topic." +
 					" Valid values are [\"yes\", \"no\"]. ",
 				AllowedValues: []string{"yes", "no"},
```
polskikiel: You can delete AllowedValues field and adjust the description.

pkg/service/schema.go
```diff
@@ -112,6 +112,7 @@ type StringPropertySchema struct {
 	AllowedPattern          string                        `json:"pattern,omitempty"` // nolint: lll
 	CustomPropertyValidator CustomStringPropertyValidator `json:"-"`
 	DefaultValue            string                        `json:"default,omitempty"` // nolint: lll
+	OneOf                   []EnumValue                   `json:"oneOf,omitempty"`   //nolint: lll
```
polskikiel: Please add space to the `//nolint` comment to align it to the other occurences.

norshtein:
No, they are not the same. Allowed values of Cosmos read region can be found [here](https://github.com/Azure/open-service-broker-azure/blob/master/pkg/services/cosmosdb/cosmos-utils.go#L276-L305). It is slightly different from location values.

pkg/services/servicebus/plan_schemas.go
piotrmiskiewicz: Fixed

piotrmiskiewicz:
I've added read locations to the Cosmos DB Mongo schema

norshtein:
I prefer to keep previous `allowedValues` for some time. Although it seems to be redundant, it can provide backward compatibility. Could you keep `allowedValues` unchanged?

zhongyi-zhang:
I think we don't need to consider backward compatibility in this case. The parsers which leverage React or  Angular won't be impacted. On the contrary, with both `AllowedValues` and `OneOf`, I am not sure whether they are impacted. And, With the announcement in release note, if the change really breaks some parsers, the user still can wait for the fix in the parser side. Make sense?

piotrmiskiewicz:
React jsonschema (https://github.com/mozilla-services/react-jsonschema-form) does not work with both. The 'oneOf' field works very nice

zhongyi-zhang:
Oh, good to know. Then the old should be removed for sure. @piotrmiskiewicz thanks a lot for nice contribution! I’ll merge it as the CI also passed.

piotrmiskiewicz:
When do you plan to do a new release

PR #674, opened 2019-02-14
Simplify getting subscription ID in docs

timja:


zhongyi-zhang:
Will skip CI as it is doc change.

PR #691, opened 2019-04-05
Including Eventhub name and namespace in the returning credentials

### LINKED ISSUE #685: Question: is there a way to discover the created Event Hub name when provisioning the Event Hub
Problem: Creating an Event Hub ServiceInstance appears to need an extra step to discover the created Event Hub name. The broker/secret returns the namespace connection string and primary key, but it does not provide the child Event Hub itself. 

The event hub name is a randomly generated GUID [here](https://github.com/Azure/open-service-broker-azure/blob/7be26814e9a5482b3a590a74ad0b629566bede30/pkg/services/eventhubs/provision.go#L26), so it's not possible to guess/use convention. 

It might be nice if the broker output the generated hub name [here](https://github.com/Azure/open-service-broker-azure/blob/7be26814e9a5482b3a590a74ad0b629566bede30/pkg/services/eventhubs/bind.go#L19)

Am I missing something?

Cheers!

Jordan.

aljannuzzi:
Fixes #685 
I described this minor change as WIP to understand what kind of tests I am required to execute.
Including the EventHub name and namespace in the secrets file generated after the binding operation makes it easier to access and use the created environment.

### DIFF
--- a/pkg/services/eventhubs/bind.go
+++ b/pkg/services/eventhubs/bind.go
@@ -17,7 +17,9 @@ func (s *serviceManager) GetCredentials(
 ) (service.Credentials, error) {
 	dt := instance.Details.(*instanceDetails)
 	return credentials{
-		ConnectionString: string(dt.ConnectionString),
-		PrimaryKey:       string(dt.PrimaryKey),
+		ConnectionString:  string(dt.ConnectionString),
+		PrimaryKey:        string(dt.PrimaryKey),
+		EventHubNamespace: string(dt.EventHubNamespace),
+		EventHubName:      string(dt.EventHubName),
 	}, nil
 }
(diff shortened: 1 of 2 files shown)


zhongyi-zhang:
@aljannuzzi you can follow [this section](https://github.com/Azure/open-service-broker-azure/blob/master/docs/developing.md#the-containerized-development-environment) to do the tests. Before merging a PR, we'll also run them via CircleCI.

aljannuzzi:
@zhongyi-zhang - could you please review this change - I made some minor changes - removed unnecessary conversion and used gomft. If it passes, would that be ok to merge it while I  open an issue to work in parallel on the life-cycle tests?

zhongyi-zhang:
Of course, thanks for your contribution! I'll merge it once the CI passes.

Issue #662, opened 2019-01-09
jdbcUrl for postgres 


olevett:
The MSSQL service provides a jdbcUrl property to the secret (e.g. [here](https://github.com/Azure/open-service-broker-azure/blob/ac21045a8e25fc64f1b3b03a88a3660c91517cc8/pkg/services/mssqldr/common_types.go#L11), which is missing from the MySQL/postgres services.

Is there a reason for this inconsistency, or would a PR for adding JDBC URLs to the other DBs be useful?

jeremyrickard:
PR would be welcome!

zhongyi-zhang:
Solved by https://github.com/Azure/open-service-broker-azure/pull/698.

PR #698, opened 2019-05-02
Add jdbcUrl to postgres binding

olevett:
Covers #662

### DIFF
--- a/pkg/services/postgresql/common_bind.go
+++ b/pkg/services/postgresql/common_bind.go
@@ -105,7 +105,7 @@ func createCredential(
 	connectionString := fmt.Sprintf(
 		connectionTemplate,
 		url.QueryEscape(username),
-		string(bindDetails.Password),
+		url.QueryEscape(string(bindDetails.Password)),
 		fqdn,
 		port,
 		databaseName,
@@ -125,7 +125,7 @@ func createCredential(
 		port,
 		databaseName,
 		url.QueryEscape(username),
-		string(bindDetails.Password),
+		url.QueryEscape(string(bindDetails.Password)),
 	)
 
 	return credentials{
--- a/pkg/services/postgresql/common_bind.go
+++ b/pkg/services/postgresql/common_bind.go
@@ -113,7 +113,7 @@ func createCredential(
 
 	var jdbcTemplate string
 	if sslRequired {
-		jdbcTemplate = 
+		jdbcTemplate =
 			"jdbc:postgresql://%s:%d/%s?user=%s&password=%s&sslmode=require"
 	} else {
 		jdbcTemplate = "jdbc:postgresql://%s:%d/%s?user=%s&password=%s"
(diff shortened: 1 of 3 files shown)


pkg/services/postgresql/common_bind.go
```diff
@@ -110,6 +110,24 @@ func createCredential(
 		port,
 		databaseName,
 	)
+
+	var jdbcTemplate string
+	if sslRequired {
+		jdbcTemplate = 
+			"jdbc:postgresql://%s:%d/%s?user=%s&password=%s&sslmode=require"
```
zhongyi-zhang: Does the `sslmode=require` work? I only had tries with `ssl=true`. Sorry I have little experience on it.

pkg/services/postgresql/common_bind.go
```diff
@@ -110,6 +110,24 @@ func createCredential(
...
+		jdbcTemplate = "jdbc:postgresql://%s:%d/%s?user=%s&password=%s"
+	}
+
+	jdbc := fmt.Sprintf(
+		jdbcTemplate,
+		fqdn,
+		port,
+		databaseName,
+		url.QueryEscape(username),
+		string(bindDetails.Password),
```
zhongyi-zhang: It should also use `url.QueryEscape`, right?

pkg/services/postgresql/common_bind.go
olevett: In my local testing it seems to, and is based on https://www.postgresql.org/docs/current/libpq-ssl.html and is consistent with the pre-existing createCredential calls - https://github.com/Azure/open-service-broker-azure/blob/master/pkg/services/postgresql/common_bind.go#L100

pkg/services/postgresql/common_bind.go
olevett: Quite possibly, I based this on the existing logic - https://github.com/Azure/open-service-broker-azure/blob/master/pkg/services/postgresql/common_bind.go#L108

Will give it a test and see if there's a reason it's not already QueryEscape'd

pkg/services/postgresql/common_bind.go
zhongyi-zhang: Looks good.

pkg/services/postgresql/common_bind.go
zhongyi-zhang: I think it works well because the generated password is alphanumeric (ref: pkg/generate/password.go). But logically it still should use `url.QueryEscape`. It would be better if you'd like to fix both :). Thanks

olevett: Ah right, that makes sense, hadn't dived that deep I'm afraid!

zhongyi-zhang: Thanks for fixing it! I am going to merge it pending CI.

zhongyi-zhang:
Please run `gofmt` to fix the lint issue: `pkg/services/postgresql/common_bind.go:1::warning: file is not gofmted with -s (gofmt)`.

zhongyi-zhang:
I plan to cut a release tomorrow.

PR #699, opened 2019-05-05
Migrate image release destination registry to Azure Container Registry

zhongyi-zhang:
As the write access to microsoft org in DockerHub was revoked, we are migrating to a public ACR instance.
This PR should hold on until ACR team makes our new ACR public.


.circleci/config.yml
```diff
@@ -154,6 +154,7 @@ jobs:
   publish-rc-images:
     <<: *base-docker-job
     environment:
+      DOCKER_REGISTRY: osbapublicacr.azurecr.io/
       DOCKER_REPO: microsoft/
```
krancour: With this set, the final URL will be `osbapublicacr.azurecr.io/microsoft/azure-service-broker:<tag>`, which looks awkward.

I think you probably want to unset this env var so that the final URL for the image will be `osbapublicacr.azurecr.io/azure-service-broker:<tag>`

fwiw... this env var should remain settable, just unset. The reason for this is it allows for overriding the final image URL if people want to build customized versions from their own forks and publish to somewhere else.

Makefile
```diff
@@ -50,11 +50,11 @@ DOCKER_CMD_INT := $(DOCKER_CMD_BASE) -it $(DEV_IMAGE)
 # See https://github.com/Azure/open-service-broker-azure/issues/100
 BASE_IMAGE_NAME        = azure-service-broker
 
-RC_IMAGE_NAME          = $(DOCKER_REPO)$(BASE_IMAGE_NAME):$(GIT_VERSION)
-RC_MUTABLE_IMAGE_NAME  = $(DOCKER_REPO)$(BASE_IMAGE_NAME):canary
+RC_IMAGE_NAME          = $(DOCKER_REGISTRY)$(DOCKER_REPO)$(BASE_IMAGE_NAME):$(GIT_VERSION)
+RC_MUTABLE_IMAGE_NAME  = $(DOCKER_REGISTRY)$(DOCKER_REPO)$(BASE_IMAGE_NAME):canary
```
krancour: Nit: I know `DOCKER_REPO` is a name I chose a long, long time ago, but in retrospect, it was wrong. Now could be a good time to correct it. Technically, in a Docker registry, a repo is a store for a single family of images (i.e. same name, different tags).

User or org names containing multiple repos are probably more accurately called "registry namespaces."

In more recent projects I've worked on, I've changed this env var to `DOCKER_REGISTRY_NAMESPACE` for clarity.

I'm not suggesting you _have_ to do this. Just raising awareness of the inaccuracy of the term `DOCKER_REPO` here and how it can be corrected if you wish.

krancour: I've also included this snippet in some recent projects because it eliminates the unintuitive need for including a trailing slash in registry and namespace env vars:

```
ifdef DOCKER_REGISTRY
	DOCKER_REGISTRY := $(DOCKER_REGISTRY)/
endif

ifdef DOCKER_REGISTRY_NAMESPACE
	DOCKER_REGISTRY_NAMESPACE := $(DOCKER_REGISTRY_NAMESPACE)/
endif
```

contrib/k8s/charts/open-service-broker-azure/README.md
```diff
@@ -101,7 +101,7 @@ Broker chart and their default values.
 | Parameter | Description | Default |
 | --------- | ----------- | ------- |
 | `logLevel` | Log level (options: PANIC, FATAL, ERROR, WARN, INFO, DEBUG). | `"INFO"` |
-| `image.repository` | Docker image location, _without_ the tag. | `"microsoft/azure-service-broker"` |
+| `image.repository` | Docker image location, _without_ the tag. | `"osbapublicacr.azurecr.io/microsoft/azure-service-broker"` |
```
krancour: If (your choice) you do anything about [this comment](https://github.com/Azure/open-service-broker-azure/pull/699/files#r281196692), this address also needs to change.

krancour:
Overall, this looks really sensible. I made a few small comments, but overall 👍

Makefile
zhongyi-zhang: Thanks for the suggestions. Totally agree!

.circleci/config.yml
zhongyi-zhang: Hmm.. I still want to keep "microsoft". With the change as you suggested "DOCKER_REPO" -> "DOCKER_REGISTRY_NAMESPACE", it looks good to me. How do you feel?

krancour:
This looks great!

PR #703, opened 2019-06-12
App Insights module supports creating API Key in binding

### LINKED ISSUE #702: Add API key binding to Application Insights
To be able to access application insights from Grafana, we need an API key. It would be awesome if this could be provisioned with the App Inisghts resource through osba, rather than having to write something separate to create and manage it.

zhongyi-zhang:
resolve #702
BTW upragde Azure Go SDK


pkg/services/appinsights/bind.go
```diff
@@ -1,22 +1,91 @@
...
 	instance service.Instance,
-	_ service.Binding,
+	binding service.Binding,
 ) (service.Credentials, error) {
 	dt := instance.Details.(*instanceDetails)
+	bd := binding.Details.(*bindingDetails)
 	return credentials{
 		InstrumentationKey: string(dt.InstrumentationKey),
+		AppInsightsName:    dt.AppInsightsName,
+		APIKey:             string(bd.APIKey),
```
norshtein: Should the lifecycle test also be updated? https://github.com/Azure/open-service-broker-azure/blob/master/tests/lifecycle/appinsights_cases_test.go#L79

zhongyi-zhang: Agree! Fixed.

zhongyi-zhang:
Thanks for review. CI passed. The last commit is doc only. Will skip CI and merge.

PR #706, opened 2019-06-20
Partially support OSBA in Mooncake

norshtein:
This PR add support for creating resources in Mooncake region. And add a temporary workaround for Mooncake postgreSQL. 

**Note: other services may also have differences between AzurePublicCloud and AzureChinaCloud, I didn't get a chance to figure them out. We should announce that all services in Mooncake are in Preview state.**

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,8 @@ server that provisions managed services in the Microsoft Azure public cloud.
 * [Azure Service Bus](docs/modules/servicebus.md)
 * [Azure Text Analytics (Cognitive Services)](docs/modules/textanalytics.md)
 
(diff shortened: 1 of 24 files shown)


zhongyi-zhang:
Please fix lint. And update the home README what's the differences for the services between Azure public cloud and Mooncake.
One more questions... Does MySQL need the similar workaround for Mooncake?

norshtein:
> Please fix lint. And update the home README what's the differences for the services between Azure public cloud and Mooncake.
> One more questions... Does MySQL need the similar workaround like PostgreSQL for Mooncake?

Updated. 
Yep, both MySQL and MSSQL need the similar workaround.

PR #708, opened 2019-06-21
Update doc of postgreSQL

norshtein:


norshtein:
Docs only, will skip CI.

PR #712, opened 2019-07-18
Fix update error in mssql dr

norshtein:
This PR mitigates lifecycle testing error in our pipeline. We need further investigation and communication with SQL team.


norshtein:
I've tested it offline so I'm going to directly merge it. I'll rebase upon master in previous PR and do a full test.

PR #717, opened 2019-07-24
Add checksum of broker.yaml to annotation in deployment.yaml

ivanov-aleksander:
There is a bug:

OSBA brokes after deployment because every release it creates new TLS certificate in  [broker.yaml](https://github.com/Azure/open-service-broker-azure/blob/master/contrib/k8s/charts/open-service-broker-azure/templates/broker.yaml) 
```yaml
{{- $ca := genCA "osba-ca" 3650 }}
{{- $cert := genSignedCert $cn nil (list $altName1 $altName2 $altName3 $altName4) 3650 $ca }}
```
Error looks like 
```bash
NAME   URL                                                             STATUS                 AGE
osba   https://osba-open-service-broker-azure.osba.svc.cluster.local   ErrorFetchingCatalog   33m
```
OSBA logs
```bash
Error getting broker catalog: Get https://osba-open-service-broker-azure.osba.svc.cluster.local/v2/catalog: x509: certificate signed by unknown authority (possibly because of "crypto/rsa: verification error" while trying to verify candidate authority certificate "osba-ca")
```

The main reason of this error is that osba pod doesn't recreates and it uses certificate from previous release, but  brocker is registered in service catalog with new certificate and ca.

To solve this issue, i have added checksum of broker.yaml to annotation in deployment.yaml related to [Chart Development best prectice](https://github.com/helm/helm/blob/master/docs/charts_tips_and_tricks.md#automatically-roll-deployments-when-configmaps-or-secrets-change)

### DIFF
--- a/contrib/k8s/charts/open-service-broker-azure/templates/deployment.yaml
+++ b/contrib/k8s/charts/open-service-broker-azure/templates/deployment.yaml
@@ -19,6 +19,8 @@ spec:
         chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
         release: "{{ .Release.Name }}"
         heritage: "{{ .Release.Service }}"
+      annotations:
+        checksum/config: {{ include (print $.Template.BasePath "/broker.yaml") . | sha256sum }}
     spec:
       containers:
         - name: open-service-broker-azure
--- a/contrib/k8s/charts/open-service-broker-azure/templates/deployment.yaml
+++ b/contrib/k8s/charts/open-service-broker-azure/templates/deployment.yaml
@@ -19,8 +19,7 @@ spec:
         chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
         release: "{{ .Release.Name }}"
         heritage: "{{ .Release.Service }}"
-      annotations:
-        checksum/config: {{ include (print $.Template.BasePath "/broker.yaml") . | sha256sum }}
+        releaseRevision: "{{ .Release.Revision }}"
     spec:
       containers:
         - name: open-service-broker-azure


contrib/k8s/charts/open-service-broker-azure/templates/deployment.yaml
```diff
@@ -19,6 +19,8 @@ spec:
         chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
         release: "{{ .Release.Name }}"
         heritage: "{{ .Release.Service }}"
+      annotations:
+        checksum/config: {{ include (print $.Template.BasePath "/broker.yaml") . | sha256sum }}
```
norshtein: One thing I'm not sure: [the doc](https://helm.sh/docs/chart_template_guide/#indenting-and-templates) says `{{ include "TEMPLATE" . }}` will render a template and then place its contents, seems when rendering `broker.yaml`, function `genCA` will be invoked, and a different CA from real rendered `broker.yaml` is used to generate the checksum. I'm not an expert of helm, ping @krancour @zhongyi-zhang , is my assumption correct?

ivanov-aleksander: OSBA pod must be restarted if certificate changed. It changes every release because it generates.
To track that certificate was changed I have added checksum of broker.yaml to annotation. @norshtein  look please at official helm recommendation [Automatically Roll Deployments When ConfigMaps or Secrets change](https://github.com/helm/helm/blob/master/docs/charts_tips_and_tricks.md#automatically-roll-deployments-when-configmaps-or-secrets-change)

norshtein: yeah, I got your point... what I'm not sure is: here `broker.yaml` is introduced using function `include`, and the doc says function `include` will **render** included file, so when rendering `broker.yaml`, function `genCA` in `broker.yaml` will be called, this generated CA will be different from CA in real `broker.yaml`

krancour:
I'm surprised this hasn't come up before now.

At any rate, there's an established pattern to ensure pods restart on every helm release. Add the following label to the deployment's pod template metadata:

```
releaseRevision: "{{ .Release.Revision }}"
```

Conceptually, this is equates to what you are already doing in this PR, but this is the already commonplace method of doing it.

ivanov-aleksander:
@krancour  you are right. I have added a label. 
@krancour @norshtein  could you please review and merge.
The lask question, when will be a new docker image that contains these changes?

krancour:
Thanks for the change.

PR #732, opened 2019-12-12
Support creating v3.6 mongoDB account

norshtein:


norshtein:
CI failed due to following error:
```
This subscription is restricted from provisioning PostgreSQL servers in this region. Please choose a different region or open a support request with service and subscription limits (quotas) issue type.
```

Seems some thing went wrong with the quota for the testing subscription. Let's check with @jeremyrickard  and @krancour . By the way, I've run the lifecycle test locally in my own subscription, and it succeeded.

norshtein:
I've temporarily set the test subscription to another subscription to fix the lifecycle error.

norshtein:
@zhongyi-zhang I'm going to merge this, FYI.

PR #738, opened 2020-05-03
Allow deployment on kubernetes 1.16+

rkapoz-tech:
Hi, I discovered an issue trying to deploy OSBA on Kubernetes 1.16, due to the promoted Deployment API. Also I changed the image tag


timja:
@norshtein would be good to take a look at this ^^

FYI we had the same issue, and also had to update the redis version this chart depends on:
https://github.com/Azure/open-service-broker-azure/compare/master...hmcts:master

